PR #326, opened 2014-06-24
adding Informix environment

struesda:
I am working with RaviKkumarReddy on the informix wrapper.
I have followed your steps to fork the repo, add Ravi's new files, and commit the changes back to my fork.
(previously we were compiling the new informix jar outside of the rest of the dbfit structure)

At this point we have 2 problems:

**1. Gradle compile**
I have updated the gradle files to verify that I can use gradle to compile the new jars.
However when I try to run the jar task - it cannot find the Informix JDBC class.
What do we need to do in order to get the gradle process to use the Informix JDBC jar when compiling?

**2. Connecting to Informix - non-transactional**
This is our core problem.  We need to connect to a database that has no transaction logging.  
Using the normal code - when trying to connect we get a 'Transactions not Supported' error.

At the moment we are getting around this by overriding the connect() method, looking for our specific database name, and handing the setAutoCommit() call appropriately. (as well as overriding the rollback() method the same way)

But we are hoping there is a better way to do this without having to resort to hard coding things like this.  Is there any way to add some connection string parameters that could be intercepted by our overridden methods?  Any other suggestions?

**3. Procedure Parameters**
Once we have resolved the issues above - we can look at the best way to handle procedure parameters from Informix.

Thanks.


MMatten:
@javornikolov - I'd like to help with this if possible to help mop up some of the open PRs.

@struesda - are you still working on this? Do you think we could look at merging this? Can I help at all?

liwenbin1991:
If it is a distributed database, in order to ensure that the query from the different nodes of the same data, may I ask how to do it in a Test page? I now think that the two test page to complete the requirement, if a test page, the better. Thank you!

MMatten:
@liwenbin1991 - please try asking the user forum at https://groups.google.com/forum/#!forum/dbfit.

Also note that we're in the process of reviewing and Informix adapter submission so that it can be formally released (sometime soon).

PR #352, opened 2015-04-08
mssql stored procedure implicit return value support

javornikolov:
Add support for capturing procedure implicit return value (integer)

This is intended to resolve  #351

Issue #351, opened 2015-04-05
Stored Proc implicit return value SQL Server 2014


OB4:
If I try the following code, I get a cast-error. It isn't possible to cast from system.DateTime to system.TimeSpan:

!|Insert|dbo.Zeit|
|Zeit|
|08:10:00|

What goes wrong?

javornikolov:
@OB4 , could you please share the exact error message and stack you're getting. An also the table definition - what's the exact data type of Zeit column in the database?

OB4:
To show  the error I have a simple table definition:
Id (PS, int, Not null)
Zeit (time(7), NULL)

Here the test:
!|Insert|dbo.Zeit|
|Zeit|
|08:00:00|

The exact error message:

System.InvalidCastException: Parameterwert konnte nicht von DateTime in TimeSpan umgewandelt werden. ---> System.InvalidCastException: Ungültige Umwandlung von "System.DateTime" in "System.TimeSpan".
   bei System.Convert.DefaultToType(IConvertible value, Type targetType, IFormatProvider provider)
   bei System.DateTime.System.IConvertible.ToType(Type type, IFormatProvider provider)
   bei System.Convert.ChangeType(Object value, Type conversionType, IFormatProvider provider)
   bei System.Data.SqlClient.SqlParameter.CoerceValue(Object value, MetaType destinationType, Boolean& coercedToDataFeed, Boolean& typeChanged, Boolean allowStreaming)
   --- Ende der internen Ausnahmestapelüberwachung ---
   bei System.Data.SqlClient.SqlParameter.CoerceValue(Object value, MetaType destinationType, Boolean& coercedToDataFeed, Boolean& typeChanged, Boolean allowStreaming)
   bei System.Data.SqlClient.SqlParameter.GetCoercedValue()
   bei System.Data.SqlClient.SqlParameter.Validate(Int32 index, Boolean isCommandProc)
   bei System.Data.SqlClient.SqlCommand.BuildParamList(TdsParser parser, SqlParameterCollection parameters)
   bei System.Data.SqlClient.SqlCommand.BuildExecuteSql(CommandBehavior behavior, String commandText, SqlParameterCollection parameters, _SqlRPC& rpc)
   bei System.Data.SqlClient.SqlCommand.RunExecuteReaderTds(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, Boolean async, Int32 timeout, Task& task, Boolean asyncWrite, SqlDataReader ds)
[...]

PR #353, opened 2015-04-11
fix mssql proc params metadata query errors

javornikolov:
This is fixing syntax errors in mssql metadata query introduced by #352 

With this the basic case of #351 is working OK.

As per #351, it has been concluded that there is no need to address the subtle case when RAISEERROR is used to have both return value and error raised.


javornikolov:
Merging this as in #351 it has been confirmed that it's working good enough.

PR #356, opened 2015-04-20
Enable hsql tests

MMatten:
This introduces the CoreTests for HSQL.

It also allows mixed case table names to be accessed by Insert/Update/Clean as HSQL converts unquoted names to upper case upon creation.

Tests are executed against a memory-only database to make set-up simpler but this might be better to start up a database as part of the VM provisioning so that the DDL for the CoreTests can be moved out of the SetUp.

``hsqldb-2.3.2.jar`` is currently being pulled from ``custom_libs``.

### DIFF
--- a/dbfit-java/hsqldb/src/test/java/dbfit/environment/HSQLDBEnvironmentFactoryTest.java
+++ b/dbfit-java/hsqldb/src/test/java/dbfit/environment/HSQLDBEnvironmentFactoryTest.java
@@ -2,12 +2,10 @@ package dbfit.environment;
 
 import dbfit.api.DBEnvironment;
 import org.junit.Test;
-import org.junit.Ignore;
 import static org.junit.Assert.*;
 
 public class HSQLDBEnvironmentFactoryTest {
 
-    @Ignore("HSQLDB driver not available during build")
     @Test
     public void newDbEnvironmentTest() throws Exception {
         DBEnvironment env = dbfit.api.DbEnvironmentFactory.newEnvironmentInstance("HSQLDB");
--- a/dbfit-java/hsqldb/src/test/java/dbfit/environment/HSQLDBRegressionTest.java
+++ b/dbfit-java/hsqldb/src/test/java/dbfit/environment/HSQLDBRegressionTest.java
@@ -1,0 +1,20 @@
+package dbfit.environment;
+
+import fitnesse.junit.FitNesseSuite;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import static fitnesse.junit.FitNesseSuite.*;
+
+public class HSQLDBRegressionTest {
+
+    @RunWith(FitNesseSuite.class)
+    @Name("DbFit.AcceptanceTests.JavaTests.HsqlTests.FlowMode")
+    @FitnesseDir("../..")
+    @OutputDir("../../tmp")
+    public static class FlowModeTest {
+        @Test public void dummy(){}
+    }
+
+}
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,7 +24,7 @@ The quickest way to get feedback on your code changes is to run the fast build:
 
     $ ./gradlew clean fastbuild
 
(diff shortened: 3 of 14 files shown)


MMatten:
Also start to address #292.

javornikolov:
> `hsqldb-2.3.2.jar` is currently being pulled from `custom_libs`.

Would it be possible to pick that from some artifact repository? E.g. maybe from here: https://maven-repository.com/artifact/org.hsqldb/hsqldb/2.3.2

dbfit-java/hsqldb/src/test/java/dbfit/environment/HSQLDBRegressionTest.java
```diff
@@ -0,0 +1,21 @@
+package dbfit.environment;
+
+import fitnesse.junit.FitNesseSuite;
+
+import org.junit.Ignore;
```
javornikolov: `Ignore` annotation is not referenced below, no need to import it.

dbfit-java/hsqldb/src/test/java/dbfit/environment/HSQLDBEnvironmentFactoryTest.java
```diff
@@ -7,7 +7,6 @@
 
 public class HSQLDBEnvironmentFactoryTest {
 
-    @Ignore("HSQLDB driver not available during build")
```
javornikolov: Now we don't need to `import Ignore`.

MMatten:
After removing the HSQL JAR from custom_libs and the gradle cache (at /home/vagrant/.gradle/...) the build then automatically pulls the JAR from ``https://repo1.maven.org/maven2/org/hsqldb/hsqldb/2.3.2/hsqldb-2.3.2.jar``.

MMatten:
One more option here - include the HSQL tests in the fastbuild task. As the current HSQL tests are using a memory only DB they run quickly.

What do you think?

benilovj:
> One more option here - include the HSQL tests in the fastbuild task. As the current HSQL tests are using a memory only DB they run quickly.

:+1: this is a good idea. The tests are much less likely to break this way.

javornikolov:
Yeah, I agree. I'm just running hsqldb tests locally - it's so quick, that I'm wondering whether it executed any test at all. I'm trying to verify...

javornikolov:
> Yeah, I agree. I'm just running hsqldb tests locally - it's so quick, that I'm wondering whether it executed any test at all. I'm trying to verify...

Verified - it's all working OK.

@MMatten, could you squash your commits into a single one and force-push the branch (since most are just fix-ups, not so interesting history). And then we can merge the PR, looks good to me.

MMatten:
Ok, will do.

I think I should now also remove ``**/hsqldb*.jar`` from the list of ``proprietaryLibs``.

javornikolov:
> I think I should now also remove **/hsqldb*.jar from the list of proprietaryLibs.

Yes

dbfit-java/hsqldb/src/test/java/dbfit/environment/HSQLDBRegressionTest.java
```diff
@@ -0,0 +1,20 @@
+package dbfit.environment;
```
javornikolov: looks like that file is in dos-style end of file

MMatten: I'd edited from Windows Eclipse but checked in from the VM and hadn't set ``autocrlf`` to ``input``.

MMatten:
One more commit to push in a minute - updating the main doc file.

MMatten:
Right, final doc edits done.

javornikolov:
Awesome! I'm going to merge this PR.
@MMatten, thanks a lot for your contribution!

MMatten:
Great. 

Does this kill off #292?

javornikolov:
> Does this kill off #292?

Yes, this resolves #292

PR #358, opened 2015-04-20
Add missing DB2 Express C dependency (installer warning)

MMatten:
This update installs libstdc++.so.6. It's absence is warned by the DB2 Express-C installer.


javornikolov:
:+1:

Issue #348, opened 2014-12-01
MS SQL Server - support for case-sensitive database object names


javornikolov:
This originally emerged on the [DbFit mailing list](https://groups.google.com/d/msg/dbfit/6pLuUl1KFGM/dFUzDM2EIVEJ).

Seems DbFit is currently unable to handle certain operations when the underlying database objects are configured as case-sensitive.

MMatten:
I'd like to pick this one up to resolve, via a PR, it as the forum post came from a user at my workplace.

I've read through the forum post and I've done some experimenting.

The code that tries to break up the table name into parts and enclose the parts with ``[`` and ``]`` definitely doesn't work properly with 3 part (DB.schema.table) names.

When the collation is set in the DB (server or DB or object) to a case specific set (e.g. ``Latin1_General_CS_AS`` rather than ``Latin1_General_CI_AS`` -- ``CS` == case specific) then the ``name.toLowerCase()`` in ``NameNormaliser.normaliseName`` causes the issue.

I was thinking of simply changing ``dc.setString(1, normaliseName(objname));`` to ``dc.setString(1, objname);`` in ``SqlServerEnvironment.readIntoParams`` or normalising the name within this method without the case conversion.

There are existing broken SQL Server tests and the tests setup SQL script and the tests are not aligned in terms of object name cases so when I've tried creating the ``FitNesseTestDB" database with case sensitive collation then some of the tests are failing anyway. I'll try updating the SQL script and tests so that the object name cases are aligned then at least I'll be able to see if the code change is breaking anything else.

Some of the broken tests appear to be because the feature hasn't been implemented in the Java version (e.g. binary data type) but presumably was in the .Net version. I guess I can disable these tests. **Can you tell me how the disabled Oracle tests have been disabled so that I can try this too?**

MMatten:
Aligning the tests to the setup SQL script (updating the table name in all fixture types and updating the column names in ``Query`` fixture queries) is all I've needed to do to get all of the tests (apart from those for unimplemented features) passing so it looks like there are no tricky changes required to ensure that row based fixtures don't have problems with mixed case column names.

MMatten:
BTW it's also easy to get the whole `CoreTests` suite running against SQL Server so we could remove some of the specific/redundant tests at some point.

javornikolov:
Hi @MMatten

> I'd like to pick this one up to resolve, via a PR, it as the forum post came from a user at my workplace.

Awesome :-)

[...]

PR #361, opened 2015-04-28
Fix SQL Server case sensitive collation issues - phase 1 - disable broken tests

MMatten:
PR1 - park broken SQL Server tests.

The tests are broken because the feature hasn't been implemented in Java. E.g. SQL Variant in StringTest and Binary data types


MMatten:
Start of resolving #348.

javornikolov:
Are the tests broken due to the case-sensitivity issue or for other reason?

Now I realized that when I suggested separate PRs in #348 I imagined we'll be moving tests which are broken in general (not due to case-sensitivity).

MMatten:
These tests, that I've moved, are broken because the feature hasn't been implemented in Java.

javornikolov:
OK, then it's good that we have that in separate PR. I'll change the PR title a bit and I'm merging.

PR #362, opened 2015-04-29
Support case sensitive collation object names in SQL Server

MMatten:
Support case sensitive/insensitive collations.
Remove redundant StandAlone mode regression test call.
Tidy and complete DB setup scripts.
Support 3 part table/view/procedure names.

### DIFF
--- a/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/SqlServerTests/properties.xml
+++ b/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/SqlServerTests/properties.xml
@@ -2,13 +2,14 @@
 <properties>
 	<Edit/>
 	<Files/>
-	<LastModified>20080330043718</LastModified>
 	<Properties/>
 	<RecentChanges/>
 	<Refactor/>
 	<Search/>
 	<Suite/>
-	<Suites/>
+	<SymbolicLinks>
+		<DisabledTests>.DbFit.AcceptanceTests.DotNetTests.SqlServerTests.DisabledTests</DisabledTests>
+	</SymbolicLinks>
 	<Versions/>
 	<WhereUsed/>
 	<saveId>1206796425000</saveId>
--- a/dbfit-java/sqlserver/src/main/java/dbfit/environment/SqlServerEnvironment.java
+++ b/dbfit-java/sqlserver/src/main/java/dbfit/environment/SqlServerEnvironment.java
@@ -6,6 +6,7 @@ import dbfit.util.DbParameterAccessor;
 import dbfit.util.DbParameterAccessorsMapBuilder;
 import dbfit.util.Direction;
 import static dbfit.util.Direction.*;
+import static dbfit.util.LangUtils.enquoteAndJoin;
 import dbfit.util.TypeNormaliserFactory;
 import static dbfit.environment.SqlServerTypeNameNormaliser.normaliseTypeName;
 
@@ -79,16 +80,10 @@ public class SqlServerEnvironment extends AbstractDbEnvironment {
         DbParameterAccessorsMapBuilder params = new DbParameterAccessorsMapBuilder();
 
         objname = objname.replaceAll("[^a-zA-Z0-9_.#]", "");
-        String[] qualifiers = objname.split("\\.");
-        StringBuilder bracketedName = new StringBuilder();
-        String separator = "";
-        for (int i = 0; i < qualifiers.length; i++) {
-            bracketedName.append(separator + "[" + qualifiers[i] + "]");
-            separator = ".";
-        }
+        String bracketedName = enquoteAndJoin(objname.split("\\."), ".", "[", "]");
 
         try (PreparedStatement dc = currentConnection.prepareStatement(query)) {
-            dc.setString(1, bracketedName.toString());
+            dc.setString(1, bracketedName);
             ResultSet rs = dc.executeQuery();
 
             while (rs.next()) {
--- a/dbfit-java/sqlserver/src/main/java/dbfit/environment/SqlServerEnvironment.java
+++ b/dbfit-java/sqlserver/src/main/java/dbfit/environment/SqlServerEnvironment.java
@@ -6,7 +6,6 @@ import dbfit.util.DbParameterAccessor;
 import dbfit.util.DbParameterAccessorsMapBuilder;
 import dbfit.util.Direction;
 import static dbfit.util.Direction.*;
-import static dbfit.util.NameNormaliser.normaliseName;
 import dbfit.util.TypeNormaliserFactory;
 import static dbfit.environment.SqlServerTypeNameNormaliser.normaliseTypeName;
 
@@ -15,7 +14,6 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.Arrays;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.regex.Pattern;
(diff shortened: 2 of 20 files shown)


FitNesseRoot/DbFit/AcceptanceTests/JavaTests/SqlServerTests/properties.xml
```diff
@@ -2,13 +2,14 @@
...
 	<Files/>
-	<LastModified>20080330043718</LastModified>
 	<Properties/>
 	<RecentChanges/>
 	<Refactor/>
 	<Search/>
 	<Suite/>
-	<Suites/>
+	<SymbolicLinks>
+		<DisabledTests>.DbFit.AcceptanceTests.DotNetTests.SqlServerTests.DisabledTests</DisabledTests>
```
javornikolov: Why do we need a link to `DisabledTests`?

dbfit-java/sqlserver/src/main/java/dbfit/environment/SqlServerEnvironment.java
```diff
@@ -80,15 +78,18 @@ protected String parseCommandText(String commandText) {
...
         DbParameterAccessorsMapBuilder params = new DbParameterAccessorsMapBuilder();
 
-        if (objname.contains(".")) {
-            String[] schemaAndName = objname.split("[\\.]", 2);
-            objname = "[" + schemaAndName[0] + "].[" + schemaAndName[1] + "]";
-        } else {
-            objname = "[" + normaliseName(objname) + "]";
+        objname = objname.replaceAll("[^a-zA-Z0-9_.#]", "");
+        String[] qualifiers = objname.split("\\.");
+        String bracketedName = "";
```
javornikolov: What about using StringBuilder?

dbfit-java/sqlserver/src/main/java/dbfit/environment/SqlServerEnvironment.java
```diff
@@ -80,15 +78,18 @@ protected String parseCommandText(String commandText) {
...
+        for (int i = 0; i < qualifiers.length; i++) {
+            if (i > 0) {
+                bracketedName += ".";
```
javornikolov: I wonder if it's worth to add a `separator` variable to avoid the nested if?

dbfit-java/sqlserver/src/test/resources/acceptancescripts-SqlServer.sql
```diff
@@ -208,4 +247,11 @@ as
...
 set @copyOfInParam = @inParam
 set @constOutParam = 123.456;
-end
\ No newline at end of file
+end
+GO
+
+create procedure [dbo].[MakeUser] AS
+begin
+	insert into Users (Name,UserName) values ('user1','fromproc');
```
javornikolov: Missing white-space between `Name,` and `Username`; and also between the values.

dbfit-java/sqlserver/src/test/resources/acceptancescripts-SqlServer2000.sql
```diff
@@ -216,4 +216,9 @@ as
 begin
 set @copyOfInParam = @inParam
 set @constOutParam = 123.456;
-end
\ No newline at end of file
+end
+
+create procedure [dbo].[MakeUser] AS
+begin
+	insert into Users (Name,UserName) values ('user1','fromproc');
```
javornikolov: Missing whitespace.

dbfit-java/sqlserver/src/test/resources/acceptancescripts-SqlServer.sql
```diff
@@ -208,4 +247,11 @@ as
...
+end
+GO
```
javornikolov: Missing newline at end of file.

javornikolov:
Thank you, @MMatten! In general looks good. I added a few minor remarks.

Have you tested that for both case-sensitive and non-case sensitive database?

MMatten:
Yes tests pass using both collation sequence ``Latin1_General_CS_AS`` and ``Latin1_General_CI_AS``.

I'll work on the above points. Thanks for the quick review.

FitNesseRoot/DbFit/AcceptanceTests/JavaTests/SqlServerTests/properties.xml
MMatten: This is so the `DisabledTests` suite is visible under the `JavaTests` suite else they're invisible unless you visit the `DotNetTests` suite.

Do you think this is the wrong thing to do? I thought I would be another reminder to get them working some time!

dbfit-java/sqlserver/src/main/java/dbfit/environment/SqlServerEnvironment.java
MMatten: I must admit to being ignorant of the benefits of `StringBuilder` over what I've implemented. Could you give me a quick briefing?

Which methods would you use, apart from, presumably, `append()`?

dbfit-java/sqlserver/src/main/java/dbfit/environment/SqlServerEnvironment.java
MMatten: I don't quite understand what you're suggesting. Could you explain in another way?

MMatten: Ah, you mean something like set it to `"."` initially, then set it to `""` after the first iteration?

dbfit-java/sqlserver/src/main/java/dbfit/environment/SqlServerEnvironment.java
javornikolov: Java Strings are immutable so when you do `+=`, a new String instance is created (the old ones are garbage collected). [`StringBuilder`](https://docs.oracle.com/javase/7/docs/api/java/lang/StringBuilder.html) is intended to avoid that overhead to hopefully improve performance. `append` method and finally - `toString`.

FitNesseRoot/DbFit/AcceptanceTests/JavaTests/SqlServerTests/properties.xml
javornikolov: OK. It's not wrong, I was just curious.

dbfit-java/sqlserver/src/main/java/dbfit/environment/SqlServerEnvironment.java
javornikolov: Yes.

MMatten: Yes, indeed. Got it.

dbfit-java/sqlserver/src/main/java/dbfit/environment/SqlServerEnvironment.java
MMatten: So would be also be better to create constants for the `[` and `]` too in the method and use those in the loop instead of literals? Or is that overkill?

javornikolov: I think it's an overkill. Literals are fine.

MMatten:
I think all of the points have been addressed now. Could you review again?

dbfit-java/sqlserver/src/test/resources/acceptancescripts-SqlServer.sql
```diff
@@ -1,211 +1,257 @@
...
+execute dbo.sp_executesql @statement = N'-- =============================================
+-- Author:		<Author,,Name>
+-- Create date: <Create Date,,>
+-- Description:	<Description,,>
+-- =============================================
+CREATE FUNCTION [dbo].[ReturnUserTable_F]
+(
+	@howmuch int
+)
+RETURNS
```
javornikolov: Trailing whitespace here and on a few other places.

javornikolov:
> I think all of the points have been addressed now. Could you review again?

Thanks! There is some trailing whitespace in `acceptancescripts-SqlServer.sql` and also it's in DOS format.

MMatten:
I can see the trailing spaces but I can't see any `CR`s, only `LF`s.

dbfit-java/sqlserver/src/test/resources/acceptancescripts-SqlServer.sql
```diff
@@ -1,211 +1,257 @@
...
+	-- Fill the table variable with the rows for your result set
+	DECLARE @i INT
+	SET @i = 0
+	WHILE (@i < @howmuch)
+	BEGIN
+		SET @i = @i + 1
+		INSERT @userTable([user], [username])
+			VALUES(''User '' + CAST(@i AS VARCHAR(10)), ''Username '' + CAST(@i AS VARCHAR(10)))
+	END
+
```
javornikolov: One more trailing space here.

javornikolov:
> I can see the trailing spaces but I can't see any CRs, only LFs.

I can't see CR either - seems OK now. Just one more space-only line remains.

MMatten:
Found the blighter. It was a single tab on the line.

MMatten:
I'm going to be away now until Monday so I'm going to, optimistically, squash the commits and push again.

javornikolov:
> I'm going to be away now until Monday so I'm going to, optimistically, squash the commits and push again.

Great! I was just going to ask for squashing the whitespace re-formatting and we're ready to merge then :-)

Issue #88, opened 2013-04-14
Travis/fastbuild - maybe just skip heavy tests but still compile and fast-unit-test all


javornikolov:
Seems that presently fastbuild doesn't even compile and run small unit tests for some of the db environments.
Maybe would be better to compile everything. And run whatever tests are considered *fast*.

javornikolov:
Resolved by #365

PR #365, opened 2015-05-05
Separate Unit and Integration Tests build tasks

MMatten:
Separate unit and integration (acceptance) tests build tasks.

Netezza is usually unavailable during builds, hence disabling the execution of the acceptance tests.

Resolves: #88


benilovj:
I can see the intent behind this change. However I think it'd be better to take the netezza build out of the main build, rather than disable the tests, because the tests are actually valid, as far as I know. Disabling the tests at JUnit level means that somebody with netezza can't actually run the tests without modification.

@javornikolov what do you think?

javornikolov:
I think we should rather try to figure out how setup some optional config file (one more) where people can customize which tests to be executed and which now.

MMatten:
Yes, I think that's better too. I've just followed the SQL Server and Teradata patterns.

Is there anything that can be specified whilst running the build / options to gradlew? Even supply a shell script?

MMatten:
Also it might be good to prevent only certain sets of *acceptance* tests running, not every test for a sub project. I.e. run all of the Java only unit tests but omit the tests that require the DB.

Is it possible to split the tests in the build into 'Java only' and 'acceptance'?

javornikolov:
Yes, that's another good idea. One related discussion I see here: https://discuss.gradle.org/t/add-support-for-integration-tests/567/2

MMatten:
That looks interesting. I'm not sure anything's actually been implemented in gradle though so perhaps the suggestion by Paul Jimenez ("Isn't it just a new task with different includes/excludes?") could do us.

What do you think?

javornikolov:
We need to research a bit - the quoted threads are a few years old now. Maybe might be new options & ideas available.

MMatten:
I'm a bit of a gradle novice but could be using a separate integration tests 'source set' be the way to go?

MMatten:
I've been playing around with this. I can add a new source set ``intTest`` and a new task ``intTest`` but it looks like I need to put this section into each of the sub projects: -

`    sourceSets {
        intTest {
            java {
                srcDir 'src/integTest/java'
            }
            resources {
                srcDir 'src/integTest/resources'
            }
        }
    }

    task intTest(type: Test) {
        testClassesDir = sourceSets.intTest.output.classesDir
        classpath = sourceSets.intTest.runtimeClasspath
    }`

MMatten:
Even if we can execute integration tests separately from Java only unit tests we still need a way to tell the build for which adaptors to execute them.

MMatten:
Actually, I don't need to add it to each project. Adding it to the `subprojects` does the trick. (Sorry, just slowly working out what this thing's doing!).

MMatten:
I've completed my experiment of using a custom `intTest` sourceSet for the DB integration tests and leaving the Java only/unit tests in the standard `test` sourceSet.

I've created a task called `integTest`. So `./gradlew check` runs the unit tests and './gradlew integTest` runs the integration (FitNesse acceptance) tests.

I've updated the `fastbuild` task to run all of the unit tests and created a new `integTest` task (perhaps this is the bit that users should modify during their builds to specify which FitNesse tests should actually run).

What do you think of this lot?

javornikolov:
@MMatten, looks awesome at first glimpse!

* I guess we have to change the PR title :-)
* Some of the moved scripts are referenced in the test vm provisioning. (Paths need to be updated)
* I'm not quite sure for the naming: intTest, integTest, integrationTest, integration-test... Right now we're using `intTest` and `integTest` - should we have a different name?

> I've updated the fastbuild task to run all of the unit tests and created a new integTest task (perhaps this is the bit that users should modify during their builds to specify which FitNesse tests which should actually run).

OK. I suspect it's possible `dependsOn` definition to be customized e.g. via loading some .gitignore-d config file.

MMatten:
Thanks @javornikolov,

Apart from the DB2/Oracle shell scripts, which I've handled, do you know what other paths needs updating for the VM provisioning, off the top of your head? I've not tried building a VM from this code base yet. I'll try that tomorrow.

As for `intTest` vs `integTest` - I left this different intentionally whilst doing these updates to help clarify exactly what was going on. Although I've just learnt a bit about gradle, I'm definitely still a novice. Any direction on changes to what I've done would be most welcome. So please let me know if you think we should rename anything. 

Regarding the `dependsOn` idea - can you explain a little more?

javornikolov:
> Apart from the DB2/Oracle shell scripts, which I've handled, do you know what other paths needs updating for the VM provisioning, off the top of your head?

You can check via `git grep 'test/resources'` to see content referring to `test/resources`. Maybe no more for the VM provisioning but there are docs and acceptance tests affected too.

MMatten:
Right, I'll look for those.

As for the `intTest` vs `integTest` vs something else - from googling around there doesn't seem to be a consistent convention in use. Any preferences?

javornikolov:
`intTest` vs `integTest`:
 - if we consider that it helps to have different names: how does that help? Maybe more descriptive name would help to distinguish between both better.
 - Presently both seem pretty similar (hard to remember which is what), obviously abbreviation of the same. If we anyway need the context to distinguish which is what - maybe just a single name would be enough (if not causing some complains in Gradle)?

javornikolov:
I can see an [example for build with integration tests](https://github.com/gradle/gradle/blob/eb4d045f12dab4fa12a7da7ee99b6a70fe024a66/subprojects/docs/src/samples/java/withIntegrationTests/build.gradle) in project Gradle itself. The naming used there (tasks, sourcesets and folders) seem reasonable to me at first glimpse.

MMatten:
Thanks for that. I think those names look neater. So `integration-test` for the source directories, `integrationTest` for the task.

Shall we go for this?

javornikolov:
> Thanks for that. I think those names look neater. So integration-test for the source directories, integrationTest for the task.
> 
> Shall we go for this?

OK, let's try it.

MMatten:
May be a subject for another PR, but I've also noticed that the SQL Server JDBC driver can be redistributed:

https://msdn.microsoft.com/en-us/sqlserver/jj589698

I've previously submitted a copy of the form.

Or even if it wasn't included in the dbfit distro could we get a copy onto the AWS repository to run the SQL Server unit tests? 

I'm still not sure what the jDTS is doing for us in the build. Removing the dependency in the build doesn't break anything and it doesn't even help the SQL Server `SqlServerDbEnvironmentFactoryTest` to run because of the different driver class name.

MMatten:
Ok, directory names updated and `intTest` task renamed to `integrationTest`. I've used the build task name `integrationBuild` in the top level `build.gradle` script so it doesn't conflict with the `integrationTest` task in the `dbfit-java/build.gradle` script because if the user runs `./gradlew integrationTest` then *all* of the adaptors' `integrationTest` tasks get run which isn't what we want.

What was that idea you mentioned about Git ignore?

FitNesseRoot/DbFit/AcceptanceTests/JavaTests/TeradataTests/FlowMode/CallingFunctions/content.txt
```diff
@@ -1,4 +1,4 @@
-!path ../dbfit-java/teradata/src/test/resources
+!path ../dbfit-java/teradata/src/test/integration-test
```
javornikolov: I guess this needs to be `integration-test/resources`

dbfit-java/build.gradle
```diff
@@ -172,7 +191,7 @@ project('sqlserver') {
     description = 'DBFit sqlserver engine'
 
     dependencies {
-        compile group: 'net.sourceforge.jtds', name: 'jtds', version: '1.2.2'
+        //compile group: 'net.sourceforge.jtds', name: 'jtds', version: '1.2.2'
```
javornikolov: I think we can merge the other ms sqlserver PR #366 first. And then rebase that on top of it - this line will like cause a small conflict. (Should be easy to handle I guess, just noting it).

javornikolov:
> What was that idea you mentioned about Git ignore?

The idea was about your comment that people need to tweak `build.gradle` if they want to select which tests to run or not (different than the default). Which needs special care not to include these local customizations when committing.

javornikolov:
Now that we have #366 - we can rebase and incorporate updates for ms sqlserver too - e.g. I guess we can remove @Ignore annotations.

Special per-user customization of tasks to run, .gitignore, etc. - perhaps we can leave for separate PR.

MMatten:
I'm going to try building a new test VM off of this branch now and run all of the tests to check for any remaining issues.

MMatten:
I think everything's in order now.

javornikolov:
Looks OK to me. Just we can tidy up a bit the commits history a bit before merging.
@benilovj?

MMatten:
Great. I've just pushed one more commit - to add the disabled integration tests as *commented out* lines to help show users how to enable/disable integration tests.

Does look reasonable?

benilovj:
:+1: looks good, although as @javornikolov pointed out, maybe a couple of commits could be squashed.

MMatten:
Now I've reformatted those builds tasks I noticed that the DB2 integration tests were missing from the `integrationBuild` task. I think that's the last fixup. I'll squash a few commits and push again.

javornikolov:
As far as I see - db2 tests are already in the `integrationBuild`. @MMatten , we're ready to merge, right?

MMatten:
Yes, the last commit before squashing added them. 

All ready to merge.

javornikolov:
OK, merging... And thank you again @MMatten for that really good contribution!

MMatten:
Thank you guys. It's a pleasure working with you.

PR #366, opened 2015-05-10
sqlserver test dependency on mssql jdbc driver

javornikolov:
build: declare sqlserver dependency on mssql jdbc driver

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -171,7 +171,7 @@ def onWindows() {
 }
 
 def proprietaryLibs() {
-    ['**/ojdbc*.jar', '**/db2jcc4*.jar', '**/terajdbc*.jar', '**/tdgssconfig*.jar', '**/nzjdbc*.jar'] as Set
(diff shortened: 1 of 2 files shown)


javornikolov:
This is to address a point raised in #365
@MMatten, can you test?

MMatten:
Great. I test this out.

MMatten:
Working well. :thumbsup:

MMatten:
It's not actually a dependency for the `compileJava` or `integrationTestCompileJava` stages but it does provide the dependency for both the `test` and the `integrationTest` tasks.

It is better to list it as a `testRuntime` and `integrationTestRuntime` dependency?

javornikolov:
Is it for both `test` and `integration` or just for integration? Anyway - already merged into master. I think we can rebase #365 and specify the more precise dependency there.

MMatten:
We need the driver for both sets of tests.

javornikolov:
Yes, I just looked and observed the same - doesn't really connect or use the driver, just loading it. OK.

PR #368, opened 2015-05-18
Migrate SQL Server to CoreTests suite

MMatten:
The main thrust of this PR is to migrate the SQL Server adaptor to use the `CoreTests` suite.
A new `CoreTests` test `UpdateTest` has been created to replace the separate tests of the same name in use by adaptors using `CoreTests` (the start of removing some of the fat from the tests).

Also, some preparation work on tests and DB object setup scripts for other adaptors has been included to path the way for further rationalisation of tests (via further migration to the central `CoreTests` suite).


benilovj:
Hi Mark,

Initial thoughts:

- there's too much going on in just one commit. The best way to move fast is to break things up into atomic commits, and even different pull requests, if the commits are unrelated
- there's not enough context about *why* the changes are being made. This is the most important thing in a year's time, in the case that we would need to revisit a decision made here

Would you be able to split things out and explain them a bit more?

MMatten:
Absolutely. It's got a bit muddied. 

I'll re-plan and run it past you here before creating new commits.

MMatten:
Hi Jake,

I think I'm ready to abandon this PR and decompose it into other branches and PRs all aimed towards resolving issue #23.

What's the process here? Do you close abandoned PRs or do I?

benilovj:
> Do you close abandoned PRs or do I?

Either is fine, really.

MMatten:
Abandoned - to be decomposed into more focused PRs.

PR #369, opened 2015-05-21
Forward default FitNesse port to allow FitNesse acccess from host

MMatten:
This PR is to allow host access to the FitNesse server running on the guest test VM.

I have had to add port fowarding to Vagrantfile in every previous test VM I've built hence I'm proposing this is added permanently.


benilovj:
:+1:

PR #371, opened 2015-05-25
Use CoreTests suite in Derby Flow and Standalone mode tests

MMatten:
Enhance test coverage for Derby and progress work on simplifying acceptance tests (issue #23).


FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/FlowMode/SetUp/content.txt
```diff
@@ -1,3 +1,9 @@
 !|dbfit.EmbeddedDerbyTest|
 
 !|Connect|jdbc:derby:memory:dbfitdb;create=true|
+
+|Execute Ddl|${CREATEUSERTABLE}|
+
+|Execute Ddl|${CREATEMAKEUSERPROC}|
+
+|Commit|
```
javornikolov: Is that `Commit` necessary here?

dbfit-java/derby/src/integration-test/java/dbfit/environment/DerbyRegressionSupportStoredProcs.java
```diff
@@ -0,0 +1,31 @@
...
+    public static void MakeUser() throws SQLException {
+        Connection conn = null;
+        PreparedStatement stmt = null;
+        ResultSet res = null;
+
+        try {
+            // "jdbc:default:connection" tells the DriverManager to use the existing connection.
+            conn = DriverManager.getConnection("jdbc:default:connection");
+            String sql = "INSERT INTO Users (Name, UserName) VALUES ('user1', 'fromproc')";
+            stmt = conn.prepareStatement(sql);
```
javornikolov: Can we use the [try-with-resources](https://github.com/dbfit/dbfit/blob/3b58e7255d78b81edcf3779bc9671ef2593c9e78/dbfit-java/netezza/src/main/java/dbfit/environment/NetezzaEnvironment.java#L69) syntax here (both for the statement and the connection)?

FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/FlowMode/SetUp/content.txt
MMatten: I'm using the `Commit` here to simulate having a permanent table in a DB where the structure (DDL) is persisted rather than created for each test.

This is to enable the tests, that include the slightly irregular `TransactionsTest`, to run properly. `TransactionsTest` includes a `Commit` and this makes commits the DDL and so the next test fails because of the `Execute Ddl` in the `SetUp`.

The `Rollback` in the `TearDown` undoes the DML that's executed in the test as would normally happen at the end of a flow mode test.

I then need to drop the DB objects and commit that drops so that any subsequent test's `SetUp` will run correctly.

It might be better to do something with the `TransactionsTest` test but I'm not sure what.

javornikolov: OK, it's fine as is. I was just curious.

dbfit-java/derby/src/integration-test/java/dbfit/environment/DerbyRegressionSupportStoredProcs.java
```diff
@@ -0,0 +1,31 @@
...
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+
+public class DerbyRegressionSupportStoredProcs {
+
+    public static void MakeUser() throws SQLException {
+        Connection conn = null;
+        PreparedStatement stmt = null;
+        ResultSet res = null;
```
javornikolov: This `ResultSet` is not used, also the related import.

javornikolov:
:+1: Looks OK to me now.

MMatten:
Cool. I thought the new test classes JAR task could be a point of discussion as it sets a convention. Have you guys done something similar before?

javornikolov:
The int-tests jar is indeed something I stared at for a while to figure out what it's doing. A bit similar thing is what we've done with `copyExtraLibs`.
One  observation is that now we end up with several empty jars (just MANIFEST inside); and the derby one only has a real class file inside.

MMatten:
Just checking situation with the integration test JARs again. I can only see one for Derby (`dbfit-derby-integration-test-3.1.0.jar`) when I run `./gradlew starthere` so I guess it's only producing an archive if there is at least one class file produced.

build.gradle
```diff
@@ -68,6 +68,21 @@ task copyExtraLibs(dependsOn: [copyLocal]) << {
...
+        from 'dbfit-java'
+        include '**/*integration-test*.jar'
+        exclude 'build/**/*.jar'
+        exclude proprietaryLibs()
+        exclude experimentalItems()
+        into 'dist/lib'
+        eachFile { details ->
+            details.path = flattenPath(details.path)
+        }
+        includeEmptyDirs = false
```
javornikolov: There is trailing whitespace here.

javornikolov:
> Just checking situation with the integration test JARs again. I can only see one for Derby (dbfit-derby-integration-test-3.1.0.jar) when I run `./gradlew starthere` so I guess it's only producing an archive if there is at least one class file produced (with Support in the class name).

When I tested I ran `./gradlew integrationTestJar` which explains why I got it for all projects: the `copyTestLibs` task invokes the derby part only.

Let's fix the trailing whitespace and squash the commits into one: then we can merge.

MMatten:
Whitespace removed and fix-ups squashed.

MMatten:
Ah, I guess I still need to rebase first.

PR #372, opened 2015-05-25
Use CoreTests suite for HSQLDB Standalone mode tests

MMatten:
Enhance test coverage for HSQLDB by introducing use of the `CoreTests` suite.

### DIFF
--- a/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/HsqlTests/StandaloneFixtures/SetUp/content.txt
+++ b/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/HsqlTests/StandaloneFixtures/SetUp/content.txt
@@ -1,5 +1,7 @@
-|dbfit.HSQLDBTest|
+|Import Fixture|
+|dbfit.fixture|
 
+|Database Environment|HSQLDB|
 |Connect|jdbc:hsqldb:mem:dbfit|
 
 |Execute Ddl|${CREATEUSERTABLE}|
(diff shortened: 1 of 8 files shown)


MMatten:
Hmmmm. CI failing with Java 7, passing with 8.

More testing required.

MMatten:
I'm still stumped. 

I can build/test this locally using JDK 1.7 and 1.8 without any issues.

@javornikolov - do you have any ideas about how to debug this? Is there anyway of getting to the FitNesse output on the build's VM (or does the VM get immediately destroyed)?

javornikolov:
I'm not sure how to read the test report files. Maybe should be copied somewhere as part of the build. Or a step can be added to the build to dump results on stdout...

javornikolov:
Just ran the build locally and hsqldb integration test failed. I'll check what's the trouble, I guess it's the same trouble on Travis CI.

javornikolov:
In all hsql flow mode tests:

`|Execute Ddl|CREATE TABLE users (name VARCHAR(20) NOT NULL, username VARCHAR(20) NOT NULL)|`

`Ambiguity between: fixture dbfit.fixture.ExecuteDdl AND executeDdl()	`

There are similar errors for all other calls too, but the above is the first (in SetUp).

FitNesseRoot/DbFit/AcceptanceTests/JavaTests/HsqlTests/StandaloneFixtures/SetUp/content.txt
```diff
@@ -0,0 +1,9 @@
+|Import Fixture|
```
javornikolov: I think this should be paired with following in `TearDown`:
```
!|dbfit.util.ExportFixture|
|dbfit.fixture|
```

This is might be leading to the errors raised during the CI build.

MMatten:
Thanks for finding this. 

I have pushed all commits from my local repo so no idea how I could get different results. 

I'll trash my VM and rebuild to verify.

MMatten:
I've destroyed everything on my local env and rebuilt the local repo and test VM.

I've still no idea why this runs fine for me without the `Export Fixture` commit.

Do we need to get to the bottom of this or can we consider merging?

javornikolov:
It's OK to merge that. (Maybe there is something with the test vm or just tests are running in different order there for some random reason).

javornikolov:
BTW, I've been running the build via `./gradlew fastbuild`.

PR #373, opened 2015-05-28
Re-enable working, but disabled, Postgres tests

MMatten:
Re-enable all disabled flow mode tests and they execute correctly.

The standalone mode test `InsertWithTableInRow` has been left as disabled as it fails and it's unclear whether the current version of DbFit is supposed to support the `Insert` fixture with the table specified in the second fixture table row rather than in the second cell of the first row.


MMatten:
Progresses work on resolving issue #1.

javornikolov:
:+1: Thanks, @MMatten! 

The `InsertWithTableInRow` is disabled for Oracle too.

PR #376, opened 2015-06-04
Support access to DB objects in connection's non default DB

MMatten:
Add support for access to objects that are not in the current database for the connection/session using the full 3 part name.

Users currently have to rely upon either the default database for the connection (specified for each DB user, or overridden for the connection using either the `database=` or `databaseName=` connection parameter), OR, set the default database using a `USE <db name>` SQL statement.

### DIFF
--- a/dbfit-java/sqlserver/src/main/java/dbfit/environment/SqlServerEnvironment.java
+++ b/dbfit-java/sqlserver/src/main/java/dbfit/environment/SqlServerEnvironment.java
@@ -69,7 +69,7 @@ public class SqlServerEnvironment extends AbstractDbEnvironment {
             throws SQLException {
         String qry = " select c.[name], TYPE_NAME(c.system_type_id) as [Type], c.max_length, "
                 + " 0 As is_output, 0 As is_cursor_ref "
-                + " from sys.columns c "
+                + " from " + objectDatabasePrefix(tableOrViewName) + "sys.columns c "
                 + " where c.object_id = OBJECT_ID(?) "
                 + " order by column_id";
         return readIntoParams(tableOrViewName, qry);
@@ -132,6 +132,15 @@ public class SqlServerEnvironment extends AbstractDbEnvironment {
     // private static string[] BinaryTypes=new string[] {"BINARY","VARBINARY", "TIMESTAMP"};
     // private static string[] GuidTypes = new string[] { "UNIQUEIDENTIFIER" };
     // private static string[] VariantTypes = new string[] { "SQL_VARIANT" };
+
+    private String objectDatabasePrefix(String dbObjectName) {
+        String objectDatabasePrefix = "";
+        String[] objnameParts = dbObjectName.split("\\.");
+        if (objnameParts.length == 3) {
+        	objectDatabasePrefix = objnameParts[0] + ".";
+        }
+        return objectDatabasePrefix;
+    }
 
     private static Direction getParameterDirection(int isOutput, String name) {
         if (name.isEmpty()) {
@@ -216,12 +225,12 @@ public class SqlServerEnvironment extends AbstractDbEnvironment {
                     + "       p.[name], TYPE_NAME(p.system_type_id) as [Type], "
                     + "       p.max_length, p.is_output, p.is_cursor_ref, "
                     + "       p.parameter_id, 0 as set_id, p.object_id "
-                    + "   from sys.parameters p "
+                    + "   from " + objectDatabasePrefix(procName) + "sys.parameters p "
                     + "   union all select "
                     + "        '' as [name], 'int' as [Type], "
                     + "        4 as max_length, 1 as is_output, 0 as is_cursor_ref, "
                     + "        null as parameter_id, 1 as set_id, object_id "
-                    + "   from sys.objects where type in (N'P', N'PC') "
+                    + "   from " + objectDatabasePrefix(procName) + "sys.objects where type in (N'P', N'PC') "
                     + ") as u where object_id = OBJECT_ID(?) order by set_id, parameter_id");
     }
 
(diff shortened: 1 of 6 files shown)


dbfit-java/sqlserver/src/main/java/dbfit/environment/SqlServerEnvironment.java
```diff
@@ -80,8 +81,28 @@ protected String parseCommandText(String commandText) {
...
+        if (objnameParts.length == 3) {
+            objectDatabaseName = objnameParts[0];
+        } else {
+            objectDatabaseName = currentDatabaseName;
+        }
+
         String bracketedName = enquoteAndJoin(objname.split("\\."), ".", "[", "]");
 
+        try (Statement s = currentConnection.createStatement()) {
+            s.execute("USE " + objectDatabaseName);
```
javornikolov: Hmm, I think this has side effect: the current database is changed; and if something dies - it will remain in the new database.

What about just modifying the query above from `sys.columns` to something like `objectDatabaseName.sys.columns`?

MMatten:
Ah. I've just noticed that I've not checked in the other bit of code that reverts the DB back to the value that I capture in `currentDatabaseName`! I was executing another statement to revert the current DB right after querying `sys.columns` so there was no chance of leaving a modified default DB in place for the remainder of the session.

However, your suggestion seems to work perfectly and is much more elegant (and I'm embarrassed I didn't think of it in the first place! :flushed:).

I'll run the tests over it.

MMatten:
Hmm. This does involve changing both `getAllProcedureParameters` and `getAllColumns` with the similar code but still pretty compact.

MMatten:
Also tidied up the DB setup script whitespace (we've no style guide for the SQL code - I generally use Steven Feuerstein-like formatting).

dbfit-java/sqlserver/src/main/java/dbfit/environment/SqlServerEnvironment.java
```diff
@@ -208,6 +216,13 @@ private static int getSqlType(String dataType) {
 
     public Map<String, DbParameterAccessor> getAllProcedureParameters(
             String procName) throws SQLException {
+
+        String objectDatabaseName = "";
+        String[] objnameParts = procName.split("\\.");
+        if (objnameParts.length == 3) {
+            objectDatabaseName = objnameParts[0] + ".";
+        }
```
javornikolov: That snippet is basically the same as the other. Maybe we can move it to a single place?

dbfit-java/sqlserver/src/main/java/dbfit/environment/SqlServerEnvironment.java
```diff
@@ -67,9 +68,16 @@ protected String parseCommandText(String commandText) {
 
     public Map<String, DbParameterAccessor> getAllColumns(String tableOrViewName)
             throws SQLException {
+
+        String objectDatabaseName = "";
+        String[] objnameParts = tableOrViewName.split("\\.");
+        if (objnameParts.length == 3) {
+            objectDatabaseName = objnameParts[0] + ".";
```
javornikolov: It sounds a bit odd at first glimpse to have something named `objectDatabaseName` to end with `.` (I wonder if we can figure out a better name).

MMatten: `objectDatabaseNameDot`?! Or I can just use a ternary operator (`+ objectDatabaseName + (objectDatabaseName == "") ? "" : ".")` to include the `.` separator.

Any views?

javornikolov: objectDatabasePrefix

MMatten:
Do you think it would be better to distribute the tests in the new page I've created into other, existing, test pages or to have an explicit test page for this?

javornikolov:
A separate test page seems OK to me.

dbfit-java/sqlserver/src/main/java/dbfit/environment/SqlServerEnvironment.java
```diff
@@ -133,6 +135,16 @@ protected String parseCommandText(String commandText) {
     // private static string[] GuidTypes = new string[] { "UNIQUEIDENTIFIER" };
     // private static string[] VariantTypes = new string[] { "SQL_VARIANT" };
 
+    private String objectDatabasePrefix(String dbObjectName) {
+
```
javornikolov: In the existing code seems we don't have such blank lines after the method header. And right now to me it looks a bit better without this extra line.

dbfit-java/sqlserver/src/main/java/dbfit/environment/SqlServerEnvironment.java
```diff
@@ -67,9 +68,10 @@ protected String parseCommandText(String commandText) {
 
     public Map<String, DbParameterAccessor> getAllColumns(String tableOrViewName)
             throws SQLException {
+
```
javornikolov: Similar comment about the blank line here.

dbfit-java/sqlserver/src/main/java/dbfit/environment/SqlServerEnvironment.java
```diff
@@ -14,6 +14,7 @@
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.sql.Statement;
```
javornikolov: Do we need this import?

MMatten:
Ok. Extra blank lines fixed.

javornikolov:
Now it looks good to me. Might be good to just to clean up the commit history and we can merge then.

MMatten:
I've squashed to a single commit and rebased.

Issue #1, opened 2013-01-29
Getting PostgreSQL tests to green


benilovj:

MMatten:
For the remaining Postgres test what's to be done. DbFit 1.1 used to support the table name in the second fixture table row (http://sourceforge.net/p/dbfit/code/HEAD/tree/tags/dbfit-1.x/impl/java/src/dbfit/fixture/Insert.java#l30) but there's no evidence of this being attempted in the 2.x code.

Do we delete the test (and the same in the Oracle tests) and formally drop this?

javornikolov:
Resolved by #373, #377

PR #377, opened 2015-06-06
Drop tests for Insert-with-table-name-in-second-table-row

MMatten:
After retesting this in DbFit 1.1 it seems this feature has never worked, despite the v1.1 code suggesting that this was supported.

Hence I'm proposing dropping the tests for this unimplemented feature.


MMatten:
Resolves issue #1.

javornikolov:
:+1:

PR #380, opened 2015-06-19
Bump up fitlibrary to final FitServer based version

MMatten:
JAR included only for upload to AWS.


javornikolov:
:+1:
The jar has been uploaded - I'll rebase to remove the commit with it before merging.

PR #384, opened 2015-07-04
Publish dbfit jars to central

### LINKED ISSUE #173: Push DbFit jars to Maven Central
Here's [how to do it with Gradle](http://jedicoder.blogspot.com/2011/11/automated-gradle-project-deployment-to.html).

MMatten:
To resolve #173.

`uploadArchives` task creates additional JAR files instead of using the existing JAR output libraries: -
* dbfit-x.y.z[-SNAPSHOT].jar
* dbfit-x.y.z-sources[-SNAPSHOT].jar
* dbfit-x.y.z-javadoc[-SNAPSHOT].jar

then signs the archives and uploads to the appropriate (SNAPSHOT or release staging) Sonatype Maven repository.

All dependencies are published in the associated POM file. The exception is FitLibrary as the version currently used by DbFit isn't available from public repositories (a point for discussion).

### DIFF
--- a/dbfit-java/build.gradle
+++ b/dbfit-java/build.gradle
@@ -4,16 +4,15 @@ ext {
 }
 
 allprojects {
+    apply plugin: 'java'
+    apply plugin: 'checkstyle'
     apply plugin: 'idea'
     apply plugin: 'eclipse'
-
     apply plugin: 'maven'
-    group = 'com.neuri.dbfit'
-    version = rootProject.dbfitVersion
-    
-    apply plugin: 'java'
+    apply plugin: 'signing'
 
-    apply plugin: 'checkstyle'
+    group = 'com.github.dbfit'
+    version = rootProject.dbfitVersion
 
     checkstyle {
         configFile = new File(parent.projectDir, "config/checkstyle/checkstyle.xml")
@@ -22,6 +21,17 @@ allprojects {
     tasks.withType(JavaCompile) {
         options.compilerArgs << "-Xlint:unchecked"
         options.compilerArgs << "-Xlint:deprecation"
+    }
+
+    // Suppress stricter Java 8 Javadoc checks.
+    if (JavaVersion.current().isJava8Compatible()) {
+        tasks.withType(Javadoc) {
+            options.addStringOption('Xdoclint:none', '-quiet')
+        }
+    }
+
+    dependencies {
+        compile "org.fitnesse:fitnesse:${fitNesseVersion}:standalone@jar"
     }
 }
 
@@ -62,7 +72,7 @@ subprojects {
         testCompile "org.hamcrest:java-hamcrest:2.0.0.0"
         testCompile "commons-io:commons-io:2.4"
         testRuntime 'com.dbdeploy:dbdeploy-ant:3.0M3'
-        integrationTestCompile "dummy:fitnesse-standalone:${fitNesseVersion}@jar"
+        integrationTestCompile "org.fitnesse:fitnesse:${fitNesseVersion}:standalone@jar"
     }
 
     task integrationTestJar(type: Jar) {
@@ -97,15 +107,18 @@ subprojects {
 task assembleAll(dependsOn: subprojects.assemble) { }
 task libs(dependsOn: subprojects.libs) { }
 
+def publicCoreCompileDeps = [
+    'org.reflections:reflections:0.9.9',
+    'commons-codec:commons-codec:1.10',
+    'org.apache.commons:commons-lang3:3.4',
+    'com.github.dbfit:fitlibrary:20091020'
+]
+
 project('core') {
     description = 'DBFit core api and fixtures'
 
     dependencies {
-        compile "dummy:fitnesse-standalone:${fitNesseVersion}@jar"
-        compile "dummy:fitlibrary:20091020@jar"
-        compile "org.reflections:reflections:0.9.9"
-        compile "commons-codec:commons-codec:1.10"
-        compile "org.apache.commons:commons-lang3:3.4"
+        compile publicCoreCompileDeps
         testCompile "org.mockito:mockito-core:${mockitoVersion}"
     }
 }
@@ -133,22 +146,30 @@ project('derby') {
     }
 }
 
+def publicHsqldbRuntimeDeps = [
+    'org.hsqldb:hsqldb:2.3.2@jar'
+]
+
 project('hsqldb') {
     description = 'DBFit HSQLDB engine'
 
     dependencies {
         compile project(':dbfit-java:core')
-        runtime 'org.hsqldb:hsqldb:2.3.2@jar'
+        runtime publicHsqldbRuntimeDeps
     }
 }
+
+def publicOracleCompileDeps = [
+    'commons-io:commons-io:2.4'
+]
 
 project('oracle') {
     description = 'DBFit Oracle engine'
 
     dependencies {
         compile project(':dbfit-java:core')
-        compile "commons-io:commons-io:2.4"
         compile 'dummy:ojdbc6:11.2.0.3.0@jar'
+        compile publicOracleCompileDeps
         testCompile "org.mockito:mockito-core:${mockitoVersion}"
     }
 
@@ -167,12 +188,16 @@ project('oracle') {
     }
 }
 
+def publicMysqlCompileDeps = [
+    'mysql:mysql-connector-java:5.1.6'
+]
+
 project('mysql') {
     description = 'DBFit MySQL engine'
 
     dependencies {
         compile project(':dbfit-java:core')
-        compile group: 'mysql', name: 'mysql-connector-java', version: '5.1.6'
+        compile publicMysqlCompileDeps
     }
 
     integrationTest.doFirst {
@@ -180,12 +205,16 @@ project('mysql') {
     }
 }
 
+def publicPostgresCompileDeps = [
+    'org.postgresql:postgresql:9.3-1101-jdbc41'
+]
+
 project('postgres') {
     description = 'DBFit PostGreSQL engine'
 
     dependencies {
         compile project(':dbfit-java:core')
-        compile group: 'org.postgresql', name: 'postgresql', version: '9.3-1101-jdbc41'
+        compile publicPostgresCompileDeps
     }
 
     integrationTest.doFirst {
@@ -198,7 +227,7 @@ project('sqlserver') {
 
     dependencies {
         compile project(':dbfit-java:core')
-        testRuntime 'com.microsoft:sqljdbc:4.1.5605@jar'
+        runtime 'com.microsoft:sqljdbc:4.1.5605@jar'
         integrationTestRuntime 'com.microsoft:sqljdbc:4.1.5605@jar'
     }
 }
@@ -222,3 +251,95 @@ project('netezza') {
     }
 }
 
+// Aggregated dependencies to include in the POM file when publishing.
+dependencies {
+    compile publicCoreCompileDeps
+    runtime publicHsqldbRuntimeDeps
+    compile publicOracleCompileDeps
+    compile publicMysqlCompileDeps
+    compile publicPostgresCompileDeps
+}
+
+// Remove the default artifact as there are no Java sources for this project.
+configurations.archives.artifacts.clear()
+
+task allJavadoc(type: Javadoc) {
+    source subprojects.collect { it.sourceSets.main.allJava }
+    destinationDir = new File(buildDir, 'doc/javadoc')
+    classpath = files(subprojects.collect { it.sourceSets.main.compileClasspath })
+}
+
+task allJavadocJar(type: Jar) {
+    baseName = 'dbfit'
+    classifier = 'javadoc'
+    from allJavadoc
+}
+
+task allSourcesJar(type: Jar) {
+    baseName = 'dbfit'
+    classifier = 'sources'
+    subprojects.each { from it.sourceSets.main.allSource }
+}
+
+task allJar(type: Jar, dependsOn: subprojects.build) {
+    baseName = 'dbfit'
+    subprojects.each { from it.configurations.archives.allArtifacts.files.collect { zipTree(it) } }
+}
+
+artifacts {
+    archives allJar, allJavadocJar, allSourcesJar
+}
+
+signing {
+    required { rootProject.isReleaseVersion && gradle.taskGraph.hasTask("uploadArchives") }
+    sign configurations.archives
+}
+
+uploadArchives {
+    repositories {
+        mavenDeployer {
+            beforeDeployment {
+                MavenDeployment deployment -> signing.signPom(deployment)
+            }
+
+            if (!project.hasProperty('ossrhUsername')) {
+                project.ext.set('ossrhUsername', '')
+            }
+
+            if (!project.hasProperty('ossrhPassword')) {
+                project.ext.set('ossrhPassword', '')
+            }
+
+            repository(url: "https://oss.sonatype.org/service/local/staging/deploy/maven2/") {
+                authentication(userName: ossrhUsername, password: ossrhPassword)
+            }
+
+            snapshotRepository(url: "https://oss.sonatype.org/content/repositories/snapshots/") {
+                authentication(userName: ossrhUsername, password: ossrhPassword)
+            }
+
+            pom.project {
+                name 'DbFit'
+                packaging 'jar'
+                description 'Database test automation for FitNesse'
+
+                url 'http://dbfit.github.io/dbfit/index.html'
+                scm {
+                    connection 'scm:git@github.com/dbfit/dbfit.git'
+                    developerConnection 'scm:git@github.com/dbfit/dbfit.git'
+                    url 'https://github.com/dbfit/dbfit.git'
+                }
+
+                licenses {
+                    license {
+                        name 'GNU General Public License, version 2'
+                        url 'https://www.gnu.org/licenses/old-licenses/gpl-2.0.en.html'
+                    }
+                }
+
+                developers {
+                }
+            }
+        }
+     }
+}
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -228,6 +228,31 @@ Please be aware that if you change any code whilst the `/.gradlew start` command
 
         dbfit$ ./gradlew starthere -PkeepTests
 
+#### Publishing to the Sonatype Central Repository (a.k.a. Maven Central)
+
+##### Uploading Development Version Snapshots
+
+* Create a new Sonatype user profile (`https://issues.sonatype.org/secure/Signup!default.jspa`).
+* Create a new support issue and register to upload DbFit (`com.github.dbfit`) artefacts.
+* Modify the DbFit version number string in the parent Gradle build script (in the DbFit development root directory). Ensure the version number string has the `-SNAPSHOT` suffix.
+* Upload the development version snapshot artefact to the Sonatype snapshots repository with:
+
+        dbfit$ ./gradlew uploadArchives
+
+##### Uploading Release Versions
+
+* Modify the DbFit version number string in the parent Gradle build script (in the DbFit development root directory). Ensure the version number string DOES NOT have the `-SNAPSHOT` suffix.
+* Create a public/private key pair for signing, cryptographically, of the build artefacts (i.e. DbFit JAR files). Follow the Sonatype guide (http://central.sonatype.org/pages/working-with-pgp-signatures.html).
+* Update `gradle.properties` in the DbFit development root directory with your public key, public key pass phrase and path to your secret keyring file.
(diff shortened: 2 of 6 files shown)


MMatten:
@benilovj , @javornikolov - I've tidied up the commits on this one now.

Any views on this PR yet?

javornikolov:
I wonder if we can have some of the changes not related to maven central publishing as separate PR (dependencies configuration refactoring, etc.)

dbfit-java/build.gradle
```diff
@@ -62,7 +72,7 @@ subprojects {
         testCompile "org.hamcrest:java-hamcrest:2.0.0.0"
         testCompile "commons-io:commons-io:2.4"
         testRuntime 'com.dbdeploy:dbdeploy-ant:3.0M3'
-        integrationTestCompile "dummy:fitnesse-standalone:${fitNesseVersion}@jar"
+        integrationTestCompile "org.fitnesse:fitnesse:${fitNesseVersion}:standalone@jar"
```
javornikolov: Sometimes (almost always) we're using an intermediate fitnesse build (not the official one published to maven). Would it be still possible to do that with the updated notation?

gradle.properties
```diff
@@ -0,0 +1,6 @@
+//signing.keyId=<publicKey>
```
javornikolov: Unless it's OK to publish these credentials publicly, I wouldn't keep them in a version controlled file. Would it be possible to have gradle.properties outside the project tree? Or if not - maybe keep a `*.sample` and add `gradle.properties` to `.gitignore`.

dbfit-java/build.gradle
```diff
@@ -222,3 +251,116 @@ project('netezza') {
...
+                }
+
+                licenses {
+                    license {
+                        name 'GNU General Public License, version 2'
+                        url 'https://www.gnu.org/licenses/old-licenses/gpl-2.0.en.html'
+                    }
+                }
+
+                developers {
```
javornikolov: Is the developers list mandatory?

dbfit-java/build.gradle
```diff
@@ -222,3 +251,116 @@ project('netezza') {
     }
 }
 
+// Aggregated dependencies to include in the POM file when publishing.
+dependencies {
```
javornikolov: I wonder if this can be somehow automatically collected from the projects' dependencies definition (if it's possible to `tag` public dependencies there). Just an idea, I'm not sure it's feasible/easy to achieve.

gradle.properties
MMatten: The idea was to leave this only as a template to show what needs to be done. I.e. As it is now and to never check in with real credentials. The credential values won't be accessed unless someone tries to publish anyway I believe. 

But we can remove this file if you think that is better.

javornikolov: OK, if it's just a template - would it be possible to publish without putting real credentials in this exactly file?

MMatten: I don't believe so. A legitimate Sonatype login and a published key pair is required.

dbfit-java/build.gradle
MMatten: I don't think so. Would you rather it was removed?

dbfit-java/build.gradle
MMatten: I think we can pull dev snapshots from the central repo as I've seen them on the Sonatype staging repo.

MMatten: BTW why do we use intermediate/snapshot versions of FitNesse? Can we use release versions instead?

dbfit-java/build.gradle
MMatten: I tried, very hard, to extend the dependency configurations of the parent `dbfit-java` project from it's child projects (`core`, etc) but I was always presented with null pointer exceptions so I had to fall back to specifying the same dependencies for the parent `dbfit-java` project and the child projects using the artefact lists. Intuitively this sounds like the wrong way of trying to inherit anyway.

I'm not sure what you mean by `tag`. What is that?

dbfit-java/build.gradle
MMatten: Regarding FitNesse dev snapshots - if we want to pin our dependencies to a specific FitNesse dev snapshot then I'm not sure that this is supported. I'd have to play with it. A release version would be better if possible.

MMatten:
Another point of discussion needs to be that more recent FitLibrary versions haven't been published (to Sonatype) as far as I can see. Therefore our dependencies can't be resolved only from public repositories which breaks the model.

One option would be to publish FitLibrary (the version we're using) under the `com.github.dbfit` groupId.

dbfit-java/build.gradle
MMatten: The other thing I'd tried was the iterate through the subprojects' dependency configurations and add them to the parent project's dependency configurations. As I remember it the iterating seemed to work but I was getting an exception (I think it was in POM or the signing stage but I can't quite remember).

gradle.properties
javornikolov: [Here](https://www.timroes.de/2013/09/22/handling-signing-configs-with-gradle/) I found a post related to keeping credentials in properties file. Alternatively it's perhaps possible in a similar way to load a groovy script where settings are being configured.

dbfit-java/build.gradle
javornikolov: Intermediate versions might be useful e.g. to pick some fixes. But maybe the release versions are usually good enough.

gradle.properties
MMatten: The settings in `gradle.properties` are only required when we run the `uploadArchives` task. Most of the time there is no need for the file to exist (unless one needs to configure stuff like we`b proxy settings to download gradle or artefacts from Maven Central). 

I think the simplest option is to add the file into `.gitignore` then and leave the existing checked in version an an illustration of what needs to be set in it when uploading. 

What do you think?

dbfit-java/build.gradle
MMatten: > I wonder if we can have some of the changes not related to maven central publishing as separate PR (dependencies configuration refactoring, etc.)

Certainly. I'm not sure what's not directly related to publishing though. Any ideas welcomed.

MMatten:
Summary of the oustanding discussion points so far: -
* is there anything that needs to be sliced off onto another PR?
* can we agree to use FitNesse release versions instead of dev/build snapshots?
* are we OK to include an example/sample `gradle.properties` to show how to set up Sonatype credentials (and a key pair for signing)?
* should/must the list of developers be removed from the POM generation?
* are we happy with the approach used to generate dependencies, in the POM file, for the aggregated JAR? (If not we could look to publish a fat JAR. I.e. `dbfit-X.Y.Z-standalone.jar` though it would be nice to not have to).
* are we OK to publish FitLibrary under `com.github.dbfit.fitlibrary`? Do we want a fork of it anyway, seeing as it appears to be dead?

MMatten:
My view on where we are with each of the above: -

> * is there anything that needs to be sliced off onto another PR?

I think all of the changes directly related to publishing the JARs.

> * can we agree to use FitNesse release versions instead of dev/build snapshots?

*I* think that *we* think that this is OK.

> * are we OK to include an example/sample gradle.properties to show how to set up Sonatype credentials (and a key pair for signing)?

I think a *sample* file would be OK.

> * should/must the list of developers be removed from the POM generation?

Your opinions required please.

are we happy with the approach used to generate dependencies, in the POM file, for the aggregated JAR? (If not we could look to publish a fat JAR. I.e. dbfit-X.Y.Z-standalone.jar though it would be nice to not have to).

Your opinions required please.

> * are we OK to publish FitLibrary under com.github.dbfit.fitlibrary? Do we want a fork of it anyway, seeing as it appears to be dead?

If we don't fork and publish then a fat JAR is the only option really. Are there any issues with that relating to breaking people's class paths by introducing additional and potentially conflicting JAR versions?

javornikolov:
> are we OK to include an example/sample gradle.properties to show how to set up Sonatype credentials (and a key pair for signing)?

Yes, let's include one but with with a different name. (We can rebase onto master first - it already has `gradle.properties` in .gitignore).


> should/must the list of developers be removed from the POM generation?

I think we can try removing them for now. There is a list to github anyway.

> can we agree to use FitNesse release versions instead of dev/build snapshots?

Seems OK at the moment.


> are we OK to publish FitLibrary under com.github.dbfit.fitlibrary? Do we want a fork of it anyway, seeing as it appears to be dead?

Whatever would work best now. I think it's OK to fork if needed to get approval for publishing the required fitlibrary version to a public repo.


> Are we happy with the approach used to generate dependencies, in the POM file, for the aggregated JAR? (If not we could look to publish a fat JAR. I.e. dbfit-X.Y.Z-standalone.jar though it would be nice to not have to).

Let's try to get a `thin` version. If needed we can add a `fat` one in addition to it later.


The rest looks fine to me.

MMatten:
Great.

For:

>> are we OK to publish FitLibrary under com.github.dbfit.fitlibrary? Do we want a fork of it anyway, seeing as it appears to be dead?

> Whatever would work best now. I think it's OK to fork if needed to get approval for publishing the required fitlibrary version to a public repo.

FitLibrary seems to have been abandoned. Very surprising seeing as how prominent it was it the FitNesse world. I tried contacting Darren a while ago but no response. I don't think we're likely to hear from him or Rick but I could try him too. For a thin JAR we'll need all dependencies resolvable from public artefact repos. 

FitLibrary we'd be to published under the fitnesse groupId up until 20080822. 

Should I persevere with Rick and Darren or fork FitLibrary 20091020 from SourceFirge under the DbFit GitHub organisation?

MMatten:
Interestingly v1 seems to also have been published:

http://mvnrepository.com/artifact/net.sf.dbfit/dbfit/1.1.2.2

javornikolov:
> Should I persevere with Rick and Darren or fork FitLibrary 20091020 from SourceFirge under the DbFit GitHub organisation?

I guess we can do both. If we get support for publishing under the original domain - we can switch to it (and abandon our fork).

MMatten:
A query for a potential later PR....

So, we'll publish the DbFit executables as a single JAR. In that case what value is there in maintaining the current gradle project structure and build targets?

javornikolov:
> So, we'll publish the DbFit executables as a single JAR. In that case what value is there in maintaining the current gradle project structure and build targets?

Hmm. Should it be single JAR or we could keep them separate?

MMatten:
> Hmm. Should it be single JAR or we could keep them separate?

I'm not sure of the benefits of the existing model. More JARs = more hassle, or just smaller footprint?

Would anyone ever mix versions of DbFit JARs?

javornikolov:
I can imagine: smaller footprint, ability to swap just a part with customized/different version, multiple files can be copied in parallel. I don't say multiple jars is necessarily better, I'm not sure (up to know it hasn't been a huge problem).

I got wondering about the possible use cases/benefits of having dbfit on maven central? Do we need to take care about the binary utility scripts?

MMatten:
My main reason for working on this issue was so that I could create a PR against the FitNesse project to help detect breaking changes like the ill-fated PR 281. 

Im not sure why the DbFit 2.0 build switched to having multiple JARs. Maybe driven by the introduction of Maven but guessing really. 

The published 1.1 versions were a single JAR. 

Do you the encryption utils? Could we bundle them as resources in the JAR? Not perfect.

MMatten:
Some updates applied: -
* FitLibrary-20091020 is now available on Maven Central under the `com.github.dbfit` groupId (http://central.maven.org/maven2/com/github/dbfit/fitlibrary/20091020/).
* Developer details removed (but it looks like the `<developers>` section is still required in the POM by Sonatype).

MMatten:
I've also updated the version number for DbFit so I can get uploading dev snapshots. I guessed at `3.2.0`. Is that next (some enhancenents have been added) or would it be a patch release?

MMatten:
There have been some user requests too:

https://groups.google.com/forum/m/#!searchin/dbfit/Maven$20central/dbfit/9-IE8ieuWLU

MMatten:
Also now renamed the skeleton gradle properties file.

javornikolov:
Cool!

`3.2.0` is OK for the version number.

> I'm not sure why the DbFit 2.0 build switched to having multiple JARs.

I think that's just the default and simplest in case of multiple sub-projects, also seems better in terms of modularity. But it should be OK to have a single jar for Maven Central (as is) if that makes things simpler.

> Do you the encryption utils? Could we bundle them as resources in the JAR? Not perfect.

Just the Java classes (as is) looks to be enough to me for now. We're publishing dbfit as a library.

MMatten:
Great!

I think I've no more updates to make then unless you spot anything else. 

You can use it to publish 3.2.0 when you release it.

dbfit-java/build.gradle
```diff
@@ -222,3 +251,114 @@ project('netezza') {
...
+    compile publicCoreCompileDeps
+    runtime publicHsqldbRuntimeDeps
+    compile publicOracleCompileDeps
+    compile publicMysqlCompileDeps
+    compile publicPostgresCompileDeps
+}
+
+// Remove the default artifact.
```
javornikolov: What is the `default artifact` and why do we want to remove it?

MMatten: The default artefact (artifact) for this project is `dbfit-java-X.Y.Z.jar` which has no class files in it as there are no Java source files in this project. 

The artefact would get signed and published, by default, along with the additional aggregate JARs we create. I can enhance the comments if needed.

javornikolov: I wonder if we could identify the `default` artifact in a bit more straightforward way (now we rely it's at the end of the list, has 'jar' in the name, etc.). I don't know if it's possible.

Or if using the same logic - maybe can be implemented a bit shorter, e.g. something like:
``` groovy
remove(configurations.archives.artifacts.archives.reverse.find { it.getType() == 'jar' })
```

MMatten: > now we rely it's at the end of the list

We could simply remove every artifact:

`configurations.archives.artifacts.with { archives ->
    archives.each {
        remove(it)
    }
}`

Do we need to even check the file type/name at all?

If we don't mind relying on the configuration only ever having one default artefact then yes let's condense to a *one-liner*.

In your snippet, how does the `reverse.find` work? What does it do?

MMatten: Ok, so `reverse()` is just to get at the last element in the `List` and `find()` finds the first matching element. So both of these mean that this will work only if the last default artefact is a JAR (which it would be for our specific case).

Or shall we just remove all of the default artefacts?

javornikolov: After some digging into forums (perhaps I should read the docs at some point :-)):
* Looks like removing all defaults artifacts would give the same result right now: `configurations.archives.artifacts.clear()`
* If a need emerges to have non-JAR which we need to keep:
`configurations.archives.artifacts.removeAll { it.archiveTask.is jar }`

MMatten: > perhaps I should read the docs at some point :-)):

Me too. There does seem to be endless ways in which to do any one thing in Gradle. Very flexible but having not really sat down to learn it properly yet it feels a bit *hit and miss* still.

> `configurations.archives.artifacts.clear()`

Looks good to me!

MMatten:
I've applied the `clear()` method now.

Anything else to address?

dbfit-java/build.gradle
```diff
@@ -97,15 +107,18 @@ subprojects {
 task assembleAll(dependsOn: subprojects.assemble) { }
 task libs(dependsOn: subprojects.libs) { }
 
+def publicCoreCompileDeps = [
+    'org.reflections:reflections:0.9.9'
+  , 'commons-codec:commons-codec:1.10'
```
javornikolov: Can we have the commas at the end of lines.

MMatten: Ok. Coming up.

MMatten:
List element separator (",") style updated now for publicly available lib dependencies.

dbfit-java/build.gradle
```diff
@@ -222,3 +251,106 @@ project('netezza') {
...
+
+                licenses {
+                    license {
+                        name 'GNU General Public License, version 2'
+                        url 'https://www.gnu.org/licenses/old-licenses/gpl-2.0.en.html'
+                    }
+                }
+
+                developers {
+		}
```
javornikolov: There are some tabs here, should be spaces.

dbfit-java/build.gradle
```diff
@@ -222,3 +251,106 @@ project('netezza') {
...
+
+// Remove the default artifact as there are no Java sources for this project.
+configurations.archives.artifacts.clear()
+
+task allJavadoc(type: Javadoc) {
+    source subprojects.collect { project ->
+        project.sourceSets.main.allJava
+    }
+    destinationDir = new File(buildDir, 'doc/javadoc')
+    classpath = files(subprojects.collect { project ->
```
javornikolov: What about using the default `it` param here and in the `source` (the definitions would be able to fit in single lines).

MMatten:
Horizontal tabs removed and I've replaced the iterator with the default `it`.

Do you think the code should be squashed to fewer lines still?

dbfit-java/build.gradle
```diff
@@ -222,3 +251,106 @@ project('netezza') {
...
+    baseName = 'dbfit'
+    classifier = 'sources'
+    subprojects.each {
+        from it.sourceSets.main.allSource
+    }
+}
+
+task allJar(type: Jar, dependsOn: subprojects.build) {
+    baseName = 'dbfit'
+    subprojects.each { subproject ->
```
MMatten: I'm not sure if I can replace this iterator with `it` because of use in the inner iteration.

MMatten: Well, it does seem to work (see below) but is it (use of `it` in outer and inner iterations) too confusing?

```groovy
task allJar(type: Jar, dependsOn: subprojects.build) {
    baseName = 'dbfit'
    subprojects.each {
        from it.configurations.archives.allArtifacts.files.collect {
            zipTree(it)
        }
    }
}

javornikolov:
> and I've replaced the iterator with the default it.
> 
> Do you think the code should be squashed to fewer lines still?

Yes, it looks a bit better to me if we squash these to one-liners.

MMatten:
Converted to one liners now. What do you think of the formatting?

javornikolov:
> Converted to one liners now. What do you think of the formatting?

It looks good to me now :+1:

javornikolov:
@MMatten - can we try to publish an initial snapshot version to Maven Central from this branch?

MMatten:
Yes, no problem.

I can do it (I already have a Sonatype login) or would you like to try (so that we can prove the process)?

You can create a login instantly. You'll need to create a support issue in order to get upload access to the `com.github.dbfit` groupId. They turn around support requests very quickly.

MMatten:
You'll need to create your own public/private key pair too and upload it to a public key server:

http://central.sonatype.org/pages/working-with-pgp-signatures.html

javornikolov:
OK, I can try to register tonight.

javornikolov:
I published via `uploadArchives`. Now I guess the artifacts are on the staging site, I'm trying to examine how to do the publishing to the central site: http://central.sonatype.org/pages/releasing-the-deployment.html. But I cannot see our project in the list of staging repositories. @MMatten, any idea what needs to be done next?

MMatten:
I can see the `3.2.0-SNAPSHOT` snapshot JARs in the search results.

Let me take a look...

MMatten:
Have you removed the `-SNAPSHOT` from the version number in `build.gradle` and uploaded a `3.2.0` set of JARs too?

javornikolov:
No, I haven't changed the version number - it's not for final release yet. I just wondered if we should publish the snapshot to the target repo, but if it's not possible to delete/replace later - it doesn't seem a good idea.

MMatten:
Ok. Cool.

Once you uploaded the release version you should be able to publish it manually via the UI.

You should be able to automatically download the artifacts from the snapshot repo though if you wanted to test including that in a test build.

MMatten:
I've just spotted that I've not updated the publicly available dependencies to include `com.github.dbfit.fitlibrary-20091020.jar`.

I'll do that now.

javornikolov:
Cool, I'm merging. Thank you @MMatten for this great contribution!

javornikolov:
BTW, there is one special think to consider - JDK version used to build the archives matters (I'm not sure if there is easy way to maintain e.g. both jvm7 and jvm8 artifacts).

MMatten:
Won't the archives be usable with Java 7 when built with Java 8 as long as the compatibility was set to 7 (1.7)?

Also, I'm trying to find out what the minimum Java version for FitNesse is. I can't find any info on the web site but Fitnesse starts up OK with Java 6.

If I create that PR against the FitNesse project, to include basic/smoke testing of DbFit, there will be issues if they're still using Java 6 for their CI.

Should we try to stay aligned with them and potentially go back to supporting Java 6!?

Issue #330, opened 2014-07-01
add getAutoCommit() check before setAutoCommit


struesda:
I was trying to create a separate pull request - but then saw you had done #327 already.

In my fork - I also added calls to getAutoCommit() around setAutoCommit in AbstractDbEnvironment.   This is a standard JDBC method and should make it easier to handle non-transactional databases.

However - I am not familiar enough with git to know how to make a new pull request from my fork with just these changes.

At the moment I have manually removed the Informix changes from my fork - but am still trying to figure out how to create a new pull request.

Let me know the best way to proceed.

javornikolov:
Closing this as it's supposed to be possible to achieve the desired behavior with the functionality added with #385.

PR #385, opened 2015-07-05
Auto-commit support

### LINKED ISSUE #383: Transactions error for Teradata Stored Procedure with DDL
There is a problem with 3.10 when I try to test Teradata stored procedure with several DDL statements inside. 
How it can be reproduced:
SQL for Procedure:
```sql
REPLACE PROCEDURE IssueWithTransactions(out sampleout char(4))
begin
	DECLARE sql_query VARCHAR(4000);
	SET sql_query='create table iwt_table1 (fakefield integer)';
	call DBC.SysExecSQL (sql_query);
	SET sql_query='alter table iwt_table1 add fakefield2 integer';
	call DBC.SysExecSQL (sql_query);
	SET sql_query='alter table iwt_table1 add fakefield3 integer';
	call DBC.SysExecSQL (sql_query);
	SET sampleout='true';
end;
```
And here is DbFit code for test:
Call procedure
!|Execute Procedure|dftest.IssueWithTransactions|
|sampleout?|
|true|

Result is java exception:
java.sql.SQLException: [Teradata Database] [TeraJDBC 15.10.00.05] [Error 3932] [SQLState 25000] ISSUEWITHTRANSACTIONS:Only an ET or null statement is legal after a DDL Statement.

Connection mode: /TMODE=TERA
Exception tells us that transaction was started and must be finished after first DDL to keep possibility to run next DDL. 
This is Teradata specifics and DbFit perfectly handles it for "Execute Ddl" fixture.

**Here is some diagnostics**
If I modify stored procedure and add "ET;" statement:
```sql
REPLACE PROCEDURE IssueWithTransactions(out sampleout char(4))
begin
	DECLARE sql_query VARCHAR(4000);
	SET sql_query='create table iwt_table1 (fakefield integer)';
	call DBC.SysExecSQL (sql_query);
	ET;	
	SET sql_query='alter table iwt_table1 add fakefield2 integer';
	call DBC.SysExecSQL (sql_query);
	SET sql_query='alter table iwt_table1 add fakefield3 integer';
	call DBC.SysExecSQL (sql_query);
	SET sampleout='true';
end;
```
then DbFit test works fine, because I explicitly close transaction. But the thing is - I didn't started that transaction. So if I try to execute this procedure outside of DbFit (from Teradata SQL Assistant for example) it will fail with message that I try to End Transaction which wasn't started. 

I'm not a java developer, but it looks for me that Execute Procedure fixture explicitly starts transaction in any case. Even if it's not needed.
Or - other suggestion - autocommit mode is explicitly off. But for /TMODE=TERA connection mode autocommit must work (that is the major point of this mode).
Teradata JDBC adapter docs (https://developer.teradata.com/doc/connectivity/jdbc/reference/current/jdbcug_chapter_2.html) tells that for TERA mode it will work in following way:
_If the caller's session has Teradata transaction semantics, then the Java XSP's default connection Auto-Commit setting will depend on whether the caller's session is within an existing transaction (for example, BT was executed). In this case, the Auto-Commit setting will be false if BT was executed, and the Auto-Commit setting will be true if BT was not executed._

So it looks (suggestion) that BT (Begin Transaction) is fired before executing procedure.
But for Teradata it would be logical to check connection mode and (ANSI vs TERA) and then leave transaction semantics to Teradata.

javornikolov:
Add support to control transaction commit/rollback
  - [x] SetAutoCommit true|false fixture
  - [x] Rollback, Commit fixtures to unify Standalone vs Flow Mode syntax and tests
  - [x] Update Documentation

This intends to resolve #383

References to another issue related to auto-commit mode: #326, #330

### DIFF
--- a/dbfit-java/core/src/main/java/dbfit/fixture/SetOption.java
+++ b/dbfit-java/core/src/main/java/dbfit/fixture/SetOption.java
@@ -1,0 +1,63 @@
+package dbfit.fixture;
+
+import dbfit.api.DBEnvironment;
+import dbfit.api.DbEnvironmentFactory;
+import dbfit.util.NameNormaliser;
+import dbfit.util.Options;
+import static dbfit.util.Options.OPTION_AUTO_COMMIT;
+
+import fit.Fixture;
+import fit.Parse;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+
+public class SetOption extends Fixture {
+    private DBEnvironment dbEnvironment;
+    private String option;
+    private String value;
+
+    public SetOption() {
+        dbEnvironment = DbEnvironmentFactory.getDefaultEnvironment();
+    }
+
+    public SetOption(DBEnvironment env, String option, String value) {
+        this.dbEnvironment = env;
+        this.option = NameNormaliser.normaliseName(option);
+        this.value = value;
+    }
+
+    public void setOption() throws SQLException {
+        Options.setOption(getOption(), getValue());
+        setEnvironmentOption();
+    }
+
+    @Override
+    public void doRows(Parse rows) {
+        try {
+            setOption();
+        } catch (Throwable e) {
+            throw new Error(e);
+        }
+    }
+
+    private String getOption() {
+        if (option == null) {
+            option = NameNormaliser.normaliseName(args[0]);
+        }
+        return option;
+    }
+
+    private String getValue() {
+        if (value == null) {
+            value = args[1];
+        }
+        return value;
+    }
+
+    private void setEnvironmentOption() throws SQLException {
+        if (OPTION_AUTO_COMMIT.equals(getOption()) && dbEnvironment != null) {
+            dbEnvironment.setAutoCommit();
+        }
+    }
+}
--- a/website/_includes/manual/transaction-control.md
+++ b/website/_includes/manual/transaction-control.md
@@ -4,7 +4,22 @@ By default, each individual test (FitNesse page) in flow mode is executed in a t
 
 If in flow mode, you can use the Commit and Rollback commands to control the transactions manually, but remember that a final rollback will be added at the end of the test. These commands have no additional arguments.
 
-In standalone mode, you will probably control transactions from outside DbFit. Utility commands to commit and rollback are still provided, if you need them, as part of the DatabaseEnvironment fixture. For example, use this table to rollback:
+In standalone mode, you will probably control transactions from outside DbFit. Utility commands to commit and rollback are still provided, if you need them. For example, use this table to rollback:
 
-    !|DatabaseEnvironment|
     |Rollback|
+
+Or in order to commit:
+
+    |Commit|
+
+### Autocommit mode
+
+By default DbFit is using `autocommit=false` mode - which means that the database statements are executed in one transaction. In some special occasions, you may need to change that. It's possible to configure database connection's autocommit mode via:
+
+    |set option|autocommit|true|
+
+Possible values are:
+
+ * `true` - automatically commit after each statement
+ * `false` (this is the default) - transaction is terminated on explicit commit or rollback (for flow mode such is performed at the end of the transaction).
+ * `auto` - driver default (DbFit doesn't explicitly set autocommt mode, the JDBC driver decides)
--- a/website/_includes/manual/transaction-control.md
+++ b/website/_includes/manual/transaction-control.md
@@ -12,6 +12,11 @@ Or in order to commit:
 
     |Commit|
 
+For standaone mode only, an alternative syntax is also still supported:
+
+    !|DatabaseEnvironment|
+    |Rollback|
+
 ### Autocommit mode
 
 By default DbFit is using `autocommit=false` mode - which means that the database statements are executed in one transaction. In some special occasions, you may need to change that. It's possible to configure database connection's autocommit mode via:
--- a/FitNesseRoot/DbFit/AcceptanceTests/CoreTests/CommonSuite/CommitAndRollback/content.txt
+++ b/FitNesseRoot/DbFit/AcceptanceTests/CoreTests/CommonSuite/CommitAndRollback/content.txt
@@ -1,3 +1,5 @@
+!|set option|autocommit|false|
+
 !|Execute Ddl|create table TESTTBL(ID int, VAL int)|
 
 !|Commit|
@@ -10,6 +12,20 @@
 
 !|Query|select ID, VAL from TESTTBL|
 |ID    |VAL?|
+
+-----
+
+!|set option|autocommit|true|
+
+!|Insert|TESTTBL|
+|ID     |VAL    |
+|1      |2      |
+
+!|Rollback|
+
+!|Query|select ID, VAL from TESTTBL|
+|ID    |VAL?|
+|1     |2   |
 
 !|Execute Ddl|drop table TESTTBL|
 !|Commit|
--- a/FitNesseRoot/DbFit/AcceptanceTests/CoreTests/CommonSuite/CommitAndRollback/content.txt
+++ b/FitNesseRoot/DbFit/AcceptanceTests/CoreTests/CommonSuite/CommitAndRollback/content.txt
@@ -1,0 +1,15 @@
+!|Execute Ddl|create table TESTTBL(ID int, VAL int)|
+
+!|Commit|
+
+!|Insert|TESTTBL|
+|ID     |VAL    |
+|1      |2      |
+
+!|Rollback|
+
+!|Query|select ID, VAL from TESTTBL|
+|ID    |VAL?|
+
+!|Execute Ddl|drop table TESTTBL|
+!|Commit|
--- a/FitNesseRoot/DbFit/AcceptanceTests/CoreTests/CommonSuite/CommitAndRollback/properties.xml
+++ b/FitNesseRoot/DbFit/AcceptanceTests/CoreTests/CommonSuite/CommitAndRollback/properties.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<properties>
+<Edit/>
+<Files/>
+<Properties/>
+<RecentChanges/>
+<Refactor/>
+<Search/>
+<Test/>
+<Versions/>
+<WhereUsed/>
+</properties>
--- a/dbfit-java/core/src/main/java/dbfit/DatabaseTest.java
+++ b/dbfit-java/core/src/main/java/dbfit/DatabaseTest.java
@@ -155,8 +155,8 @@ public class DatabaseTest extends Fixture {
         return new dbfit.fixture.CompareStoredQueriesHideMatchingRows(environment, symbol1, symbol2);    	
     }
 
-    public void setOption(String option, String value) {
-        Options.setOption(option, value);
+    public Fixture setOption(String option, String value) {
+        return new dbfit.fixture.SetOption(environment, option, value);
     }
 }
 
--- a/dbfit-java/core/src/main/java/dbfit/DatabaseTest.java
+++ b/dbfit-java/core/src/main/java/dbfit/DatabaseTest.java
@@ -42,22 +42,18 @@ public class DatabaseTest extends Fixture {
 
     public void connect(String dataSource, String username, String password, String database) throws SQLException {
         environment.connect(dataSource, username, password, database);
-        environment.getConnection().setAutoCommit(false);
     }
 
     public void connect(String dataSource, String username, String password) throws SQLException {
         environment.connect(dataSource, username, password);
-        environment.getConnection().setAutoCommit(false);
     }
 
     public void connect(String connectionString) throws SQLException {
         environment.connect(connectionString);
-        environment.getConnection().setAutoCommit(false);
     }
 
     public void connectUsingFile(String filePath) throws SQLException, IOException, FileNotFoundException {
         environment.connectUsingFile(filePath);
-        environment.getConnection().setAutoCommit(false);
     }
 
     public void close() throws SQLException {
@@ -121,12 +117,10 @@ public class DatabaseTest extends Fixture {
 
     public Fixture rollback() {
         return new dbfit.fixture.Rollback(environment);
-        environment.getConnection().setAutoCommit(false);
     }
 
     public Fixture commit() {
         return new dbfit.fixture.Commit(environment);
-        environment.getConnection().setAutoCommit(false);
     }
 
     public Fixture queryStats() {
--- a/dbfit-java/core/src/main/java/dbfit/DatabaseTest.java
+++ b/dbfit-java/core/src/main/java/dbfit/DatabaseTest.java
@@ -119,15 +119,13 @@ public class DatabaseTest extends Fixture {
     //      return new TestData(environment, type);
     //  }
 
-    public void rollback() throws SQLException {
-        // System.out.println("Rolling back");
-        environment.rollback();
+    public Fixture rollback() {
+        return new dbfit.fixture.Rollback(environment);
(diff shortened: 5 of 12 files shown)


MMatten:
Do you think this could / should be an option rather than a new fixture?

MMatten:
What about initialising this to false to minimise impact on existing users test suites?

javornikolov:
> Do you think this could / should be an option rather than a new fixture?

What we currently have via `setOption` (and `dbfit.util.Options`) are `passive` name-value pairs of settings. For connection auto-commit - we're altering the current database connection. I'm not sure about having options with active behavior. Also the scope of `Options` is maybe a bit different (for FlowMode it's reset per test page; for standalone - I'm not quite sure).

Maybe we can add such option which would affect connections which are opened after the option is set.

Another alternative might be to `set session option` for setting various options per the session/connection.

javornikolov:
> What about initialising this to false to minimise impact on existing users test suites?

I believe it's false by default - it's set in `AbstractDbEnvironment.afterConnectionEstablished`.

MMatten:
May be a change for another day but as DatabaseTest and DatabaseEnvironment extend SequenceFuxture I think it should be possible to add methods to AbstractDbEnvironmment and wrap the environment object as the system under test, instead of creating new Fixture classes. I think DbFit Java used to work this way until it hit a FitLibrary issue which might have been fixed since.

MMatten:
Our of interest Fitsharp's version has Command Timeout option which is a similar consideration - affects session characteristics. 

I guess either works fine.

MMatten:
Hmmmm. I guess the Fixture classes are only really needed for StandAlone mode when the user could set the default environment object in DbEnvironmentFactory and then use the fixtures without having to have first used a DatavaseEnvironment fixture. In flow mode the methods could live on the Environment and be exposed via DatabaseTest's SUT object. 

I think I get it now. Sorry for the rambling!

MMatten:
I think adding new Commit and Rollback fixtures isn't necessarily the right thing to do. We already have these from DatabaseEnvironment in standalone mode.

javornikolov:
> I think adding new Commit and Rollback fixtures isn't necessarily the right thing to do. We already have these from DatabaseEnvironment in standalone mode.

Right, but invoking commit/rollback was a bit different in standalone vs flow mode and the test page couldn't work in both modes.

MMatten:
Because the fixturing is different between the modes I think there has to be separate test pages. 
Perhaps a test reorg would be better to improve reuse of test page content somehow. 
Perhaps a CommonSuite (universal), FlowModeCommonSuite (common to all DBs flow mode tests) and StandAloneModeCommonSuite (common to all DB's stand alone mode tests). [I was think of something similar recently to allow us to get ConnectUsingFile tests back into the build].

javornikolov:
> Because the fixturing is different between the modes I think there has to be separate test pages.

In general, yes - there are some differences between flow and standalone mode. Often the different parts can be isolated into `SetUp`, `TearDown`  and the main test page to be reused - and SetUp, TearDown are anyway not in the common suite.

In the particular case - I kind of tried to unify syntax and now the page works in both standalone and flow mode. (Transaction control is not something typical for flow mode but some edge cases are emerging).


> I was think of something similar recently to allow us to get ConnectUsingFile tests back into the build]

Hmm, DbFit test connections now are now initialized via `ConnectUsingFile` (with exception of the embedded databases).

MMatten:
It kind of feels like we're changing the application to suit the tests rather than build the appropriate tests. 

Transaction control is the awkward case for reusing test content even within FlowMode for a single adaptor. Eg in PR #371 I had to do something a bit hacky to cater for TransctionsTest.

javornikolov:
I'll try to see how would it look like if auto-commit support is for standalone mode only.

MMatten:
Or what about allowing this to be set via a `Database Environment` fixture method, for standalone mode?

javornikolov:
> Or what about allowing this to be set via a Database Environment fixture method, for standalone mode?

Yes, sounds possible.

> I think this is one of those cases where it's simpler to have two test pages (one for each mode).

Do you mean separate tests for autoCommit=false and for autoCommit=true? Or for standalone and flow mode (in case we add support flow mode at all)?

> If we take the approach of new Commit and Rollback fixtures then we end up with duplicated ways of committing/rolling back and would we start directing people to using these instead of Database Environment? That was my concern.

Good point. On the other hand - having different syntax is causing some inconvenience when converting or sharing a test between modes. But if we move auto-commit as method in `Database Environment`, it's anyway for standalone mode only and no need to bother about compatibility with flow mode syntax.

MMatten:
> Also the scope of Options is maybe a bit different (for FlowMode it's reset per test page; for standalone - I'm not quite sure

I guess that could have been an oversight in the original code as there's no documentation to say otherwise.

MMatten:
> I think this is one of those cases where it's simpler to have two test pages (one for each mode).
>> Do you mean separate tests for autoCommit=false and for autoCommit=true? Or for standalone and flow mode (in case we add support flow mode at all)?

I meant separate test pages for Flow and StandAlone modes.

MMatten:
Thinking about the new Commit and Rollback fixtues - I think it's a good enhancement. We just need to document the new alternative ways of committing and rolling back in StandAlone mode.

May be we should push this even further some time by creating a standalone `Connect` fixture too.

MMatten:
For the `Set Auto Commit` option this still feels more like a `|Set Option|Auto Commit|true/false|` to me but that's only a personal preference anyway.

javornikolov:
I'll rebase on top of master to pick the refactored Options class. I'll try to implement that the feature as an option `|setOption|autocommit|`. (Seems that would having SetOption as a fixture).

MMatten:
Or I guess we can add thin methods to both DatabaseTest and DatabaseEnvironment and use SequenceFixture's feature of handling these like fixtures. 

Adding a new SetOption fixture would allow users to not require using a DatabaseEnvironment fixture at all in StandAlone mode (once they'd created the default environment in DbEnvironmentFactory's instance). I wonder if anyone actually uses DbFit like that though.

MMatten:
More thoughts...If this was a new (passive) option (instead of modifying the connection immediately) then that would mean we'd really have to centralise all statement execution into AbstractDbEnvironment and set the auto commit before every execution. I guess that could be a larger change.

MMatten:
Ok, so I think I've finally come around to your original idea. May be an active option isn't best for this. An external process, in stand alone mode, might modify the auto commit property anyway and the user might expect an option value to be in effect until they explicitly change it via another Set Option.

Then again...may be that would just need to be documented. 

Then again....does anyone actually use standalone mode like this!?

May be I'm over complicating this all.

What are your current thoughts?

javornikolov:
I just pushed ability to set this via `Set Option` (without removing the original SetAutoCommit fixture). Tests are not updated yet.
We can experiment a bit and see. The main point in SetOption version is to allow setting it before connection  - to "true", "force" or "auto" mode (I think some databases don't support calling setAutoCommit, e.g. Informix).

But I'm facing an issue with autoCommit in Derby (rollback after insert in auto-commit mode somehow roll-backs what has been inserted).

javornikolov:
The problematic case (in standalone mode):
```
!|SetOption|autocommit|true|

!|Execute Ddl|create table TESTTBL(ID int, VAL int)|

!|Commit|

!|Insert|TESTTBL|
|ID     |VAL    |
|1      |2      |

!|Rollback|

!|Query|select ID, VAL from TESTTBL|
|ID     |VAL    |
|1      |2      | -- reported as missing
```

MMatten:
Does the data even get committed at all? Can you run the DB as a server instead of embedded and does that give a different result?

javornikolov:
If there is an explicit `commit` after the insert - then it's committed and preserved after the rollback.

MMatten:
Does moving the commit after the DDL to be after the insert change anything? Probably not but I was wondering if an explicit commit is setting auto commit to false.

MMatten:
How might PR #397 be useful for this work?

javornikolov:
The idea was to have a bit more straightforward version of  `if (dbEnvironment == null || dbEnvironment.getConnection().isClosed())` in `SetOption` but maybe we can just add a `setAutoCommit` to DBEnvironment.

MMatten:
If I get some time tomorrow I'll play around with Derby and see if can spot anything strange in its auto commit behaviour.

javornikolov:
It's not for Derby only, similar behavior is with HSQLDB. I think I found out what's the problem. Autocommit mode is being reset after `commit` or `rollback`.  (I'm just removing the reset after `commit` from dbfit code. For `rollback` - I the jdbc driver is doing that, but I need to double check).

javornikolov:
> I'm just removing the reset after commit from dbfit code

I force-pushed that fix and seems it helps to resolve the mentioned issue.

Now auto-commit option may be set before connecting. But as mentioned earlier - for standalone mode there we may need to think triggering `Options.reset()` at some points.

dbfit-java/core/src/main/java/dbfit/fixture/SetOption.java
```diff
@@ -0,0 +1,72 @@
+package dbfit.fixture;
```
MMatten: A general query about future direction: do you think we should now be creating new fixtures for all DbFit commands rather than using `SequenceFixture` methods or looking to use `SequenceFixture`'s 'System Under Test' (i.e. just wrap a `DBEnvironment` and expose its methods like `Fixture`s)?

If we create fixture for all commands then I think we'd no longer really need `DatabaseEnvironment` at all (standalone mode) and the only use of `SequenceFixture` in Flow Mode would be to auto rollback at the end of the page.

javornikolov: Two points to consider are:
* Trying to keep uniform syntax between Standalone and Flow modes (when it makes sense) so that it's easier to reuse/migrate a test page between modes.
* Compatibility with existing tests.

javornikolov:
I also updated the transaction control docs a bit about autocommit mode.

@MMatten, @benilovj - any thoughts whether we need anything else to merge this PR?

MMatten:
If we replace `SequenceFixture` methods with `Fixture`s won't we break standalone mode tests that include things like: -

`|Database Environment|`
`|Commit|`
`|Close|`

(I have lots of test that use this construct)

or even

`|Database Environment|`
`|Set Option|fixed length string parsing|true|`

(I use this less).

I suppose it's less likely that people do something like: -

|dbfit.OracleTest|
|Set Option|fixed length string parsing|true|

(I think that should work although it looks strange).

MMatten:
In `DatabaseTest` the method's we're replacing with `Fixture`s are technically 'in flow' whereas `Fixture`s are alien fixtures ('traverses' in FitLibrary terms as far as I can tell).

Are there any potential impacts we've not considered yet? What about any DbFit fixture subtyping people might have done? I guess unlikely but some concerns were raised when opinion on deprecating the 2.0 .net code was canvassed.

website/_includes/manual/transaction-control.md
```diff
@@ -4,7 +4,22 @@ By default, each individual test (FitNesse page) in flow mode is executed in a t
 
 If in flow mode, you can use the Commit and Rollback commands to control the transactions manually, but remember that a final rollback will be added at the end of the test. These commands have no additional arguments.
 
-In standalone mode, you will probably control transactions from outside DbFit. Utility commands to commit and rollback are still provided, if you need them, as part of the DatabaseEnvironment fixture. For example, use this table to rollback:
+In standalone mode, you will probably control transactions from outside DbFit. Utility commands to commit and rollback are still provided, if you need them. For example, use this table to rollback:
 
-    !|DatabaseEnvironment|
```
MMatten: If this is no longer supported then I'll have lots of ETL tests to update! Can we support both?

javornikolov:
Right now things like the following are still working actually:
```
|Database Environment|
|Commit|
```

> In DatabaseTest the method's we're replacing with Fixtures are technically 'in flow' whereas Fixtures are alien fixtures ('traverses' in FitLibrary terms as far as I can tell).

`DatabaseTest` *is* what's used in flow mode. Standalone mode relies on `DatabaseEnvironment` fixture.


> Are there any potential impacts we've not considered yet? What about any DbFit fixture subtyping people might have done? I guess unlikely but some concerns were raised when opinion on deprecating the 2.0 .net code was canvassed.

Well, that's a good point. Maybe we need to increase version number to 3.x.x?

website/_includes/manual/transaction-control.md
javornikolov: I believe it's still supported. I removed it to simplify the reference a bit. But maybe we can leave a note that the "old" syntax is still supported?

MMatten:
> Right now things like the following are still working actually
Good to know!

> `DatabaseTest` *is* what's used in flow mode. Standalone mode relies on `DatabaseEnvironment` fixture.
Yes but methods that return Fit Fixtures are treated as 'alien' objects to DoFixture/SequenceFixture so perhaps some, as yet unknown, unintended impact on any subtypes people might have created although unlikely I guess. 

> Well, that's a good point. Maybe we need to increase version number to 3.x.x?
Or up to 4.0 even?

javornikolov:
> Or up to 4.0 even?

Ah, that's what I meant :-)

> Yes but methods that return Fit Fixtures are treated as 'alien' objects to DoFixture/SequenceFixture so perhaps some, as yet unknown, unintended impact on any subtypes people might have created although unlikely I guess.

We already have many such cases in `DatabaseTest` (the majority of methods there are returning Fixture). I just used them as an example when implementing this change. I haven't yet investigated deeply what does that mean and might be the consequences.

website/_includes/manual/transaction-control.md
MMatten: Sounds good. Perhaps 'alternate syntax'.

MMatten:
Ok. Nothing else from me to discuss on this one. Looks good.

MMatten:
One more query:

> Right now things like the following are still working actually:
```
|Database Environment|
|Commit|
```

But what about:

```
|Database Environment|
|Commit|
|Close|
```

Won't the new Commit fixture take over processing of the table and then ignore the 'Close'? That will break quite a few of my ETL tests. If so can we still support this (may be by pushing more into the Database Environment)?

javornikolov:
We should test that, if it's important to have it working - perhaps we should have it test suite.

MMatten:
More info on the breakages: not closing connections (in Teradata) results in sessions staying around in the database (until some kind of periodic garbage collection mops them up) but the accumulation results in max sessions for the DB being breached and tests suites start to fail consistently.

I'd have to update a large number of tests to move the 'Close' into a second 'Database Environment' table (often multiple times per test page).

javornikolov:
@MMatten, I'm just trying the following for Derby in the `auto-commit-support` branch and it seems to close the connection (the next statement is failing with `java.lang.IllegalArgumentException: No open connection to a database is available` which mean that the connection has been closed): 
```
!|DatabaseEnvironment|
|Commit|
|Close|
```

Do you mean you tested that for Teradata and it fails?

MMatten:
@javornikolov, that sounds good then.

I haven't run any tests yet with this branch.

My previous note was just to explain why I need this working (rather than leave it to the DBMS to close them down after the client had detached).

My thought was that the new commit Fixture would take over the processing of the `Database Environment` from the row containing the `Commit` (as the `SequenceFixture` method `commit` returns a Fit Fixture subclass - this is what the FitLibrary `DoFixture` / `SequenceFixture` docs suggest) and then the `Close` would get ignored.

javornikolov:
OK, good - to me it looks like the we're not breaking the existing behaviour. I'm going to merge this PR.

MMatten:
Ok. I think the noted behaviour only relates to Fixtures that are in flow. So it's not relevant for our standalone mode. We'd need to be wary in flow mode `DatabaseTest` though perhaps.

PR #395, opened 2015-07-11
bigint unsigned support for mysql

javornikolov:
Add support for  MySQL `BIGINT UNSIGNED` data type.

### DIFF
--- a/dbfit-java/core/src/main/java/dbfit/util/FitNesseTestHost.java
+++ b/dbfit-java/core/src/main/java/dbfit/util/FitNesseTestHost.java
@@ -1,6 +1,7 @@
 package dbfit.util;
 
 import java.math.BigDecimal;
+import java.math.BigInteger;
 
 import dbfit.api.TestHost;
 import fit.TypeAdapter;
@@ -9,6 +10,7 @@ public class FitNesseTestHost implements TestHost {
 
     static {
         TypeAdapter.registerParseDelegate(BigDecimal.class, BigDecimalParseDelegate.class);
+        TypeAdapter.registerParseDelegate(BigInteger.class, BigIntegerParseDelegate.class);
         TypeAdapter.registerParseDelegate(java.sql.Date.class, SqlDateParseDelegate.class);
         TypeAdapter.registerParseDelegate(java.sql.Timestamp.class, SqlTimestampParseDelegate.class);
         TypeAdapter.registerParseDelegate(java.sql.Time.class, SqlTimeParseDelegate.class);
--- a/dbfit-java/mysql/src/main/java/dbfit/environment/MySqlEnvironment.java
+++ b/dbfit-java/mysql/src/main/java/dbfit/environment/MySqlEnvironment.java
@@ -94,6 +94,8 @@ public class MySqlEnvironment extends AbstractDbEnvironment {
             "TINYINT", "SMALLINT", "MEDIUMINT", "INT", "INTEGER" });
     private static List<String> longTypes = Arrays.asList(new String[] {
             "BIGINT", "INTEGER UNSIGNED", "INT UNSIGNED" });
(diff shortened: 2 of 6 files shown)


dbfit-java/core/src/main/java/dbfit/util/FitNesseTestHost.java
```diff
@@ -9,6 +10,7 @@
 
     static {
         TypeAdapter.registerParseDelegate(BigDecimal.class, BigDecimalParseDelegate.class);
+        TypeAdapter.registerParseDelegate(BigInteger.class, BigIntegerParseDelegate.class);
```
MMatten: Is the type in MySQL 64 but unsigned and so will this class support the same range of values?

dbfit-java/mysql/src/main/java/dbfit/environment/MySqlEnvironment.java
```diff
@@ -94,6 +94,8 @@ protected String parseCommandText(String commandText) {
             "TINYINT", "SMALLINT", "MEDIUMINT", "INT", "INTEGER" });
     private static List<String> longTypes = Arrays.asList(new String[] {
             "BIGINT", "INTEGER UNSIGNED", "INT UNSIGNED" });
+    private static List<String> bigIntegerTypes = Arrays.asList(new String[] {
```
MMatten: Should the new group be called unsignedBigIntegerTypes? BIGINT is in the group LONG.

dbfit-java/core/src/main/java/dbfit/util/FitNesseTestHost.java
javornikolov: In [`MySQL jdbc driver docs`](http://dev.mysql.com/doc/connector-j/en/connector-j-reference-type-conversions.html) it's mentioned that `BIGINT UNSIGNED` is mapped to `java.math.BitInteger`.

javornikolov: The range of values of `java.math.BigInteger` I think is wider than `BIGINT UNSIGNED`.

dbfit-java/mysql/src/main/java/dbfit/environment/MySqlEnvironment.java
javornikolov: It's a bit confusing indeed. I'll rename it to `unsignedBigIntTypes`.

javornikolov: Renamed to `bigIntUnsignedTypes'

dbfit-java/core/src/main/java/dbfit/util/FitNesseTestHost.java
```diff
@@ -9,6 +10,7 @@
...
         TypeAdapter.registerParseDelegate(java.sql.Date.class, SqlDateParseDelegate.class);
         TypeAdapter.registerParseDelegate(java.sql.Timestamp.class, SqlTimestampParseDelegate.class);
         TypeAdapter.registerParseDelegate(java.sql.Time.class, SqlTimeParseDelegate.class);
```
MMatten: For my own understanding of how this works: when do we actually need to create parse delegates? Only when the class doesn't automatically provide a useful 'parse' method?

javornikolov: There are some predefined type adapters in [`fit.TypeAdapter`](https://github.com/unclebob/fitnesse/blob/master/src/fit/TypeAdapter.java); for anything else - a default one is used supporting String, Date and objects with `parse` method. And for anything else - usually a custom parse delegate is needed (otherwise an error is failed when trying to parse the values).

PR #398, opened 2015-07-12
Refactor checkConnectionValid

javornikolov:
Reduce direct calls to `checkConnectionValid` to simplify dependencies (as validity check is already performed by `getConnection`)


benilovj:
This change looks fine, but what's the context of it? That'd be useful to include in the commit message...

javornikolov:
> This change looks fine, but what's the context of it? That'd be useful to include in the commit message...

Commit message refined a bit. The refactoring came around #397. I just spotted that this `checkConnectionValid` is public static but not used in non-ancestor classes. For now I removed direct calls to it just to make things a bit simpler.

PR #402, opened 2015-07-17
Support long Teradata object names

### LINKED ISSUE #400: Long object name in Teradata
Hi,

Found problem using long object names in DBFit tests. Long = 30+ characters.
Test to reproduce problem (on tables, but e.g. Execute Procedure fails with same problem):
```|Import|
|dbfit.fixture|

Connect in Teradata mode:
|Database Environment|teradata|
|Connect|jdbc:teradata://192.168.31.139/USER=dbc,PASSWORD=dbc,TMODE=TERA|

|Execute Ddl|CREATE TABLE dftest.t12345678_123456789_123456789_123456789 (field1 number)|

|Insert|dftest.t12345678_123456789_123456789_123456789|
|field1|
|1|

|Execute Ddl|CREATE TABLE dftest.t12345678_123456789 (field1 number)|

|Insert|dftest.t12345678_123456789|
|field1|
|1|

|Execute Ddl|DROP TABLE dftest.t12345678_123456789_123456789_123456789|
|Execute Ddl|DROP TABLE dftest.t12345678_123456789|
```

First "Insert" fails with: java.sql.SQLException: Cannot retrieve list of columns for dftest.t12345678_123456789_123456789_123456789 - check spelling and access rights
Second "Insert" succeeds 

Problem: Before Teradata 14.0 (not sure) object name (tables, procedures, views etc) must be less 31 characters. But starting from 14.0 version 30+ character object names are allowed.
I assume DbFit uses dbc.tables dictionary to get list of columns etc. dbc.tables shrinks object name to 30. That is how it works. Meanwhile dbc.tablesv display all names (including long) correctly. So I assume adding "v" to metadata request should solve the issue. 

P.S. Problem proved for: Teradata 14.0, JDBC Driver 14.10, JDBC Driver 14.11, JDBC Driver 15.0

MMatten:
* Add support for long table and column names in Teradata.
* Add new `CoreTests` test to test long names for all supported DB types.

resolves #400


MMatten:
Long tables/column names are supported in Teradata from version 14.10.

Teradata Express is only available up to version 14 at the moment so we'll need assistance to test names between 31 and 128 characters.

javornikolov:
Thanks @MMatten .
Looking at the travis output: here is the [the dbfit build](https://s3.amazonaws.com/dbfit/artifacts/71451580/71451581/dbfit-complete-3.1.0.zip).

MMatten:
Thanks. I'll need to ask the raise of #400 to download this and test with Teradata 14.10+.

MMatten:
The new `CoreTests` test it an attempt to add consistent structure to the tests even when the content has to vary by DBMS. I guess we could end up with lots of `!define` variables with this approach though.

javornikolov:
> The new CoreTests test it an attempt to add consistent structure to the tests even when the content has to vary by DBMS. I guess we could end up with lots of !define variables with this approach though

What about DB2, Derby, HSQLDB, MS SQL Server?

MMatten:
> What about DB2, Derby, HSQLDB, MS SQL Server?

Aha! I've forgotten to commit the symlink updates in the `properties.xml` files...

MMatten:
Missing symlinks added for adaptors not using all of `CoreTests`.

Note that adapters that have a 128 char name length limit use the names from the `!define`s set in `FitNesseRoot/DbFit/AcceptanceTests/content.txt`.

build.gradle
```diff
@@ -182,30 +182,30 @@ task starthere(type: Exec, dependsOn: [copyLocal, copyExtraLibs, copyTestLibs, c
...
-    ':dbfit-java:hsqldb:check',
-    ':dbfit-java:mysql:check',
-    ':dbfit-java:netezza:check',
-    ':dbfit-java:oracle:check',
-    ':dbfit-java:postgres:check',
-    ':dbfit-java:sqlserver:check',
-    ':dbfit-java:teradata:check',
-    ':dbfit-java:derby:integrationTest',
-    ':dbfit-java:hsqldb:integrationTest'
+      ':dbfit-java:core:check'
```
javornikolov: I would vote for keeping the commas at the end of line.

MMatten: Ok for the `fastbuild`.

The pain is on the `integrationBuild` - when I uncomment `':dbfit-java:teradata:integrationTest'` I have to remember to remove the comma from the end of `':dbfit-java:sqlserver:integrationTest'` - always forget!

Ok to have different formatting/style?

javornikolov: Leading comma is a bit unnatural for people to read. Besides - I think we rarely commit changes to this list. If its some personal customization - it's perhaps better to keep customizations in separate (optional) non-version controlled file which to load if present (since it will be personal, anyone can use own style and formatting).

MMatten: Ok, I don't mind that.

Any ideas on how to extend the build customisation using another (non version controlled) file (or something like environment variables)?

MMatten:
I've reverted the comma position in `build.gradle` and added an idea for your feedback please - using the user's `gradle.properties` to exclude `integrationBuild` tasks.

I've not bothered making this iterate over a list of DB type names/integration test tasks but just to show the basic idea.

What do you think?

javornikolov:
Let's extract to a separate PR the configuring excluded/included integration tests. (Seems a bit off-topic in the context of Teradata long names support). The rest looks OK to me.

MMatten:
Good idea. It's getting too in depth for this PR.

Now removed from this one.

PR #404, opened 2015-07-19
Allow exclusion of DB types in Gradle `integrationBuild` task via gradle.properties

MMatten:
Exclude certain DB types during the `integrationBuild` task using a Gradle property.


CONTRIBUTING.md
```diff
@@ -36,7 +36,8 @@ If you have to make changes to `core`, please run all integration tests (because
 
     $ ./gradlew integrationBuild
 
-Note that this runs the integration tests for the adaptors listed under the `integrationBuild` task in the main Gradle build script (`build.gradle` in dbfit project root directory). You can edit this task to run the integration tests for the set of adaptors for which you have test databases available.
+Note that this runs the integration tests for the adaptors listed under the `integrationBuild` task in the main Gradle build script (`build.gradle` in dbfit project root directory). 
+To exclude certain integration tests, during execution of this task, configure the Gradle property `excludeIntegrationTests` (in your `gradle.properties` in the DbFit project root directory) by giving it a value of a comma separated list of DB types to exclude.
```
javornikolov: It might be helpful to have an example.

javornikolov:
I just spotted that Gradle has a feature called [init scripts](https://docs.gradle.org/current/userguide/init_scripts.html) which seems like another alternative for customizing builds. I'm not sure yet if it's suitable for the particular case here, it's just an idea to research.

build.gradle
```diff
@@ -196,17 +196,21 @@ task fastbuild(dependsOn: [
...
-    ':dbfit-java:oracle:integrationTest',
-    ':dbfit-java:postgres:integrationTest'
-    //':dbfit-java:sqlserver:integrationTest',
-    //':dbfit-java:teradata:integrationTest',
-    ])
+def integrationBuildTasks = []
+def excludedIntegrationBuildDbTypes = []
+
+if (project.hasProperty('excludeIntegrationTests')) {
+    excludedIntegrationBuildDbTypes=project.excludeIntegrationTests.split(",")
```
javornikolov: Missing space around `=`.
split(",") seems to retain white space, new line, etc. I wonder if there is some built-in parsing which to reuse.

build.gradle
```diff
@@ -196,17 +196,21 @@ task fastbuild(dependsOn: [
...
-    ':dbfit-java:derby:integrationTest',
-    ':dbfit-java:hsqldb:integrationTest',
-    ':dbfit-java:mysql:integrationTest',
-    //':dbfit-java:netezza:integrationTest',
-    ':dbfit-java:oracle:integrationTest',
-    ':dbfit-java:postgres:integrationTest'
-    //':dbfit-java:sqlserver:integrationTest',
-    //':dbfit-java:teradata:integrationTest',
-    ])
+def integrationBuildTasks = []
```
javornikolov: What if we just have a pure function returning the list of tasks? It's a bit harder to follow the side effects when we have kind of global variables which are later amended.

MMatten:
> I just spotted that Gradle has a feature called [init scripts](https://docs.gradle.org/current/userguide/init_scripts.html) which seems like another alternative for customizing builds. I'm not sure yet if it's suitable for the particular case here, it's just an idea to research.

I guess we could define the list of integration test tasks to be executed, as a list, using one if these, and check in the main build if that's been set but I don't think that makes it any easier. 

Any other ideas on using these?

javornikolov:
> I just spotted that Gradle has a feature called init scripts which seems like another alternative for customizing builds. I'm not sure yet if it's suitable for the particular case here, it's just an idea to research.
>> I guess we could define the list of integration test tasks to be executed, as a list, using one if these, and check in the main build if that's been set but I don't think that makes it any easier.
>>
>> Any other ideas on using these?

My idea was similar. Yes, no obvious advantage over the `gradle.properties` approach (just a minor one for parsing the comma-separated string to list).

build.gradle
```diff
@@ -196,17 +196,21 @@ task fastbuild(dependsOn: [
...
+def excludedIntegrationBuildDbTypes = []
+
+if (project.hasProperty('excludeIntegrationTests')) {
+    excludedIntegrationBuildDbTypes=project.excludeIntegrationTests.split(",")
+}
+
+subprojects.each { sp ->
+    if (!(sp.name in ["dbfit-java", "core"])) {
+        sp.name in excludedIntegrationBuildDbTypes ?
```
javornikolov: This may be defined as set operation, i.e. something like
``` groovy
(subprojects.collect { it.name } - excludedIntegrationBuildTypes).collect {
    ":dbfit-java:${it}:integrationTest"
}
```

javornikolov:
To summarize the suggestions so far + some additions:
* The approach based on `gradle.properties` seems OK
* Consider setting a default (empty) value for `excludeIntegrationTests` if the property is still unset. It's easier to work with it later
* It's good to trim values after `split(",")`
* Consider adding an `assert` to make sure `excludeIntegrationTests` contains valid project names only
* Prefer declarative syntax and set operations; avoid side effects (assignment to global variables)

MMatten:
Ok. Some updates made.

Also, perhaps we could add a default `gradle.properties` with SQL Server, Teradata and Netezza pre-disabled for convenience when using the test VM.

build.gradle
```diff
@@ -196,17 +196,33 @@ task fastbuild(dependsOn: [
...
+        assert dbType in validSubProjectNames
+    }
+
+    excludedIntegrationBuildDbTypes.add("dbfit-java")
+    excludedIntegrationBuildDbTypes.add("core")
+
+    return (subprojects.collect { it.name } - excludedIntegrationBuildDbTypes).collect {
+               ":dbfit-java:${it}:integrationTest"
+           }
+}
```
javornikolov: Here is an attempt to make it a bit more concise:

``` groovy
if (!project.hasProperty('excludeIntegrationTests')) {
    project.ext.set('excludeIntegrationTests', '')
}

def excludedIntegrationBuildDbTypes() {
    (["dbfit-java", "core"] + project.excludeIntegrationTests.split(",").collect { it.trim() }) as Set
}

assert (excludedIntegrationBuildDbTypes() - subprojects.collect { it.name }).empty

def integrationBuildTasks() {
    (subprojects.collect { it.name } - excludedIntegrationBuildDbTypes()).collect {
        ":dbfit-java:${it}:integrationTest"
    }
}

task integrationBuild(dependsOn: [integrationBuildTasks()])
```

javornikolov:
> Also, perhaps we could add a default gradle.properties with SQL Server, Teradata and Netezza pre-disabled for convenience when using the test VM.

Yes. Having that - it would also be a good example as a starting point for such customizations.

MMatten:
Thanks for that. 

It's nice and concise.The set operations are very cool.

`split()` has an issue though when `excludeIntegrationTests=` (empty) in `gradle.properties` so I've used `tokenize()` instead which works ok (and left a comment - keep or remove this?).

I've defined the function as returning `Set` instead of `Object` (via `def`). What are your views on this more specific declaration?

gradle.properties
```diff
@@ -0,0 +1 @@
+excludeIntegrationTests=netezza,sqlserver,teradata
```
javornikolov: I think we shouldn't have this `gradle.properties` here but rather in the test vm provisioning.

javornikolov:
> split() has an issue though when excludeIntegrationTests= (empty) in gradle.properties so I've used tokenize() instead which works ok (and left a comment - keep or remove this?).

We can remove the comments.

> I've defined the function as returning Set instead of Object (via def). What are your views on this more specific declaration?

It's OK.

gradle.properties
MMatten: Is a new Chef recipe what you had in mind? I've just pushed something to create the file but I can't get to test this until later.

test_vm/vendor/cookbooks/dbfit_test/recipes/gradle.rb
```diff
@@ -0,0 +1,8 @@
+project_root = node['dbfit']['project_root']
+  
+file '#${project_root}/gradle.properties' do
+  content 'excludeIntegrationTests=netezza,sqlserver,teradata'
+  mode '0644'
+  owner 'vagrant'
+  group 'vagrant'
```
MMatten: I'm not sure what the user group is for this user ATM.

gradle.properties
MMatten: I guess I we could use the basic `:shell` provisioner too. A simple `echo` would probably do it!

test_vm/vendor/cookbooks/dbfit_test/recipes/gradle.rb
javornikolov: The user is vagrant, the group I think too.

gradle.properties
javornikolov: Yes, that's what I had in mind.

javornikolov: > I guess I we could use the basic :shell provisioner too. A simple echo would probably do it!

For this case Chef is OK. (Shell is a bit lower level).
We may `template` - like in `environment.rb` recipe.

MMatten: I've left it using `file` for now. I need to test the provisioning still.

MMatten:
All appears to be working well now.

javornikolov:
Great. Thank you, @MMatten! I'm merging...

PR #405, opened 2015-07-20
RAW Type Added

Auguronomics:
Allow to work with RAW datatype 

!|Execute|create table raw_line(objectid integer, myGuid RAW)|

!|Insert |raw_line |
|objectid|myGuid |
|10      |0ce894b4-bf19-e511-840f-0050568969e2|

!|Query  |select objectid, RAW from spatial_line|
|objectid|myGuid? |
|10      |0ce894b4-bf19-e511-840f-0050568969e2|

### DIFF
--- a/dbfit-java/oracle/src/main/java/dbfit/environment/OracleEnvironment.java
+++ b/dbfit-java/oracle/src/main/java/dbfit/environment/OracleEnvironment.java
@@ -390,7 +390,7 @@ public class OracleEnvironment extends AbstractDbEnvironment {
     // map types
     private static List<String> stringTypes = Arrays.asList(new String[] {
             "VARCHAR", "VARCHAR2", "NVARCHAR2", "CHAR", "NCHAR", "CLOB",
-            "NCLOB", "ROWID", "BOOLEAN","RAW"
+            "NCLOB", "ROWID", "BOOLEAN", "RAW"
         });
     private static List<String> decimalTypes = Arrays.asList(new String[] {
             "BINARY_INTEGER", "NUMBER", "FLOAT" });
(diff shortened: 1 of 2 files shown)


MMatten:
As `RAW` is a binary type, like `BLOB` would `byte[]` be a more appropriate, than `String` Java type for this? There is also `oracle.sql.RAW' available in the JDBC driver.

MMatten:
We could do with an acceptance test for this added into:

`FitNesseRoot/DbFit/AcceptanceTests/JavaTests/OracleTests/FlowMode/DataTypes`

javornikolov:
I'm not sure why are these travis.yml updates - if nothing else, seems something out of scope of the RAW data type support.

dbfit-java/oracle/src/main/java/dbfit/environment/OracleEnvironment.java
```diff
@@ -390,7 +390,7 @@ private CallableStatement openDbCallWithParameters(String query,
     // map types
     private static List<String> stringTypes = Arrays.asList(new String[] {
             "VARCHAR", "VARCHAR2", "NVARCHAR2", "CHAR", "NCHAR", "CLOB",
-            "NCLOB", "ROWID", "BOOLEAN"
+            "NCLOB", "ROWID", "BOOLEAN","RAW"
```
javornikolov: Missing space after the `,`.

Auguronomics: I correct the space lack, now

javornikolov:
> As RAW is a binary type, like BLOB would byte[] be a more appropriate, than String Java type for this? There is also oracle.sql.RAW available in the JDBC driver.

Also, since RAW may contain non-printable characters, maybe it makes sense to use e.g. HEX representation on the test pages.

Auguronomics:
RAW looks like CLOB,
I'll add the dbfit Documentation soon, copy paste from the post comment.
I'm in trouble to download the travis build form S3.

javornikolov:
> RAW looks like CLOB,

It's more like BLOB. (We don't have BLOB support yet either).

PR #412, opened 2015-08-01
Update method for artifacts upload to s3

javornikolov:
Switching to the new method for uploading artifacts to S3 as per: http://docs.travis-ci.com/user/uploading-artifacts/

The old project (https://github.com/travis-ci/travis-artifacts) is no longer maintained.

**Not ready for merging yet!**

 - [x] Set aws key details
 - [x] Remove debug option

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,20 @@
-env:
- global:
-   - "ARTIFACTS_AWS_REGION=us-east-1"
-   - "ARTIFACTS_S3_BUCKET=dbfit"
-   - secure: "NLAj1hPcTlzBO6hEbZ3X6AOfXUVeFl0oSytrd02m+UaXzbNA9PZ6ZjYBwCGA\nvpNdVd5YVNMaUWBX5IN336mscvJswg+9lyt3y8cgBzOUWNE/UXDRXYoL8xwx\nxXnMnBatW2+0M4QtuXyLZpWyPxKa+1AYuCVJXHaANrkjufuXm04="
-   - secure: "li8rtIJ0NxvXPn2o1TltT+CaiMyfN0neFoBBA7CXCRWOt0Gjs6ytzP89Hclu\nbdPuZsCXXKckW3UIIzqR6YDU0NhLZxkZuOz0u3+dD1856jqa2umXJi3MnQ6j\nlulUkNwfmE+ul0Hly0Hia4nNivHQ3f4I7BNSQlNM0GVYy+ZL4Oo="
 sudo: false
 language: java
 before_install:
-  - "./gradlew -version"
+- "./gradlew -version"
 install: "./gradlew assemble"
 script: "./gradlew clean fastbuild"
 jdk:
-  - oraclejdk7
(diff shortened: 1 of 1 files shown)


javornikolov:
Hi @benilovj, this new approach of uploading artifacts to S3 seems which I tried to setup seems to require different type of AWS keys (as described http://docs.travis-ci.com/user/uploading-artifacts/). Do you have an idea how can we set them up here?

Looks like the old way of travis artifacts upload is already causing some troubles so I wonder if we can switch it.

benilovj:
@javornikolov I've added a set of credentials to Travis, which should now allow pushing to the `dbfit` S3 bucket using the new method.

I'm not sure that the `.travis.yml` is completely correct though; do we need the `key`, `secret` and all those other bits?

benilovj:
Oh, I see, apologies, you're using the public signing method described [here](http://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings).

benilovj:
Well, something seems to be working:

image

although it's complaining about no uploading for PRs, which I don't quite understand:

image

see https://travis-ci.org/dbfit/dbfit/jobs/89615927

benilovj:
Oh, I think I understand what's going on. For a branch that's also a PR, 2 builds are triggered (one for the PR and another for the branch changes), but only the branch one will upload the artefact...

.travis.yml
```diff
@@ -1,21 +1,20 @@
...
-# the next line is a workaround for https://github.com/travis-ci/travis-artifacts/issues/23
-  - "gem install --version 0.8.9 faraday"
-  - "gem install travis-artifacts"
-  - "travis-artifacts upload --path zips --target-path artifacts/$TRAVIS_BUILD_ID/$TRAVIS_JOB_ID"
+- "./gradlew bundle"
+addons:
+  artifacts:
+    paths:
+    - zips/
+    target_paths: artifacts/$TRAVIS_REPO_SLUG/$TRAVIS_BUILD_NUMBER/$TRAVIS_JOB_NUMBER:artifacts/$TRAVIS_REPO_SLUG/$TRAVIS_COMMIT
```
benilovj: with this config, the build artefact is uploaded 3 times:

- once with the build number, from the JDK7 build
- once with the build number, from the JDK8 build
- once against the commit

This means that each build produces about 50Mb of files. Is this necessary?

javornikolov: It might make some sense to have a build both for jdk7 and jdk8 though the 3-rd one is perhaps overkill.
Also I guess it might be good to purge some of the old artifacts.

benilovj: OK, i've gotten rid of the second target path.

benilovj:
OK, that should be set up now, and I've also added a bucket policy which should make those artefacts public-readable.

javornikolov:
Awesome :-) Thanks!

benilovj:
I've cleaned up the old artefacts... hope that's not a problem but the new upload method is still working fine.

javornikolov:
OK

Issue #425, opened 2015-09-17
Access to guest test VM FitNesse server is blocked by firewall by default


MMatten:
Seems to be `firewalld`.

So the vagrant port forwarding of 8085 is currently unusable.

When running the commands listed at http://www.liquidweb.com/kb/how-to-stop-and-disable-firewalld-on-centos-7/ the home page becomes accessible.

MMatten:
There is a Chef supermarket cookbook to manage the firewall: -

https://supermarket.chef.io/cookbooks/firewalld

MMatten:
Resolved by #426.

PR #426, opened 2015-09-18
Configure guest VM firewall for host access to FitNesse server

MMatten:
Allow access to VM TCP port 8085 for host browser access to FitNesse.


MMatten:
Resolves #425.

javornikolov:
@MMatten, I guess some actions should be invoked from the imported cookbook?

MMatten:
>@MMatten, I guess some actions should be invoked from the imported cookbook?

Ah, indeed! I've omitted to commit the `firewalld.rb` recipe to configure the firewall! Will add now.

MMatten:
Now with added recipe.

javornikolov:
Looks good now :+1:

PR #432, opened 2015-09-27
Informix adaptor initial review

MMatten:
For review of the initial Informix adaptor implementation from https://github.com/dbfit/dbfit/pull/326 with: -
* Informix test VM installer added.
* CoreTests used for flow and stand-alone modes.

Suggestions for more to do: -
* How do we handle original commits by PR #326 author (retain or squash)?
* Add more tests for Informix data types.
* Apply good style to Java and other code.

What else?


javornikolov:
> How do we handle original commits by PR #326 author (retain or squash)?

Looking at the commits themselves - maybe it would be better to tidy up and squash them, leaving attribution to the original authors in the commit messages. Might be several commits if they represent significant enough changes.

dbfit-java/informix/src/integration-test/resources/deinstall_informix.sh
```diff
@@ -0,0 +1,64 @@
...
+INFORMIXDIR=/informix
+INFORMIX_SCRIPTS=$DBFIT_ROOT/dbfit-java/informix/src/integration-test/resources
+
+if [ ! -x "$INFORMIXDIR/uninstall/uninstall_ids" ]
+then
+	echo "$EM cannot find Informix deinstaller $INFORMIXDIR/uninstall/uninstall_ids" >&2
+	exit 1
+fi
+
+. $INFORMIXDIR/dbfitifserver.ksh
```
javornikolov: Is that safe enough? Sounds like `Korn Shell` script being sourced inside bash (we have `#!/usr/bin/env bash` on 1st line). Maybe it's OK, it depends on the script content.

test_vm/vendor/cookbooks/dbfit_test/recipes/firewalld.rb
```diff
@@ -0,0 +1 @@
+firewalld_port '8085/tcp'
```
javornikolov: We can rebase on top of master.

test_vm/vagrant_config_defaults.rb
```diff
@@ -2,6 +2,6 @@
   {nfs: true, nfs_version: 4, mount_options: ["noatime", "nodiratime"]}
 
 # Default options
-CONFIG['memory'] = "1100"
+CONFIG['memory'] = "2048"
```
javornikolov: Do we need to change the default? Memory size can be customized in `vagrant_config_sample` (that's needed e.g. for Oracle too).

build.gradle
```diff
@@ -194,7 +195,7 @@ task fastbuild(dependsOn: [
     ':dbfit-java:sqlserver:check',
     ':dbfit-java:teradata:check',
     ':dbfit-java:derby:integrationTest',
-    ':dbfit-java:hsqldb:integrationTest'
+    ':dbfit-java:hsqldb:integrationTest',
```
javornikolov: Do we need that comma?

FitNesseRoot/DbFit/AcceptanceTests/JavaTests/InformixTests/StandaloneFixtures/SetUp/content.txt
```diff
@@ -0,0 +1,5 @@
+|Import Fixture|
+|dbfit.fixture|
+
+|Database Environment|Informix|
+|Connect|localhost:16466|dbfit|DBFIT|dbfit|
```
javornikolov: We might alternatively connect via `Connect Using File`. This is the case with Oracle - in theory it allows easier switch to a custom test database outside the VM.

dbfit-java/informix/src/integration-test/resources/deinstall_informix.sh
MMatten: This shell script is created by the Informix installer. It just sets and exports some environment variables so that the tools can be used and looks fine to use with `bash`: -

```
INFORMIXDIR=/informix
INFORMIXSERVER=dbfitifserver
ONCONFIG=onconfig.dbfitifserver
INFORMIXSQLHOSTS=/informix/etc/sqlhosts.dbfitifserver
GL_USEGLU=1
PATH=${INFORMIXDIR}/bin:${INFORMIXDIR}/extend/krakatoa/jre/bin:${PATH}
export INFORMIXDIR INFORMIXSERVER ONCONFIG INFORMIXSQLHOSTS GL_USEGLU PATH

File: dbfit-java/build.gradle [hunk 6]
```diff
@@ -159,6 +159,19 @@ project('hsqldb') {
...
+project('informix') {
+    description = 'DBFit Informix engine'
+
+    dependencies {
+        compile project(':dbfit-java:core')
+        compile 'com.informix:ifxjdbc:4.10.JC4W1@jar'
+    }
+
+    integrationTest.doFirst {
+        dbdeploy('com.informix.jdbc.IfxDriver', 'jdbc:informix-sqli://localhost:16466/dbfit:INFORMIXSERVER=dbfitifserver', 'dbfit', 'DBFIT', 'row')
```
javornikolov: Remark similar to tests connection string: can we easily switch to custom test database instance (on another host/port)?

dbfit-java/informix/src/integration-test/resources/deinstall_informix.sh
javornikolov: Nice, it's simple enough - it's fine to use within bash, yes.

dbfit-java/informix/src/main/java/dbfit/environment/InformixEnvironment.java
```diff
@@ -0,0 +1,416 @@
...
+import static dbfit.util.Direction.*;
+
+@DatabaseEnvironment(name="Informix", driver="com.informix.jdbc.IfxDriver")
+public class InformixEnvironment extends AbstractDbEnvironment  {
+
+   @Override
+     public void afterConnectionEstablished() throws SQLException {
+          if(currentConnection.getTransactionIsolation()==0){
+        	  System.out.println("afterConnectionEstablished method");
+        	currentConnection.setAutoCommit(true);
```
javornikolov: As we already added some auto-commit related features - this might need some refinement.

test_vm/vagrant_config_defaults.rb
MMatten: With the default we don't have enough to run the Informix server. Do you think we should make users change this if they want to run Informix?

javornikolov: If Oracle+DB2+Informix are started altogether even 2GB won't be enough. My considerations in terms of keeping the default small were that it seemed less risky to run out of memory inside the VM than in the host OS.

dbfit-java/informix/src/main/java/dbfit/environment/InformixEnvironment.java
MMatten: Yes. I've created the current test DB in `ANSI` mode (it supports transactions) and we need to test in Informix mode too. I was thinking of creating a second test DB for that. One more TODO then.

dbfit-java/build.gradle
MMatten: Yes we can. I've just followed the precedent (MySQL and Postgres). Do you think we externalise these details into another file? On a separate issue/PR?

test_vm/vagrant_config_defaults.rb
MMatten: I think I've run Oracle+DB2+Informix fine with the 2GB on the guest. I'll see how much I can reduce this before the installers or the tests start failing.

dbfit-java/build.gradle
javornikolov: If we doesn't complicate things we can implement it here. (E.g. just do it like it's in Oracle).

FitNesseRoot/DbFit/AcceptanceTests/JavaTests/InformixTests/StandaloneFixtures/SetUp/content.txt
MMatten: Now changed to use `Connect Using File`.

test_vm/vagrant_config_defaults.rb
MMatten: Reduced to `1500` and tests including Oracle+DB2+Informix are still passing. I'll reduce further and see where things break.

MMatten: At `1300` now and still running ok.

javornikolov: 1300 is perhaps good enough.

MMatten: Perhaps I'm going mad. I've just run it through fine with `1100`. I'll revert to `1100`.

MMatten:
Ok, so left to do is something like: -
* [ ] add Informix (in addition to ANSI) mode database and test auto-commit.
* [ ] tests for Informix data types.
* [ ] tidy up commits.

Anything else obviously missing/wrong at this point?

javornikolov:
Sounds good as major points. Some refactoring and reformatting would be appropriate but I guess that goes into the 3rd point anyway.

MMatten:
Hi @javornikolov, I've added a fair number of acceptance tests now but I'm having problems getting tests for the `DECIMAL` and `MONEY` types to pass.

The `setObject` method of `StatementExecution` fails on the Informix `CommonTests.DataTypes.NumberTests` test when a `NormalisedBigDecimal` is passed to it. This is specific to the Informix JDBC driver.

I've tested that it can handle a standard `BigDecimal` correctly.

I can't see anywhere where we can cast the `NormalisedBigDecimal` back to a `BigDecimal` before passing it to `setObject`. 

I can only think that the driver does something like checking for specific class types rather something like `instanceof` for each supported type. Error message and stack trace: -

    java.sql.SQLException: System or internal error     java.lang.NoSuchMethodException:com.informix.jdbc.IfxDecimal.fromnull(dbfit.util.NormalisedBigDecimal)
	at com.informix.util.IfxErrMsg.getSQLException(IfxErrMsg.java:472)
	at com.informix.jdbc.IfxObject.findMethod(IfxObject.java:163)
	at com.informix.jdbc.IfxObject.fromObject(IfxObject.java:186)
	at com.informix.jdbc.IfxValue.makeInstance(IfxValue.java:556)
	at com.informix.jdbc.IfxValue.makeInstance(IfxValue.java:518)
	at com.informix.jdbc.IfxPreparedStatement.setObject(IfxPreparedStatement.java:2113)
	at dbfit.fixture.StatementExecution.setObject(StatementExecution.java:44)
	at dbfit.util.DbParameterAccessor.set(DbParameterAccessor.java:84)
	at dbfit.util.DbParameterAccessorTypeAdapter.set(DbParameterAccessorTypeAdapter.java:22)
	at fit.Binding$SetBinding.doCell(Binding.java:212)
	at dbfit.util.SymbolAccessSetBinding.doCell(SymbolAccessSetBinding.java:20)
	at dbfit.fixture.DbObjectExecutionFixture.runRow(DbObjectExecutionFixture.java:151)
	at dbfit.fixture.DbObjectExecutionFixture.doRows(DbObjectExecutionFixture.java:79)
	at fit.Fixture.doTable(Fixture.java:156)
	at fitlibrary.traverse.AlienTraverseHandler.doTable(AlienTraverseHandler.java:21)
	at fitlibrary.traverse.workflow.DoTraverseInterpreter.interpretWholeTable(DoTraverseInterpreter.java:104)
	at fitlibrary.traverse.workflow.DoTraverseInterpreter.interpretWholeTable(DoTraverseInterpreter.java:89)
	at fitlibrary.DoFixture.interpretWholeTable(DoFixture.java:73)
	at fitlibrary.suite.InFlowPageRunner.run(InFlowPageRunner.java:27)
	at fitlibrary.DoFixture.interpretTables(DoFixture.java:42)
	at dbfit.DatabaseTest.interpretTables(DatabaseTest.java:26)
	at fit.Fixture.doTables(Fixture.java:81)
	at fit.FitServer.process(FitServer.java:81)
	at fit.FitServer.run(FitServer.java:56)
    at fit.FitServer.main(FitServer.java:41)

Any ideas on the casting? 

I did start looking at the type normalising code to see if there was anything useful in there but I couldn't spot anything. In fact I'm a bit confused as to what it actually does and why. Can you explain it?

MMatten:
@javornikolov, regarding tidying up the commits - my current thinking is to squash to a single commit. There's a lot of them and I think they're all directly related to the Informix adapter.

Any concerns if I were to create a single commit (for the work done so far - I'm sure there's more to do before merging of course).

javornikolov:
For `BigDecimal` - there is custom parsing and normalization - they're configured in `FitNesseTestHost`.

Well `NormalisedBigDecimal` extends `BigDecimal` - in general there is no need of explicit cast.

Which exactly test is failing for example?  What exactly numeric value is causing the above-mentioned exception?
`java.lang.NoSuchMethodException:com.informix.jdbc.IfxDecimal.fromnull(dbfit.util.NormalisedBigDecimal)`

I wonder is it null or not null?

> Any concerns if I were to create a single commit (for the work done so far - I'm sure there's more to do before merging of course).

Sounds OK to me.

dbfit-java/informix/src/main/java/dbfit/environment/InformixEnvironment.java
```diff
@@ -0,0 +1,344 @@
...
+import java.util.regex.Pattern;
+
+import static dbfit.util.Direction.*;
+
+@DatabaseEnvironment(name="Informix", driver="com.informix.jdbc.IfxDriver")
+public class InformixEnvironment extends AbstractDbEnvironment  {
+
+    @Override
+    public void afterConnectionEstablished() throws SQLException {
+        TypeAdapter.registerParseDelegate(java.math.BigDecimal.class, new dbfit.util.BigDecimalParseDelegate());
```
javornikolov: Looks like this `registerParseDelegate` call is already present in `FitNesseTestHost`. Do we need it here again?

MMatten:
> Which exactly test is failing for example? What exactly numeric value is causing the above-mentioned exception?
`java.lang.NoSuchMethodException:com.informix.jdbc.IfxDecimal.fromnull(dbfit.util.NormalisedBigDecimal)`

I've tried it with simple decimal values. E.g.: -

    |Execute Ddl|CREATE TABLE datatypetest (d1 DECIMAL(5,2))|
    
    |Insert|datatypetest|
    |d1  |
    |-1.2|

I've tried creating a simple JDBC application to test a `PreparedStatement` and calling `setObject` and passing `BigDecimal` and a `NormalisedBigDecimal`. It too fails, with this error, when passed a `NormalisedBigDecimal` but works with a `BigDecimal`.

dbfit-java/informix/src/main/java/dbfit/environment/InformixEnvironment.java
MMatten: Nope. I'll get rid of that.

javornikolov:
> I've tried creating a simple JDBC application to test a PreparedStatement and calling setObject and passing BigDecimal and a NormalisedBigDecimal. It too fails, with this error, when passed a NormalisedBigDecimal but works with a BigDecimal.

I guess we can test whether it's due to the customized behavior of overridden methods (compared to plan BigDecimal) or due to having different class name.

MMatten:
After creating a basic sub type with: -

	public class MyBigDecimal extends java.math.BigDecimal {
        public MyBigDecimal(String val) {
    		super(val);
    	}
	}

I tried again. 

    pst.setObject(1, new MyBigDecimal("123.45"), java.sql.Types.DECIMAL);

fails, but

    pst.setObject(1, new BigDecimal("123.45"), java.sql.Types.DECIMAL);

works fine.

javornikolov:
We can try to modify BigDecimalParseDelegate to directly return `BigDecimal` (instead of NormalisedBigDecimal). Maybe that would be enough.

MMatten:
> We can try to modify `BigDecimalParseDelegate` to directly return `BigDecimal` (instead of `NormalisedBigDecimal`). Maybe that would be enough.

I think we need the overridden `equals` method though as the one from `BigDecimal` considers objects having the same decimal value, but different scale/precision, to be unequal.

javornikolov:
> I think we need the overridden equals method though as the one from BigDecimal considers objects having the same decimal value, but different scale/precision, to be unequal.

It's needed when we compare objects. But no need to do so when passing them as query parameters. I can quickly draft out a few ideas:
* Don't rely on `equals` method directly. Use some kind of comparator which should be possible to override/customize. (That might need major changes since Fit/FitNesse I think is using `equals` internally with no ability to change that. Maybe such kind of change would be in a direction of decoupling from Fit - which we're anyway aiming to).
* Add additional step of optional objects' conversion before passing them as JDBC parameters.
* As workaround: if by chance it appears that we only need `equals` on objects coming from JDBC, and the ones parsed from test tables are on the right side only (i.e. passed as parameters to equals): it might work to have `normalisation` produce `NormalisedBigDecimal`, and `*ParserDelegate` - plain `BigDecimal`.

MMatten:
Thanks for that. 

Yes, I think Fit uses `parse` and `equals` within it and so we must use those which I guess is one of two reasons why we have `NormalisedBigDecimal` (the other is the precision/scale independent comparisons).

I've started looking at how we can convert objects, using the `TypeNormaliser` code. I could specify a type normaliser for `NormalisedBigDecimal` to convert it to `BigDecimal` and include that in the `set` method of `DbParameterAccessor` (the reverse of that done by the `get` / `normaliseValue` methods. What do you think of that approach?

I'll look more into your third option above too.

javornikolov:
I'm not sure about reusing `TypeNormaliser`, it may be better to introduce some new interface. But besides that - yes, we can do something similar.

Another alternative may be to chase IBM to convince them about enhancing their driver. But I guess that won't be easy.

MMatten:
May be something like `TypeSpecifier` (make a more specific type) might be a better name. Tests so far look good but I'd need to run all adapters' tests.

I can try to log issue with IBM but as I'm not a paying customer I'm not sure if it's even possible yet.

MMatten:
The next weird thing with the Informix driver is that function return values need to be obtained by calling the function with `executeQuery` of `CallableStatement`! 

Functions can have `OUT` params, which can be accessed like stored procedure `OUT` params, as well as a return value. That's a trickier change.

MMatten:
I've been looking at what's needed to handle the return value via executeQuery and the ResultSet. 

I'm not too sure where to attack this. I guess StatementExecution needs to make the right calls, and return the return value, probably via a new method, and a higher level object needs to handle the return value to simulate OUT parameter 1 for the other DB types.

I feel like I'm hacking at things at the moment. Do you have any graceful ideas about this?

javornikolov:
No immediate ideas for a simple way to resolve that. Maybe we need to add another layer of abstraction before JDBC.

MMatten:
@javornikolov - I've got things working now for functions that return values via `ResultSet` objects.

Please let me know what you think of the changes to the core. Should I consider sub-classing? Perhaps `StatementExecution`...?

dbfit-java/core/src/main/java/dbfit/api/DbStoredProcedureCall.java
```diff
@@ -51,7 +51,11 @@ void bindParametersTo(StatementExecution cs) throws SQLException {
     }
 
     public StatementExecution toStatementExecution() throws SQLException {
-        StatementExecution cs = new StatementExecution(this.environment.getConnection().prepareCall(toSqlString()));
+System.out.println("DbStoredProcedureCall: toStatementExecution");
```
javornikolov: Do we need that `System.out.println`?

javornikolov:
> @javornikolov - I've got things working now for functions that return values via ResultSet objects.

Great!

> Please let me know what you think of the changes to the core. Should I consider sub-classing? Perhaps StatementExecution...?

I guess we may experiment with a sub-classing alternative and see if it helps.

dbfit-java/core/src/main/java/dbfit/api/DbStoredProcedureCall.java
MMatten: Ah. Definitely not.

MMatten:
> I'm wondering whether we need both `functionReturnValueViaResultSet` and  `discountFunctionReturnValueParameter`.

We could have `functionReturnValueViaResultSet` being `true` implies `discountFunctionReturnValueParameter` being `true` until we ever across a case where they could be different. I was thinking that making them separate makes them more understandable but perhaps updating the Javadoc comment in the interface would be enough. Any thoughts?

MMatten:
> We could have `functionReturnValueViaResultSet` being `true` implies `discountFunctionReturnValueParameter` being `true`...

So, what I meant is that we could keep just `functionReturnValueViaResultSet` and document that this will skip the return value parameter in `CallableStatement`s in the Javadoc for `functionReturnValueViaResultSet()`. Or keep them separate as they are now.

I have no real preference. Which do you think makes things simpler or more understandable?

javornikolov:
Do I understand correctly that' we're still calling functions via `{? = call ...}` but we don't bind an output parameter for the initial "?"? (I.e. - we have more `?` symbols than parameters passed).

MMatten:
> Do I understand correctly that' we're still calling functions via `{? = call ...}` but we don't bind an output parameter for the initial "?"? (I.e. - we have more ? symbols than parameters passed).

That's correct. It's very weird.

MMatten:
Another option could be to do this like we do for `DdlStatementExecution`. I.e. create `public StatementExecution createStatementExecution()` and allow each environment type to return a different subtype if required (an example of this for `DdlStatementExecution` being in the Teradata adapter).

MMatten:
Also, it might be good to move `StatementExecution` from `dbfit.fixture` to `dbfit.util` as per `DdlStatementExecution`. Any view?

javornikolov:
Sounds a good idea to try.

MMatten:
@javornikolov I've just added PR #436 with an alternative approach.

PR #434, opened 2015-10-01
Support non-transactional databases

MMatten:
Do not attempt to change auto-commit behaviour for non-transactional databases.

This change is required to support Informix databases running in Informix rather than ANSI mode.

### DIFF
--- a/dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
+++ b/dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
@@ -126,7 +126,7 @@ public abstract class AbstractDbEnvironment implements DBEnvironment {
 
     public void closeConnection() throws SQLException {
         if (currentConnection != null) {
-            currentConnection.rollback();
+            rollback();
             currentConnection.close();
             currentConnection = null;
         }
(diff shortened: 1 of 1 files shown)


javornikolov:
Looks good to me :+1: 
 - does `Connection.getAutocmmit` return `true` when the database is `Informix` mode (when there is no transaction support)?
 - Does it fail on attempt to execute `setAutocommit(false)`?

MMatten:
> does Connection.getAutocmmit return true when the database is Informix mode (when there is no transaction support)?

Yes. It's the default mode when connecting to an ANSI mode DB too.

>Does it fail on attempt to execute setAutocommit(false)?

Yes. We get an error say "Transactions not supported".

MMatten:
I was thinking of setting the auto-commit option `Options` to `true` too when we detect no transaction support. Any view?

javornikolov:
I wondered about a similar question: what if we have auto-commit = `auto`?

MMatten:
> I wondered about a similar question: what if we have auto-commit = `auto`?

Can you remind me what that does? What happens in `Options` method `public static boolean is(String option)` when the option value is `auto`?

javornikolov:
`is("auto")` will return `false`. Currently we use `Options.get(OPTION_AUTO_COMMIT)` to test for `auto`. The `auto` mode leaves the JDBC driver defaults - it doesn't invoke `setAutoCommit`.

javornikolov:
Anyway - I think the current version looks good enough for now. I'm merging it.

PR #435, opened 2015-10-01
Allow use of CoreTests suite for non-transactional databases

MMatten:
Split out the `CommitAndRollback` test from the `CoreTests` suite so that this suite can be used for non-transactional databases (such as Informix running in Informix, rather than ANSI, mode).


javornikolov:
Is there any difference between `non-transactional` (informix) mode and `autocommit=true`?

FitNesseRoot/DbFit/AcceptanceTests/CoreTests/CommonSuite/CommitAndRollback/content.txt
```diff
@@ -1,35 +0,0 @@
-!|Rollback|
```
javornikolov: As far as I see this file is removed, but not moved to somewhere else. Should it be that way?

FitNesseRoot/DbFit/AcceptanceTests/JavaTests/Db2Tests/FlowMode/properties.xml
```diff
@@ -11,6 +11,7 @@
 <Suite/>
 <Suites>(Shared Tests)</Suites>
 <SymbolicLinks>
+<CommitAndRollback>.DbFit.AcceptanceTests.LibraryTests.CommitAndRollback</CommitAndRollback>
```
javornikolov: Is `LibraryTests` a good name?

javornikolov:
Hmm, interesting: `LibraryTests` folder is missing, there are symbolic links to them in our tests. Yet, TravisCI check have passed successfully...

MMatten:
The moved `CommitAndRollback` test now added again.

FitNesseRoot/DbFit/AcceptanceTests/JavaTests/Db2Tests/FlowMode/properties.xml
MMatten: There's probably a better name out there somewhere. I was lacking inspiration at the time. Any offers accepted!

The idea was to create a library of tests that can be reused/included where appropriate (where the database supports the feature). Half way between the common `CoreTests` and the tests that have been duplicated, for different DB types, that we still have hanging around.

MMatten:
> Is there any difference between non-transactional (informix) mode and autocommit=true?

I think the behaviour is the same.

javornikolov:
> The moved `CommitAndRollback` test now added again.

Great! I still wonder why tests didn't fail providing that this piece was missing?

MMatten:
I'll try running the tests from the FitNesse UI. May be a bug in FitNesse?

javornikolov:
> May be a bug in FitNesse?

Maybe. As far as I see - links pointing to nowhere are not listed in the page content at all.

MMatten:
When I run the tests (e.g. `DerbyTests.FlowMode`) via the web browser interface when the `CommitAndRollback` directory is missing then there are no errors displayed - the suite passes.

We could create an issue on the FitNesse project if we want to follow this up.

javornikolov:
> We could create an issue on the FitNesse project if we want to follow this up.

Sounds like a plan.

PR #436, opened 2015-10-25
Support Informix

MMatten:
Support for Informix (using alternate approach of allowing `StatementExecution` sub-classing instead of adding more parameters to existing class methods and more property getters on the `DBEnvironment` interface).

TODO:
- [x] Simplify testing expected exceptions
- [x] Test the class of the Big Decimal transform() output. (The reason which forced us to have custom handling for Informix).
- [x] Decide about the big decimal to exact java.math.BigDecimal transformer name
- [x] Some refactoring - db accessor set method
- [x] DbParameterAccessor: select the name of the `transform to JDBC - compatible type` method
- [x] Rename `JDBC` to `Jdbc` for the newly created fields and classes

PR #442, opened 2015-11-09
Allow DB-specific type-casting before PreparedStatement object binding

MMatten:
Allow DB environment-specific type casting (of objects normalised by `TypeNormaliser` objects) before passing to `PreparedStatement`s `setObject` method.

For consideration and in support of enhancement of PR #436.

* Units tests still to be created - using a dummy `DBEnvironment` object.
* `StatementExecution` constructors - it probably makes sense to remove the passed `PreparedStatement` object as this could be created via the passed `DBEnvironment` object, BUT, then unit tests will require a DB connection (potentially making the `core` interdependent upon one or more `DBEnvironment` implementations).

### DIFF
--- a/dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
+++ b/dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
@@ -257,5 +257,9 @@ public abstract class AbstractDbEnvironment implements DBEnvironment {
     public void setTypeSpecifier(Class<?> targetClass, TypeSpecifier specifier) {
         typeSpecifiers.put(targetClass, specifier);
     }
+    
+    public Map<Class<?>, TypeSpecifier> getTypeSpecifierMap() {
+        return typeSpecifiers;
+    }
 }
 
--- a/dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
+++ b/dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
@@ -7,6 +7,8 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.sql.*;
 import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.Properties;
@@ -246,5 +248,14 @@ public abstract class AbstractDbEnvironment implements DBEnvironment {
         return (conn != null && !conn.isClosed());
     }
 
+    private Map<Class<?>, TypeSpecifier> typeSpecifiers = new HashMap<Class<?>, TypeSpecifier>();
+
+    public TypeSpecifier getTypeSpecifier(Class<?> targetClass) {
+    	return typeSpecifiers.get(targetClass);
+    }
+
+    public void setTypeSpecifier(Class<?> targetClass, TypeSpecifier specifier) {
+        typeSpecifiers.put(targetClass, specifier);
+    }
 }
 
--- a/dbfit-java/core/src/test/java/dbfit/environment/TypeSpecifierTest.java
+++ b/dbfit-java/core/src/test/java/dbfit/environment/TypeSpecifierTest.java
@@ -4,6 +4,8 @@ import org.junit.*;
 import static org.junit.Assert.*;
 
 import java.sql.SQLException;
+import java.util.Map;
+
 import dbfit.api.DBEnvironment;
 
 import dbfit.util.TypeSpecifier;
@@ -17,18 +19,11 @@ public class TypeSpecifierTest {
     }
 
     @Test
-    public void setAndGetTypeSpecifierTest() {
-        DBEnvironment env = dbfit.api.DbEnvironmentFactory.newEnvironmentInstance("Dummy");
-        env.setTypeSpecifier(Integer.class, new DummyTypeSpecifier());
-        TypeSpecifier ts = env.getTypeSpecifier(Integer.class);
-        assertNotNull(ts);
-    }
-
-    @Test
     public void castTypeTest() {
         DBEnvironment env = dbfit.api.DbEnvironmentFactory.newEnvironmentInstance("Dummy");
         env.setTypeSpecifier(java.util.Date.class, new DummyTypeSpecifier());
-        TypeSpecifier ts = env.getTypeSpecifier(java.util.Date.class);
+        Map<Class<?>, TypeSpecifier> tsm = env.getTypeSpecifierMap();
+        TypeSpecifier ts = tsm.get(java.util.Date.class);
         String s = null;
         try {
             s = (String) ts.specify(new java.util.Date());
--- a/dbfit-java/core/src/test/java/dbfit/environment/TypeSpecifierTest.java
+++ b/dbfit-java/core/src/test/java/dbfit/environment/TypeSpecifierTest.java
@@ -1,0 +1,40 @@
+package dbfit.environment;
+
+import org.junit.*;
+import static org.junit.Assert.*;
+
+import java.sql.SQLException;
+import dbfit.api.DBEnvironment;
+
+import dbfit.util.TypeSpecifier;
+
+public class TypeSpecifierTest {
+
+    private class DummyTypeSpecifier implements TypeSpecifier {
+        public Object specify(Object o) throws SQLException {
+            return new String("abcdef");
+        }
+    }
+
+    @Test
+    public void setAndGetTypeSpecifierTest() {
+        DBEnvironment env = dbfit.api.DbEnvironmentFactory.newEnvironmentInstance("Dummy");
+        env.setTypeSpecifier(Integer.class, new DummyTypeSpecifier());
+        TypeSpecifier ts = env.getTypeSpecifier(Integer.class);
+        assertNotNull(ts);
+    }
+
+    @Test
+    public void castTypeTest() {
+        DBEnvironment env = dbfit.api.DbEnvironmentFactory.newEnvironmentInstance("Dummy");
+        env.setTypeSpecifier(java.util.Date.class, new DummyTypeSpecifier());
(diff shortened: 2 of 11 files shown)


dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
```diff
@@ -246,5 +248,14 @@ private static boolean isConnected(final Connection conn) throws SQLException {
         return (conn != null && !conn.isClosed());
     }
 
+    private Map<Class<?>, TypeSpecifier> typeSpecifiers = new HashMap<Class<?>, TypeSpecifier>();
+
+    public TypeSpecifier getTypeSpecifier(Class<?> targetClass) {
+    	return typeSpecifiers.get(targetClass);
+    }
```
MMatten: Perhaps this method should return the transformed value instead of an object of the class that performs the translation.

javornikolov:
> Allow DB environment-specific type casting (of objects normalised by TypeNormaliser objects) before passing to PreparedStatements setObject method.

I think these are coming not from `TypeNormaliser` but from `*ParseDelegate` processing? Which gets me wondering about the naming (is `Specifiers` clear enough?).


>   Units tests still to be created - using a dummy `DBEnvironment` object.

Which seems to get a bit complicated. We don't depend on the whole environment as far as I see. And if we add new methods to the environment in future - that dummy becomes a burden.

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
javornikolov: Can we make it without adding new methods to `DBEnvironment`? If we pick the change which we had in the Informix PR - the one that moves statement creation in the environment, then we won't need to expose these specifier accessors as public API.

MMatten: Remember in the Informix adapter that I'm reusing/overloading `TypeNormaliserFactory` to cast back to a more specific type (the opposite of normalising) before binding to the `PreparedStatement` and we thought a separate interface would be more appropriate.

It doesn't feel quite right to have this mechanism as a global mapping as it's only required to make specific types for specific adapters.

javornikolov: What about passing type specifier instead of env to the StatementExecution c-tor (or alternatively - to inject it via setter)?

MMatten:
> Can we make it without adding new methods to DBEnvironment?

If it's code that people potentially need to provide to create an adapter shouldn't it be made explicit via the API? It feels like it's hidden away with the sub-classing approach.

MMatten:
> What about passing type specifier instead of env to the StatementExecution c-tor (or alternatively - to inject it via setter)?

I can certainly give this a go and see what it looks like but I would like to understand the case against updating the API/interface (which isn't something I'm particularly familiar with so any information on good practices would be welcomed). At first sight to me the sub-classing seems to make it less obvious what developers need to produce to create an adapter.

javornikolov:
> I can certainly give this a go and see what it looks like but I would like to understand the case against updating the API/interface (which isn't something I'm particularly familiar with so any information on good practices would be welcomed). At first sight to me the sub-classing seems to make it less obvious what developers need to produce to create an adapter.

The public methods are in general about providing interface for other objects to call our methods and is an invitation for more dependencies. (And more dependencies are making things more complicated to maintain). I guess we can keep the method protected. What do you think is less obvious?

MMatten:
Please bear with me, this is new ground for me!


Is anything outside of the `DbFit` code-base ever going to actually call the methods defined in the API? I don't really see what the `DBEnvironment` interface of the API does for us as we always sub-class `AbstractDbEnvironment` anyway to create adapters. *What is the benefit of it?*


The `obviousness` query was because as developers could need to provide sub-classes of `StatementExecution` (as in PR #436) then shouldn't the requirements/specifications of those classes be declared via an API too? It's not obvious what "Create a procedure statement execution object for the given command text" (as per the DBEnvironment Java doc in PR #436) really requires an adapter developer to do (if they need to). *Do you see what I mean?*

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
MMatten: I've made this change - to pass the `TypeSpecifier` map instead of the `DBEnvironment`.

MMatten:
> If we pick the change which we had in the Informix PR - the one that moves statement creation in the environment

I will try this next by applying some of the changes in this PR to PR #436.

javornikolov:
Thanks for raising questions and excuse my delayed and not so clear posts. (I guess I need more sleep :-)).

> Is anything outside of the DbFit code-base ever going to actually call the methods defined in the API?

My point was about calling the methods inside the environment class vs others (let's say DbFit ones).


> I don't really see what the DBEnvironment interface of the API does for us as we always sub-class AbstractDbEnvironment anyway to create adapters. What is the benefit of it?

It provides a common interface to our test fixtures (Update, Insert, etc.). So DBEnvironment specifies the contract which a fixture `needs` to implement (as a minimum) in order to be eligible for plugging it into our test fixtures.
We don't want to use the abstract class there because it may have additional implementation details  - it may make that minimal set of methods unnecessary lager. (Even if it doesn't - it's a bit less obvious).
Possibly people are using DbFit to implement their own proprietary fixtures - it might happen that they're not happy with AbstractDbEnvironment and skip it.


> The obviousness query was because as developers could need to provide sub-classes of StatementExecution (as in PR #436) then shouldn't the requirements/specifications of those classes be declared via an API too?

The initial version of `StatementExecution` was simple and sufficient for all environments without having plans to extend it: separate interface would be overkill. Now that we came to more complicated scenario - it might make sense to extract an interface too. Or we may come up with extracting override-able part with just the part for setting params/getting results only (so we may go back to single StatementExecution).


> It's not obvious what "Create a procedure statement execution object for the given command text" (as per the DBEnvironment Java doc in PR #436) really requires an adapter developer to do (if they need to).

Good point - it may worth to see what we can do about that. In general - sub-classing is not requirement, it may be just creating plain StatementExecution initialized with specific parameters.


In any case - don't take my words as that I'm mandating certain approach. I'm just looking at the surface and sharing some gut feelings and questions out of curiosity.

dbfit-java/core/src/test/java/dbfit/environment/TypeSpecifierTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+    @Test
+    public void castTypeTest() {
+        DBEnvironment env = dbfit.api.DbEnvironmentFactory.newEnvironmentInstance("Dummy");
+        env.setTypeSpecifier(java.util.Date.class, new DummyTypeSpecifier());
+        Map<Class<?>, TypeSpecifier> tsm = env.getTypeSpecifierMap();
+        TypeSpecifier ts = tsm.get(java.util.Date.class);
+        String s = null;
+        try {
+            s = (String) ts.specify(new java.util.Date());
```
javornikolov: Maybe we should rather start from unit testing around StatementExecution as it seems closer to the problem we want to solve.

javornikolov: What would happen if we pass not exactly `java.util.Date` but e.g. a subclass of it?

MMatten: > excuse my delayed and not so clear posts

Of course, no problem, please excuse my vast PRs! And thanks for the explanations above too.

> it may be just creating plain StatementExecution initialized with specific parameters.

I think that's kind of where I started off with PR #432.I'd rather not have to return that that PR now!

> What would happen if we pass not exactly java.util.Date but e.g. a subclass of it?

I will fail to transform the input object as it won't find a `TypeSpecifier`. I've deliberately kept things simple for this PR and didn't create functionality similar to what's in `TypeNormaliser`.

MMatten:
> excuse my delayed and not so clear posts

Of course, no problem, please excuse my vast PRs! And thanks for the explanations above too.

> it may be just creating plain StatementExecution initialized with specific parameters.

I think that's kind of where I started off with PR #432.I'd rather not have to return that that PR now!

> What would happen if we pass not exactly java.util.Date but e.g. a subclass of it?

I will fail to transform the input object as it won't find a `TypeSpecifier`. I've deliberately kept things simple for this PR and didn't create functionality similar to what's in `TypeNormaliser`.

dbfit-java/core/src/test/java/dbfit/environment/TypeSpecifierTest.java
MMatten: > Maybe we should rather start from unit testing around `StatementExecution` as it seems closer to the problem we want to solve.

How about using a Derby or HSQLDB DB for this (unless there's some kind of mocking available)?

MMatten:
>> If we pick the change which we had in the Informix PR - the one that moves statement creation in the environment

> I will try this next by applying some of the changes in this PR to PR #436.

I've ported some of the changes from this PR into #436. Can we switch the discussion back to there (and look at abandoning this PR)?

dbfit-java/core/src/test/java/dbfit/environment/TypeSpecifierTest.java
javornikolov: >> Maybe we should rather start from unit testing around StatementExecution as it seems closer to the problem we want to solve.

> How about using a Derby or HSQLDB DB for this (unless there's some kind of mocking available)?

We have Mockito already available (it's in use in some of the tests). But the thing is currently we're calling many things in this test and I'm not exactly sure what exactly we're testing. (It's good to have smaller and more specific tests - might be several ones /if it fails - it's easier to isolate the reason/).

javornikolov:
> I've ported some of the changes from this PR into #436. Can we switch the discussion back to there (and look at abandoning this PR)?

OK, let's go back.

PR #443, opened 2015-11-14
Generalise TypeNormaliser mechanism

MMatten:
Generalise the `TypeNormaliser` mechanism - to provide a new factory class that can be used for non-global type-casting operations in support of PR #436.

### DIFF
--- a/dbfit-java/core/src/main/java/dbfit/util/DbParameterAccessor.java
+++ b/dbfit-java/core/src/main/java/dbfit/util/DbParameterAccessor.java
@@ -20,8 +20,8 @@ public class DbParameterAccessor {
 
     public static Object normaliseValue(Object currVal) throws SQLException {        
         if (currVal==null) return null;
-        TypeNormaliser tn=TypeNormaliserFactory.getNormaliser(currVal.getClass());
-        if (tn!=null) currVal=tn.normalise(currVal);
+        TypeTransformer tn=TypeNormaliserFactory.getNormaliser(currVal.getClass());
+        if (tn!=null) currVal=tn.transform(currVal);
         return currVal;
     }
 
--- a/dbfit-java/core/src/main/java/dbfit/util/DbParameterAccessor.java
+++ b/dbfit-java/core/src/main/java/dbfit/util/DbParameterAccessor.java
@@ -18,10 +18,14 @@ public class DbParameterAccessor {
     private int position; //zero-based index of parameter in procedure or column in table
     protected StatementExecution cs;
 
-    public static Object normaliseValue(Object currVal) throws SQLException {        
-        if (currVal==null) return null;
-        TypeTransformer tn=TypeNormaliserFactory.getNormaliser(currVal.getClass());
-        if (tn!=null) currVal=tn.transform(currVal);
+    public static Object normaliseValue(Object currVal) throws SQLException {
+        if (currVal == null) {
+            return null;
+        }
+        TypeTransformer tn = TypeNormaliserFactory.getNormaliser(currVal.getClass());
+        if (tn != null) {
+            currVal = tn.transform(currVal);
+        }
         return currVal;
     }
 
--- a/dbfit-java/core/src/main/java/dbfit/util/TypeNormaliserFactory.java
+++ b/dbfit-java/core/src/main/java/dbfit/util/TypeNormaliserFactory.java
@@ -1,50 +1,13 @@
 package dbfit.util;
 
-import java.util.HashMap;
-import java.util.Map;
-
-@SuppressWarnings("unchecked")
 public class TypeNormaliserFactory {
-    private static Map<Class, TypeNormaliser> normalisers = new HashMap<Class, TypeNormaliser>();
+    private static TypeTransformerFactory normaliserFactory = new TypeTransformerFactory();
 
-    private static Class getCandidateIfAncestor(Class target, Class candidate) {
-        return candidate.isAssignableFrom(target) ? candidate : null;
+    public static void setNormaliser(Class<?> targetClass, TypeTransformer normaliser) {
+        normaliserFactory.setTransformer(targetClass, normaliser);
     }
 
-    private static Class getCandidateIfBetter(Class currentBest, Class nextCandidate) {
-        return currentBest.isAssignableFrom(nextCandidate) ? nextCandidate : currentBest;
-    }
-
-    private static Class findClosestAncestor(Class targetClass) {
-        Class currentBest = null;
-
-        for (Class candidate: normalisers.keySet()) {
-            if (currentBest == null) {
-                currentBest = getCandidateIfAncestor(targetClass, candidate);
-            } else {
-                currentBest = getCandidateIfBetter(currentBest, candidate);
-            }
-        }
-
-        return currentBest;
-    }
-
-    public static void setNormaliser(Class targetClass, TypeNormaliser normaliser) {
-        normalisers.put(targetClass, normaliser);
-    }
-
-    public static TypeNormaliser getNormaliser(Class targetClass) {
-        TypeNormaliser normaliser = normalisers.get(targetClass);
-
-        if (normaliser == null) {
-            Class bestCandidate = findClosestAncestor(targetClass);
-
-            if (bestCandidate != null) {
-                normaliser = normalisers.get(bestCandidate);
-                normalisers.put(targetClass, normaliser);
-            }
-        }
-
-        return normaliser;
+    public static TypeTransformer getNormaliser(Class<?> targetClass) {
+        return normaliserFactory.getTransformer(targetClass);
     }
 }
--- a/dbfit-java/core/src/main/java/dbfit/util/TypeTransformer.java
+++ b/dbfit-java/core/src/main/java/dbfit/util/TypeTransformer.java
@@ -1,0 +1,7 @@
+package dbfit.util;
+
+import java.sql.SQLException;
+
+public interface TypeTransformer {
+    public Object transform(Object o) throws SQLException;
+}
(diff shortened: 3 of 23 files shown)


javornikolov:
I like the move towards `no-static-fields` implementation.
The renaming seems pretty widespread affecting many pieces. I'm wondering do we really want it? (If we want it - might be better to slice it as separate commit. If we don't want it - technically, we can keep two interfaces /normaliser and transformer/ and do some wrapping to convert between).

dbfit-java/core/src/main/java/dbfit/util/TypeNormaliserFactory.java
```diff
@@ -1,50 +1,13 @@
...
-        return candidate.isAssignableFrom(target) ? candidate : null;
-    }
-
-    private static Class getCandidateIfBetter(Class currentBest, Class nextCandidate) {
-        return currentBest.isAssignableFrom(nextCandidate) ? nextCandidate : currentBest;
-    }
-
-    private static Class findClosestAncestor(Class targetClass) {
-        Class currentBest = null;
+    private static TypeTransformerFactory ttf = new TypeTransformerFactory();
```
javornikolov: `ttf` is a bit cryptic name :-)

MMatten:
> Just a reminder in case we go on with that and since some may have implemented extensions on top of DbFit: we should bump up DbFit major version (to indicate it's API incompatible with the previous version).

Related to this - we have some interfaces in the `dbfit.util` package. Should they all really be in the `dbfit.api` package?

MMatten:
> The renaming seems pretty widespread affecting many pieces. I'm wondering do we really want it?

The `TypeNormaliser` interface seems to be redundant if we have `TypeTransformer`. If we want to have two interfaces then I guess they should be `TypeNormaliser` and `TypeSpecifier`. They'd really be identical though.

Do you have a preference?

MMatten:
I could simply re-introduce `TypeNormaliser` and cast the types in `TypeNormaliserFactory` as in: -

    public static void setNormaliser(Class<?> targetClass, TypeNormaliser normaliser) {
        ttf.setTransformer(targetClass, (TypeTransformer) normaliser);
    }

    public static TypeNormaliser getNormaliser(Class<?> targetClass) {
        return (TypeNormaliser) ttf.getTransformer(targetClass);
    }

What do you think of this?

javornikolov:
> Related to this - we have some interfaces in the dbfit.util package. Should they all really be in the dbfit.api package?

It might make sense for some, I'm not sure. But that's a good point - `util` package is more or less internal and maybe we can afford more freedom in changing it.

> The `TypeNormaliser` interface seems to be redundant if we have `TypeTransformer`.

Maybe yes. Unless we want to emphasise a difference in their purpose (JDBC -> DbFit or DBFit -> JDBC) - I'm not sure it makes sense to bother with that though.

> If we want to have two interfaces then I guess they should be `TypeNormaliser` and `TypeSpecifier`. They'd really be identical though.

> Do you have a preference?

No strong preference yet. So we have 3 ideas for naming so far - `normaliser`, `specifier`, `transformer`. Within the options I guess we can consider keeping single interface named `normaliser`.

javornikolov:
> I could simply re-introduce TypeNormaliser and cast the types in TypeNormaliserFactory as in: -
> 
> public static void setNormaliser(Class<?> targetClass, TypeNormaliser normaliser) {
>     ttf.setTransformer(targetClass, (TypeTransformer) normaliser);
> }
> 
> public static TypeNormaliser getNormaliser(Class<?> targetClass) {
>     return (TypeNormaliser) ttf.getTransformer(targetClass);
> }
> 
> What do you think of this?

I guess it would be a challenge to cast between two unrelated interfaces (2-way).

MMatten:
> Within the options I guess we can consider keeping single interface named `normaliser`.

I think gone *full circle* then to some extent. If we think it's important to not break dbfit.util then I think I'd lean towards having `TypeNormaliser` and `TypeSpecifier` interfaces but then how do we reuse code for factory classes?

javornikolov:
> I think gone full circle then to some extent. If we think it's important to not break dbfit.util then I think I'd lean towards having TypeNormaliser and TypeSpecifier interfaces but then how do we reuse code for factory classes?

It's possible to use adapter/wrapper (anonymous or permanent) to convert between e.g.:
``` java
new TypeNormaliser() {
    public Object normalise(Object o) {
        return transformer.transform(o);
    }
};
```

But possibly it's not worth the complications. If it's cleaner to go with a single transformer interface, maybe should break the old dbfit.util interface.

javornikolov:
>> Within the options I guess we can consider keeping single interface named normaliser.
> 
> I think gone full circle then to some extent.

Right. Yet, maybe we're fighting with the wrong thing. What was the issue with using the `Normaliser`/`normalise` for both types of conversions? Maybe it's the factory names (or just their instance names) where to address the issue with resolving the confusion (e.g. could be Jdbc2DbfitNormaliserFactory and Dbfit2JdbcNormaliserFactory /a lousy example, we could think for better names/)?

dbfit-java/core/src/main/java/dbfit/util/TypeNormaliserFactory.java
javornikolov: I got wondering if we can get rid of the global static state for that factory too. (I'm not sure yet if this PR is the place. Possibly yes.)

MMatten:
> What was the issue with using the `Normaliser`/`normalise` for both types of conversions? Maybe it's the factory names (or just their instance names)

It's the interface and method name that I dont think are appropriate for the use case where we need to take a [normalised] object and transform it for a specific JDBC driver.

MMatten:
> But possibly it's not worth the complications

Wouldn't we need to change every existing class implementing `TypeNormaliser` though? I.e. to sub-class something.

javornikolov:
> Wouldn't we need to change every existing class implementing TypeNormaliser though? I.e. to sub-class something.

To convert between the interfaces - no need of sub-classing. It's composition and translating the call to the wrapped object (as in the example above).

dbfit-java/core/src/main/java/dbfit/util/TypeNormaliserFactory.java
MMatten: Where could we host an object that would provide the global transformation/mapping mechanism? In the `FitNesseTestHost` singleton?

javornikolov: > Where could we host an object that would provide the global transformation/mapping mechanism? In the FitNesseTestHost singleton?

It's the opposite direction I was thinking about - to avoid global hosting it in a global singleton.

> Sounds like a different PR would make this easier.

Yes, let's try to keep this PR simpler

javornikolov:
> It's the interface and method name that I dont think are appropriate for the use case where we need to take a [normalised] object and transform it for a specific JDBC driver.
> 
> So "normalise" makes sense to me for the the JDBC-to-DbFit case but not the other case.

If transformers makes sense in both contexts: maybe indeed we should get rid of the normaliser interface.

MMatten:
> If transformers makes sense in both contexts: maybe indeed we should get rid of the normaliser interface.

I think retaining two interfaces at this time adds no value. If we need to split them later on we could extend the `TypeTransformer` interface.

So, apart from renaming `ttf, what else needs to happen before we could merge this?

MMatten:
Perhaps we should move the `TypeTransformer` interface to the `dbfit.api` package asap (in another PR?).

dbfit-java/core/src/main/java/dbfit/util/DbParameterAccessor.java
```diff
@@ -20,8 +20,8 @@
 
     public static Object normaliseValue(Object currVal) throws SQLException {        
         if (currVal==null) return null;
-        TypeNormaliser tn=TypeNormaliserFactory.getNormaliser(currVal.getClass());
-        if (tn!=null) currVal=tn.normalise(currVal);
+        TypeTransformer tn=TypeNormaliserFactory.getNormaliser(currVal.getClass());
+        if (tn!=null) currVal=tn.transform(currVal);
```
javornikolov: As we're anyway touching these lines, let's reformat and add space around `=` and `!=`.

javornikolov:
> I think retaining two interfaces at this time adds no value. If we need to split them later on we could extend the TypeTransformer interface.
> 
> So, apart from renaming ttf what else needs to happen before we could merge this?

OK. There is just one minor formatting remark I just added. Besides that - I think we're good to go and merge the PR.

MMatten:
I've taken the liberty of reformatting the other bits of `normaliseValue`. Let me know if this is too much.

dbfit-java/core/src/main/java/dbfit/util/TypeTransformer.java
```diff
@@ -0,0 +1,7 @@
+package dbfit.util;
+
+import java.sql.SQLException;
+
+public interface TypeTransformer {
+    public Object transform(Object o) throws SQLException ;
```
javornikolov: There's an excessive space before `;`

javornikolov:
> I've taken the liberty of reformatting the other bits of normaliseValue. Let me know if this is too much.

Maybe a bit too much but let's keep it.

MMatten:
Thanks.

Updates applied.

javornikolov:
> Perhaps we should move the TypeTransformer interface to the dbfit.api package asap (in another PR?).

I'm not sure about moving to api (there are various opinions on the topic about having interfaces separate). It will need more tweaking around import statements. But if we decide to move to to api, maybe now is the moment.

javornikolov:
Merged :-)

MMatten:
> I'll take your view on this as I'm not *up to speed* (apologies for the awful corporate cliche) regarding "there are various opinions"!

PR #447, opened 2015-11-25
Move type transformation responsibility into TypeTransformerFactory

MMatten:
Move type normalisation out of `DbParameterAccessor` into `TypeTransformerFactory` and combine with selection of the appropriate normaliser class.

### DIFF
--- a/dbfit-java/core/src/main/java/dbfit/util/DbParameterAccessor.java
+++ b/dbfit-java/core/src/main/java/dbfit/util/DbParameterAccessor.java
@@ -19,14 +19,11 @@ public class DbParameterAccessor {
     protected StatementExecution cs;
 
     public static Object normaliseValue(Object currVal) throws SQLException {
-        if (currVal == null) {
-            return null;
+        Object newValue = null;
+        if (currVal != null) {
+            newValue = TypeNormaliserFactory.transform(currVal);
         }
-        TypeTransformer tn = TypeNormaliserFactory.getNormaliser(currVal.getClass());
-        if (tn != null) {
-            currVal = tn.transform(currVal);
-        }
-        return currVal;
+        return newValue;
     }
 
     @Override
--- a/dbfit-java/core/src/main/java/dbfit/util/TypeNormaliserFactory.java
+++ b/dbfit-java/core/src/main/java/dbfit/util/TypeNormaliserFactory.java
@@ -1,5 +1,7 @@
 package dbfit.util;
 
+import java.sql.SQLException;
+
 public class TypeNormaliserFactory {
     private static TypeTransformerFactory normaliserFactory = new TypeTransformerFactory();
 
@@ -7,7 +9,7 @@ public class TypeNormaliserFactory {
         normaliserFactory.setTransformer(targetClass, normaliser);
     }
 
-    public static TypeTransformer getNormaliser(Class<?> targetClass) {
-        return normaliserFactory.getTransformer(targetClass);
+    public static Object transform(Object value) throws SQLException {
+        return normaliserFactory.transform(value);
     }
 }
--- a/dbfit-java/core/src/main/java/dbfit/util/TypeTransformerFactory.java
+++ b/dbfit-java/core/src/main/java/dbfit/util/TypeTransformerFactory.java
@@ -1,5 +1,6 @@
 package dbfit.util;
 
+import java.sql.SQLException;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -32,7 +33,7 @@ public class TypeTransformerFactory {
         transformers.put(targetClass, normaliser);
     }
 
-    public TypeTransformer getTransformer(Class<?> targetClass) {
+    private TypeTransformer getTransformer(Class<?> targetClass) {
         TypeTransformer normaliser = transformers.get(targetClass);
 
         if (normaliser == null) {
@@ -45,5 +46,19 @@ public class TypeTransformerFactory {
         }
 
         return normaliser;
+    }
+
+    public Object transform(Object inValue) throws SQLException {
+        if (inValue == null) {
+            return null;
+        }
+        Object newValue = null;
+        TypeTransformer typeTrans = getTransformer(inValue.getClass());
+        if (typeTrans != null) {
+            newValue = typeTrans.transform(inValue);
+        } else {
+            newValue = inValue;
+        }
+        return newValue;
     }
 }
--- a/dbfit-java/core/src/test/java/dbfit/util/TypeTransformerFactoryTest.java
+++ b/dbfit-java/core/src/test/java/dbfit/util/TypeTransformerFactoryTest.java
@@ -3,62 +3,73 @@ package dbfit.util;
 import org.junit.Test;
 import org.junit.Before;
 
-import java.util.ArrayList;
-import java.util.AbstractCollection;
-import java.util.AbstractList;
+import java.sql.SQLException;
 
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertEquals;
 
 public class TypeTransformerFactoryTest {
-    private static TypeTransformer createTypeTransformerFake(final String tag) {
-        return new TypeTransformer() {
-            @Override
-            public Object transform(Object o) {
-                return null;
-            }
 
-            @Override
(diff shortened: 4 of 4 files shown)


javornikolov:
That's not a `factory` anymore (factories are about creation of object instances).

And in general I have some concerns about **static** methods (`transform`). So far the factory anyway used to have statics so it's not something new. I wonder if it makes sense to do something about that now (can we figure out a better solution). Needs some thought.

``` if (currVal != null) ```
This case I think we can handle that case in the transformation service too (no need to bother param accessor about that).

**Comment by MMatten** [comment]
> That's not a `factory` anymore (factories are about creation of object instances). We may need to rename or create a new class.

Yes, it's creating new objects but based upon transforming existing ones.

I'm not sure of the new names for `TypeNormaliserFactory` and `TypeTransformerFactory`. I can't really just rename `TypeTransformerFactory` to `TypeTransformer` as that's the name of the interface.

**Comment by MMatten** [comment]
> I'm not sure of the new names for `TypeNormaliserFactory` and `TypeTransformerFactory`. I can't really just rename `TypeTransformerFactory` to `TypeTransformer` as that's the name of the interface.

How about something like `TypeNormaliserFactory` -> `ValueNormaliser` and `TypeTransformerFactory` -> `ValueTransformer`?

**Comment by MMatten** [comment]
> And in general I have some concerns about static methods (transform). So far the factory anyway used to have statics so it's not something new. I wonder if it makes sense to do something about that now (can we figure out a better solution). Needs some thought.

If we don't have static fields and methods then where could the normaliser mechanism live/exist? Would we need to host an instance in the test host?

**Comment by MMatten** [comment]
> ...I have some concerns about static methods...can we figure out a better solution...

I'm not knowledgeable on this subject; what do you see as the issue with these?

**Comment by MMatten** [comment]
How about adding a
```java
    private static TypeTransformer typeNormaliser`
```
and new methods
```java
    public Object transform(Object o);
    public void setTransformer(Class<?> c, TypeNormaliser t);
```
 to `FitNesseTestHost` and the associated `TestHost` interface? Are are we just moving the `static`s issue somewhere else?

javornikolov:
http://www.scriptscoop.net/t/d1bec629f421/java-static-methods-are-sort-of-code-smell.html

javornikolov:
> How about something like TypeNormaliserFactory -> ValueNormaliser and TypeTransformerFactory -> ValueTransformer?

What will be the difference between `ValueNormaliser` and `ValueTransformer` after all these modifications?

MMatten:
>> How about something like TypeNormaliserFactory -> ValueNormaliser and TypeTransformerFactory -> ValueTransformer?

> What will be the difference between `ValueNormaliser` and `ValueTransformer` after all these modifications?

This was really a consideration of naming changes only for the current classes.

MMatten:
But thinking more about `ValueXXX` as names, I like them less. 

It's really about changing structure (type) but content is obviously bound to structure.

javornikolov:
> This was really a consideration of naming changes only for the current classes.

I mean - maybe the classes are not so different - if only one would be enough, then maybe we simplify the `how-to-name-it` doubts.

javornikolov:
Maybe we can keep the existing factories and just use them to initialize value transformers.

MMatten:
>>> ...I have some concerns about static methods...can we figure out a better solution...
>> I'm not knowledgeable on this subject; what do you see as the issue with these?

> http://www.scriptscoop.net/t/d1bec629f421/java-static-methods-are-sort-of-code-smell.html

So is dependency injection is the way to go? E.g. supply the normaliser/transformer collection wrapper to each `DbParameterAccessor`, `<dbType>Environment` and the other classes using  either via a setter or via the constructor.

I was thinking perhaps something like the `DatabaseEnvironment` annotation could be use to construct the collection of `TypeTransformers` to load into a `TypeNormaliserFactory` (until renamed) but we wouldn't want normalisers associated with every DB type (just the active one) to be loaded each time we create one of those.

MMatten:
Ok, I've added a few more experimental commits. I'm not sure what you'll make of these!

I've eliminated `TypeNormaliserFactory` by hosting a, now renamed, `TypeTransformerFactory` (now `DbParameterTypeTransformer` aligned with `DbParameterAccessor`) instance in the `FitNesseTestHost`.

Any feedback welcomed!

javornikolov:
What reasons moved you to select `TestHost`/`FitnesseTestHost` for being responsible for the parameters transformation?

MMatten:
This was to just to reduce the consolidate the classes containing static fields and methods. Admittedly, there is little chance of the test system host allowing being the appropriate place for hosting this.

But, I'm not sure what you've really been getting at regarding removing the statics. Can you expand please?

javornikolov:
In general global state makes dependencies a bit more obscure, certain behaviors are harder to test. Also some day we may come to running different tests in parallel - then global state would need special care (synchronized access, etc).

But it's not an absolute rule about removing static stuff (actually we have it in multiple places). It's just a thing to think about since we're anyway changing some things around.

javornikolov:
Some thoughts how it may look like:
* `DbParameterTypeTransformer` - as the implementation is not strictly bound to DbParameter: I'm wonder if we can come up with more generic, yet descriptive name. There are 2 specifics as I look at it:
  - It's kind of compound (allows registering multiple simple transformers into it)
  - Looks up int the class ancestry hierarchy.
* We may hold the instance of the "compound normaliser" in the environments (AbstractEnvironment) 
* Allow environment implementations to register their own normalisers. It might be via some internal protected API registerNormaliser(cass, normaliser) - in order to minimize the future changes (e.g. now a class rename results in renaming thins in several places).
* Now we come to the issue how DbParameterAccessor could access this normaliser... If we decide to get rid of global stuff: we can do as we did for the other transformer (pass it via the constructor; or a setter).
* As we saw - the constructor change was a bit of a burden (all environments need change). We can possibly use a `builder` and/or a `factory` initialized with these common items which we need to pass to DbParameterAccessor c-tor.

MMatten:
Ok. Thanks. That makes sense.

I'll discard the `TestHost` related commits.

What about the names changes: -
* `TypeTransformerFactory` -> `DbParameterTypeTransformer`
* `TypeNormaliserFactory` -> `DbParameterTypeNormaliser` (the thin wrapper over a static `DbParameterTypeTransformer`, not yet drafted in code).

I'm not even sure about the `DbParamaterType` prefix. Perhaps these, and `DbParameterAccessor` should renamed to something like `DbAttributeAccessor` / `DbAttributeTypeTransformer` / `DbAttributeTypeNormaliser`.

javornikolov:
Well, I'm not sure about `DbParameter` prefix either. As I mentioned - it seems a bit more general.

MMatten:
> We may hold the instance of the "compound normaliser" in the environments (AbstractEnvironment)

Each adaptor would need to register the same, common, normalisers such as `BigDecimalNormaliser`. Perhaps these could be fixed ones when we create instances of the transformer collection/compound.

javornikolov:
> Each adaptor would need to register the same, common, normalisers such as BigDecimalNormaliser

We can have the default list of normalisers initialized in `AbstractDbEnvironment`. And then each specific environment may register it's own ones.

MMatten:
We also need to work how to give `Inspect` and `DataRow` access to the normalisers too.

I'm worried about the scope of this PR creeping out of control (and I'm nearly out of time for getting my PRs completed). Are the any compromises we can make now and log the additional stuff as new tech debt issues?

javornikolov:
Yes, I agree. In the Informix PR we got looking into [this part of DbParameterAccessor.set](https://github.com/dbfit/dbfit/pull/436/files#diff-eb2c9113cbfa1c679b99d6cdf964229fR93). The simplest thing I think is just to follow as example what we have in `normaliseValue` and `get`.

MMatten:
> Yes, I agree. In the Informix PR we got looking into this part of DbParameterAccessor.set. The simplest thing I think is just to follow as example what we have in normaliseValue and get.

I'm not quite with you. 

What, from which versions, do you think we should apply to what?

javornikolov:
I was about the current state of PR #436 - in order to simplify DbParameterAccessor.set we can extract part of it in separate method. (More substation refactoring - like this #447, we can perhaps handle after PR #436 is merged).

PR #448, opened 2015-12-02
Allow object transformation to compatible values before binding to JDBC

MMatten:
Allow value (type or content) transformation before binding to JDBC `PreparedStatement`s to allow for JDBC drivers not supporting DbFit sub-classes of common types.


dbfit-java/sqlserver/src/test/java/dbfit/environment/SqlServerDbEnvironmentUnitTests.java
```diff
@@ -1,14 +1,21 @@
 package dbfit.environment;
 
+import java.util.Map;
+import java.util.HashMap;
```
javornikolov: Are we using this Map and HashMap here? Also the`TypeTransformer` import?

javornikolov:
Now that we have that separately - I see we don't have a test covering the case that accessors are transforming the objects (in set method). We have Mockito in use in the project, I guess it won't be hard to setup such a test.

MMatten:
> We have Mockito in use in the project, I guess it won't be hard to setup such a test.

Here's my first go at using Mockito. 

What do you think?

dbfit-java/core/src/test/java/dbfit/util/DbParameterAccessorTest.java
```diff
@@ -0,0 +1,53 @@
...
+        when(ttf.getTransformer(inputValue.getClass())).thenReturn(null);
+
+        StatementExecution se = mock(StatementExecution.class);
+
+        dbpa = new DbParameterAccessor("dummy", Direction.INPUT_OUTPUT, sqlType, userDefinedTypeName, javaType, position, ttf);
+        dbpa.bindTo(se, position);
+        dbpa.set(inputValue);
+        verify(se).setObject(position, inputValue, sqlType, userDefinedTypeName);
+    }
+}
```
javornikolov: In general these tests seem quite long, good to think if we can move some setup part outside, maybe some helper methods.

Yet, maybe first it's good to come up with simpler test - here we exercise `bindTo`, `set`. What if we just test just `set(...)`? I think as part of this refactoring we're not touching anything related to `bindTo`, right?

MMatten: I'm not sure how else I can check to see what the `set()` call has actually done.

Where else can I intercept things using the Mockito bits?

javornikolov: Actually now I looked in more detail and I see we initialize the statement via `bindTo`. So perhaps indeed that's the way to go for the current implementation of the accessor.

MMatten: Great.

Apart from squashing the commits to one, any other observations that need sorting out before merging?

javornikolov: It's good to simplify the tests themselves. The main retain about up to 4-5 statements in the test themselves.
`Direction.INPUT_OUTPUT` can be shortened to `INPUT_OUTPUT` by static import.
Maybe more self-explaining variable names would make sense.

MMatten:
I'm not sure what else I can simplify now.

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
```diff
@@ -1,6 +1,8 @@
 package dbfit.api;
 
 import dbfit.util.*;
+import dbfit.fixture.StatementExecution;
```
javornikolov: I think we're not using the `StatementExecution` import here.

dbfit-java/core/src/test/java/dbfit/util/DbParameterAccessorTest.java
```diff
@@ -0,0 +1,43 @@
...
+    DbParameterAccessor dbpa = new DbParameterAccessor("dummy", INPUT_OUTPUT, sqlType,
+                                                       userDefinedTypeName, javaType, position, inputTransformerFactory);
+
+    public DbParameterAccessorTest() throws Exception {
+        dbpa.bindTo(statement, position);
+    }
+
+    @Test
+    public void inputObjectIsTransformedBeforeJdbcBindingWhenTransformerSpecifierTest() throws Exception {
+
```
javornikolov: What do you think about removing these blank lines?

dbfit-java/oracle/src/main/java/dbfit/util/OracleDbParameterAccessor.java
```diff
@@ -1,11 +1,13 @@
 package dbfit.util;
 
+import java.util.Map;
```
javornikolov: Are we using that `Map`?

javornikolov:
The tests look good enough to me now. I added a few more minor remarks, mostly about unused imports. I think no other issues.

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
MMatten: Thanks. A leftover from a bygone change some time.

dbfit-java/oracle/src/main/java/dbfit/util/OracleDbParameterAccessor.java
MMatten: No. Another relic from when I was using `Map<Class<?>, TypeTransformer>` as the collection of transformers.

dbfit-java/core/src/test/java/dbfit/util/DbParameterAccessorTest.java
MMatten: I'm easy. I'll remove them.

I'm not sure about the line after the class declaration though. We seem to have no standard. Some are followed by a blank and some not, even some of the newer classes added.

MMatten:
Ok, applied updates and squashed commits.

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
```diff
@@ -1,6 +1,7 @@
 package dbfit.api;
 
 import dbfit.util.*;
+
```
javornikolov: No need to add this empty line.

dbfit-java/core/src/test/java/dbfit/util/DbParameterAccessorTest.java
```diff
@@ -0,0 +1,41 @@
...
+    int position = 1;
+    Integer outputValue = 5;
+    String userDefinedTypeName = "whatever";
+    TypeTransformer inputTransformer = mock(TypeTransformer.class);
+    TypeTransformerFactory inputTransformerFactory = mock(TypeTransformerFactory.class);
+    StatementExecution statement = mock(StatementExecution.class);
+    DbParameterAccessor dbpa = new DbParameterAccessor("dummy", INPUT_OUTPUT, sqlType,
+                                                       userDefinedTypeName, javaType, position, inputTransformerFactory);
+
+    public DbParameterAccessorTest() throws Exception {
```
javornikolov: @Before method should be used instead of constructor for such initializations. (Maybe part of the attributes are good to move in such a block too - since some are just temporary and not used in the tests).

MMatten: Is `@Before` better than `@BeforeClass`?

MMatten: For the main initialisation stuff, not what's currently in the constructor.

MMatten:
How about this approach?

javornikolov:
Nice, it looks good now! (In other places seems we named the Before method `prepare` but that doesn't matter). So I'm merging this :-)

PR #450, opened 2015-12-18
Allow escaped column and table names  in postgres.

igieon:
Allow to have table names like public."Pos++Pers" and column names like "Pers++Pos+id"

### DIFF
--- a/dbfit-java/postgres/src/integration-test/resources/002_escapped_names.sql
+++ b/dbfit-java/postgres/src/integration-test/resources/002_escapped_names.sql
@@ -1,0 +1,6 @@
+CREATE TABLE "Pers+""+Pos" (
+  "Pers++Pos+id" SERIAL PRIMARY KEY,
+  "user" VARCHAR(50) UNIQUE, 
+  "Info""+desc" VARCHAR(50)
+)
+;
--- a/dbfit-java/postgres/src/main/java/dbfit/environment/PostgresEnvironment.java
+++ b/dbfit-java/postgres/src/main/java/dbfit/environment/PostgresEnvironment.java
@@ -2,6 +2,7 @@ package dbfit.environment;
 
 import dbfit.annotations.DatabaseEnvironment;
 import dbfit.api.AbstractDbEnvironment;
+import dbfit.environment.postgres.NameNormaliserPostgres;
 import dbfit.util.DbParameterAccessor;
 import dbfit.util.Direction;
 import dbfit.util.NameNormaliser;
@@ -44,15 +45,14 @@ public class PostgresEnvironment extends AbstractDbEnvironment {
 
     public Map<String, DbParameterAccessor> getAllColumns(String tableOrViewName)
             throws SQLException {
-        String[] qualifiers = NameNormaliser.normaliseName(tableOrViewName)
-                .split("\\.");
+        String[] qualifiers = tableOrViewName.split("\\.");
         String qry = " select column_name, data_type, character_maximum_length "
                 + "	as direction from information_schema.columns where ";
 
         if (qualifiers.length == 2) {
-            qry += " lower(table_schema)=? and lower(table_name)=? ";
+            qry += " table_schema=? and table_name=? ";
         } else {
-            qry += " (table_schema=current_schema() and lower(table_name)=?)";
+            qry += " (table_schema=current_schema() and table_name=?)";
         }
         qry += " order by ordinal_position";
         return readIntoParams(qualifiers, qry);
@@ -63,7 +63,7 @@ public class PostgresEnvironment extends AbstractDbEnvironment {
         try (PreparedStatement dc = currentConnection.prepareStatement(query)) {
             for (int i = 0; i < queryParameters.length; i++) {
                 dc.setString(i + 1,
-                        NameNormaliser.normaliseName(queryParameters[i]));
+                        NameNormaliserPostgres.normaliseName(queryParameters[i]));
             }
             ResultSet rs = dc.executeQuery();
             Map<String, DbParameterAccessor> allParams = new HashMap<String, DbParameterAccessor>();
@@ -72,9 +72,11 @@ public class PostgresEnvironment extends AbstractDbEnvironment {
                 String paramName = rs.getString(1);
                 if (paramName == null)
                     paramName = "";
+                //fix escaping
+                paramName = paramName.replace("\"","\"\"");
                 String dataType = rs.getString(2);
                 DbParameterAccessor dbp = createDbParameterAccessor(
-                        paramName,
+                        '"' + paramName + '"',
                         Direction.INPUT, getSqlType(dataType),
                         getJavaClass(dataType), position++);
                 allParams.put(NameNormaliser.normaliseName(paramName), dbp);
--- a/dbfit-java/postgres/src/main/java/dbfit/environment/postgres/NameNormaliserPostgres.java
+++ b/dbfit-java/postgres/src/main/java/dbfit/environment/postgres/NameNormaliserPostgres.java
@@ -1,0 +1,23 @@
+package dbfit.environment.postgres;
+
+public class NameNormaliserPostgres {
+    private NameNormaliserPostgres() {
+        // utility classes should not be instanciated
+    }
+
+    private static boolean isEscaped(String name) {
+        // http://www.postgresql.org/docs/9.1/static/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS
+        return name.charAt(0) == '"' && name.charAt(name.length() - 1) == '"';
+    }
+
+    public static String normaliseName(final String name) {
+        if (name == null || name.isEmpty()) {
+            return "";
+        }
+
+        if (isEscaped(name)) {
+            return name.substring(1, name.length() - 1).replace("\"\"", "\"");
+        } 
+        return dbfit.util.NameNormaliser.normaliseName(name);
+    }
+}
--- a/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/PostgresTests/FlowMode/SimpleUpdate2/content.txt
+++ b/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/PostgresTests/FlowMode/SimpleUpdate2/content.txt
@@ -1,0 +1,15 @@
+|insert|Users|
+|Name|Username|
+|arthur dent|adent|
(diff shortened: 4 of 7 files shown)


dbfit-java/postgres/src/main/java/dbfit/environment/PostgresEnvironment.java
```diff
@@ -44,15 +45,14 @@ protected String parseCommandText(String commandText) {
...
             throws SQLException {
-        String[] qualifiers = NameNormaliser.normaliseName(tableOrViewName)
-                .split("\\.");
+        String[] qualifiers = tableOrViewName.split("\\.");
         String qry = " select column_name, data_type, character_maximum_length "
                 + "	as direction from information_schema.columns where ";
 
         if (qualifiers.length == 2) {
-            qry += " lower(table_schema)=? and lower(table_name)=? ";
+            qry += " table_schema=? and table_name=? ";
```
javornikolov: Are we making tables case-sensitive here? Are the other tests still working fine, will both below succeed:
```
|update|users|
|username=|name|
|adent2|arthur dent|
```

```
|UPDATE|USERS|
|USERNAME=|NAME|
|adent2|arthur dent|
```

dbfit-java/postgres/src/main/java/dbfit/environment/PostgresEnvironment.java
```diff
@@ -63,7 +63,7 @@ protected String parseCommandText(String commandText) {
         try (PreparedStatement dc = currentConnection.prepareStatement(query)) {
             for (int i = 0; i < queryParameters.length; i++) {
                 dc.setString(i + 1,
-                        NameNormaliser.normaliseName(queryParameters[i]));
+                		NameNormaliserPostgres.normaliseName(queryParameters[i]));
```
javornikolov: The convention for indentation is 4 spaces, not tabs.

dbfit-java/postgres/src/main/java/dbfit/environment/PostgresEnvironment.java
```diff
@@ -74,7 +74,7 @@ protected String parseCommandText(String commandText) {
                     paramName = "";
                 String dataType = rs.getString(2);
                 DbParameterAccessor dbp = createDbParameterAccessor(
-                        paramName,
+                		'"'+paramName+'"',
```
javornikolov: Is it necessary to enclose that in `"` here? So far the behavior has been to store the original name in the parameter accessors.

javornikolov: Formatting notes: spaces are missing around `+`; and indentation should be spaces.

dbfit-java/postgres/src/main/java/dbfit/environment/postgres/NameNormaliserPostgres.java
```diff
@@ -0,0 +1,23 @@
+package dbfit.environment.postgres;
+
+public class NameNormaliserPostgres {
+    private NameNormaliserPostgres() {
+        // utility classes should not be instanciated
+    }
+
+
```
javornikolov: That empty line should should be rather moved after the closing `}` of `isEscaped`.

dbfit-java/postgres/src/main/java/dbfit/environment/postgres/NameNormaliserPostgres.java
```diff
@@ -0,0 +1,23 @@
...
+    private static boolean isEscaped(String name) {
+    	// http://www.postgresql.org/docs/9.1/static/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS
+    	return name.charAt(0) == '"' && name.charAt(name.length() - 1) == '"';
```
javornikolov: This would raise `IndexOutOfBoundsException` if name is empty string - is this what we want or we rather expect `false` in such case?

dbfit-java/postgres/src/main/java/dbfit/environment/postgres/NameNormaliserPostgres.java
```diff
@@ -0,0 +1,23 @@
...
+    }
+    public static String normaliseName(final String name) {
+        if (name == null) {
+            return "";
+        }
+
+        if (isEscaped(name)) {
+        	return name.substring(1,name.length() - 1).replace("\"\"", "\"");
```
javornikolov: Missing space after `1`.

javornikolov: Why do we need the `replace` at the end? It may be good to add some unit tests for this class.

dbfit-java/postgres/src/integration-test/resources/002_escapped_names.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE "Pers++Pos" (
+  "Pers++Pos+id" SERIAL PRIMARY KEY,
+  "user" VARCHAR(50) UNIQUE, 
+  "Info+desc" VARCHAR(50)
+)
+;
```
javornikolov: Git shows `No new line at end of file` warning here. Is it possible to get rid of it?

javornikolov:
Thank you @igieon for submitting this pull request! Looks like a useful feature, just needs some polishing. I just posted a few comments inline.

dbfit-java/postgres/src/main/java/dbfit/environment/postgres/NameNormaliserPostgres.java
igieon: I add check.

javornikolov: > I add check.

Alternatively you  may look at `String` built-in methods - there are other ways to check what's at start/end.

dbfit-java/postgres/src/main/java/dbfit/environment/PostgresEnvironment.java
igieon: Just to put in correct way if you create table or columns without " postgres automatically convert it to lower case adn then it will by stored in information schema. I added that test into test suite.

dbfit-java/postgres/src/main/java/dbfit/environment/postgres/NameNormaliserPostgres.java
igieon: We need replace because "" is used to to escape ". I change table name in test to Pers+"+Pos

dbfit-java/postgres/src/main/java/dbfit/environment/PostgresEnvironment.java
javornikolov: > Just to put in correct way if you create table or columns without " postgres automatically convert it to lower case adn then it will by stored in information schema. I added that test into test suite.

Have you pushed that test, I don't see it yet?

igieon: Now is pushed

javornikolov: OK... Update actually works a bit different than e.g. Insert. It's interesting to check also Insert with both upper and lower case - maybe in the same test can be split to check both scenarios:

```
|insert|users|
|name|username|
|arthur dent|adent|

|INSERT|USERS|
|NAME|USERNAME|
|ford prefect|fpref|
|zaphod beeblebrox|zaphod|
```

Will this work?

igieon: I pushed modified test. For lower case is everything in SimpleUpdate  all modification upper case   are in SimpleUpdate2

dbfit-java/postgres/src/main/java/dbfit/environment/PostgresEnvironment.java
```diff
@@ -74,7 +74,7 @@ protected String parseCommandText(String commandText) {
                     paramName = "";
                 String dataType = rs.getString(2);
                 DbParameterAccessor dbp = createDbParameterAccessor(
-                        paramName,
+                        '"' + paramName + '"',
```
javornikolov: I wonder if we can avoid escaping that here (since it has always been the original name unmodified)?

igieon: I just added one new test for this. We need to put there escapped character because its column name and we need to escape collumn name, because in information schema is alway unescapped stored. Meaning that if you use unexcapped character is column always stored as lower case and we can escaped it.

javornikolov: > I just added one new test for this. We need to put there escapped character because its column name and we need to escape collumn name, because in information schema is alway unescapped stored. Meaning that if you use unexcapped character is column always stored as lower case and we can escaped it.

Yes, but what I was wondering is if we can do the escaping somewhere else? I guess for `buildInsertCommand` we can easily handle since it's inside environment class; for `Update` - hmm, maybe we can introduce some general ability for case-sensitive escaping (there is similar issue for other database types). I'm not sure if anything else is affected?

igieon: For me when i look into code everything could be change if we modify NameNormaliser and little refactor code around it. As i look now my change doesn't affect weird function parameters. I want just implement this case, because we need to transform old weird naming schema to new one understandable views  and we want to have covered with tests ;-).

javornikolov: Right. I think the current change is well isolated (only plsql code is affected) and is working good enough. We can go with it as initial version and think of possible refactoring later.

javornikolov:
Merging... @igieon, thanks a lot for that contribution!

javornikolov:
Just a final touch: @igieon, could you squash your commits into single one before we merge the PR?

igieon:
Squashed

javornikolov:
> Squashed.

Can we just have the commit message description refined a bit (says `Add more tests.` right now).

> Just the question when you make new release ?

I think we can make it pretty soon. Do you have any preference?

igieon:
Could be like that ?

javornikolov:
Yes, it looks good now :+1: . Thank you!

PR #451, opened 2015-12-27
Allow customisation of RDBMS function/procedure execution

MMatten:
Allow use of `StatementExecution` subtypes/customisations, with a particular view of the upcoming Informix adapter.

### DIFF
--- a/dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
+++ b/dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
@@ -1,6 +1,7 @@
 package dbfit.api;
 
 import dbfit.util.*;
+import dbfit.fixture.StatementExecution;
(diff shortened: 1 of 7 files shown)


javornikolov:
Thanks @MMatten! The commit message sounds a bit unclear to me (`subtyping` is maybe a bit confusing/not enough). Would it be possible to refine it?

http://chris.beams.io/posts/git-commit/

MMatten:
I've updated it. How does it look now?

MMatten:
Note that I haven't included all of the changes to `StatementExecution` from PR #436.

These other changes were (also) driven by the requirements for Informix. Do you think I should include these in this PR?

javornikolov:
> I've updated it. How does it look now?
Thanks. It looks good now :+1: 

> These other changes were (also) driven by the requirements for Informix. Do you think I should include these in this PR?

No need to include these here. If there is something general enough - another PR may be extracted. So I'm merging this one as is.

PR #452, opened 2015-12-28
Support IBM Informix

### LINKED ISSUE #331: Procedure/Function with return value  for InformixWrapper
I have created a function mentioned below. 
create  function `Multiply(n1 integer, n2 integer) RETURNING  integer  ; return n1*n2; end function;`
In case of oracle and mysql is is showing out parameter as RETURN,In case of Informix  it is showing empty.
In Informix: After digged into the code ,exceuting procedure is fine and while setting return value in UI It is failing.
Can you please assist.

### LINKED ISSUE #423: Add support for Informix
I'm creating a new central issue for adding Informix support to tie together references from: -

* Issue #331.
* PR #326.
* Branch `install-informix-if-available`.

MMatten:
Adds a full adapter implementation to add support for IBM Informix: -
* Java code.
* Acceptance Tests.
* Doc updates.
* Informix Developer Edition shell script installer/deinstaller.
* Build task updates.
* Test VM provisioner for the Informix Developer Edition shell script installer.

closes #432 closes #436 closes #326 closes #442
resolves #423 resolves #331


MMatten:
Note that I still have to run the full set of adaptor regression tests but I've created the PR now to get some visibility and reviews asap.

FitNesseRoot/DbFit/AcceptanceTests/CoreTests/StoredProcNoParams/content.txt
```diff
@@ -1,5 +1,9 @@
 !3 If there are no parameters, Execute Procedure needs just one row
 
+|Clean|
+|table|clean?|
+|users|true|
```
javornikolov: That perhaps means that some other tests don't perform tear down at the end. Do we know which ones?

MMatten:
I've run a full build now over all adapters and it's all looking good.

MMatten:
I've recently added the missing regression test class `InformixRegressionTest.java`.

FitNesseRoot/DbFit/AcceptanceTests/CoreTests/StoredProcNoParams/content.txt
MMatten: The `Users` table is persistant. DML usually gets rolled back, automatically in flow mode, and explicitly via the `Rollback` in the stand alone mode tests.

The Informix tests, running in Informix mode, auto commit so the data hangs around and causes a problem for the next time the test is run (in either of the modes).

javornikolov: I wonder if we can add a `TearDown` page to `CoreTests` which to do the necessary cleanup. (And include in case of Informix; and possibly other similar cases in future).

javornikolov: I took a bit more detailed look at core tests - seems like all other tests except `StoredProcNoParams` are already taking care to tear down their changes. So perhaps it's OK to keep the cleanup here in `StoredProcNoParams` but I wold rather vote to move it to the end of the test page (not in the beginning).

MMatten: All the other `CoreTests` look like they create and drop the tables as part of the test page.

I'll move the `Clean` to the end of the test page.

It's an interesting point of course. When I create tests I often clean tables at the start of the test page to ensure that there's nothing hanging around from previous activities (DbFit tests or manual testing) and I guess one could consider initialising the database objects (to empty) to be set-up as well as tear-down work.

MMatten: The `Users` table is also populated by `CommonTests.BulkInsertWithReturning` so I'll also have to change this test to clean up after itself (as we want the `Clean`s at the end of the test pages).

MMatten:
Updates have been applied - `Users` now cleaned in both tests at the end of the page. I've updated the 2 separate commits.

javornikolov:
OK, great!

[StatementExecution customizations](5a926b71bc0a61a659a55df913faf03444f916ae) - I'd rather vote for one more separate PR for that.

Maybe for [Add missing whitespace in function execution text](af79821ffc4d7170b3d5925473f51f4bcd40c00d) too.

MMatten:
> StatementExecution customizations - I'd rather vote for one more separate PR for that.

I've created #453.

MMatten:
I've removed the commits for the recent PRs we've peeled off from this one.
All retested and ready for re-review.

javornikolov:
In [Clean previous test's data for StoredProcNoParams](36e7c6ff7160c80d0141182276759d24193705bc) we have the clean in the beginning and in the other commit we move at at the end. We can either move the final change (commit at the end) in the separate commit; or we can squash the two commits into one.

dbfit-java/informix/src/main/java/dbfit/environment/BigDecimalExactClassTransformer.java
```diff
@@ -0,0 +1,21 @@
+package dbfit.environment;
+
+import java.sql.SQLException;
+
+import dbfit.util.TypeTransformer;
+
+public class BigDecimalExactClassTransformer implements TypeTransformer {
+
+    @Override
+    public Object transform(Object o) throws SQLException {
```
javornikolov: BTW  - no statements in the implementation are throwing `SQLException` so I think we can remove that `throws` clause here (and the related import).

javornikolov:
I reviewed the PR - besides my comments above it looks OK. Let's tidy up these bits and we can finally merge it :-)

MMatten:
> In Clean previous test's data for StoredProcNoParams we have the clean in the beginning and in the other commit we move at at the end.

I've messed that up somehow. I'll fix it up and squash.

MMatten:
> In Clean previous test's data for StoredProcNoParams we have the clean in the beginning and in the other commit we move at at the end. We can either move the final change (commit at the end) in the separate commit; or we can squash the two commits into one.

Actually, the `Clean` is already at the end: https://github.com/dbfit/dbfit/blob/support-informix/FitNesseRoot/DbFit/AcceptanceTests/CoreTests/StoredProcNoParams/content.txt

I'm not sure what you mean.

MMatten:
> In Clean previous test's data for StoredProcNoParams we have the clean in the beginning and in the other commit we move at at the end. We can either move the final change (commit at the end) in the separate commit; or we can squash the two commits into one.

Right. I see what you mean now...

MMatten:
I've created some replacement commits to repackage the `StoredProcNoParams` change into only 1 commit.

I've also removed the redundant `throws SQLException` clauses too from `BigDecimalExactClassTransformer` and the associated test class.

javornikolov:
> I've created some replacement commits to repackage the StoredProcNoParams change into only 1 commit.

The content is OK now but the commit message first line doesn't quite reflect what it does.

> I've also removed the redundant throws SQLException clauses too from BigDecimalExactClassTransformer and the associated test class.

Looks nice now!

MMatten:
I've updated that commit message. What do you think?

javornikolov:
Looks good now :shipit:, thanks!
I'll take a final look and merge it.

MMatten:
> I'll take a final look and merge it.

Fingers crossed!

dbfit-java/informix/src/integration-test/java/dbfit/environment/InformixRegressionTest.java
```diff
@@ -0,0 +1,30 @@
+package dbfit.environment;
+
+import fitnesse.junit.FitNesseRunner;
+import org.junit.Ignore;
```
javornikolov: This `Ignore` import seems unused.

dbfit-java/informix/src/main/java/dbfit/environment/InformixEnvironment.java
```diff
@@ -0,0 +1,319 @@
+package dbfit.environment;
+
+import dbfit.annotations.DatabaseEnvironment;
+import dbfit.api.AbstractDbEnvironment;
+import dbfit.fixture.StatementExecution;
+import dbfit.util.DbParameterAccessor;
+import dbfit.util.DdlStatementExecution;
```
javornikolov: This import seems unused.

dbfit-java/informix/src/main/java/dbfit/environment/InformixEnvironment.java
```diff
@@ -0,0 +1,319 @@
...
+import dbfit.util.Direction;
+import dbfit.util.NameNormaliser;
+import dbfit.util.Options;
```
javornikolov: Options import seems unused too.

dbfit-java/informix/src/main/java/dbfit/environment/InformixEnvironment.java
```diff
@@ -0,0 +1,319 @@
...
+import fit.TypeAdapter;
```
javornikolov: Are we using TypeAdapter import?

javornikolov:
Ah... I managed to spot a few unused imports. Can we tidy them up.

MMatten:
Redundant imports removed now.

javornikolov:
Nice.
Last bit to tackle: I think some `InformixEnvironment` methods can be annotated with `@Override` for a bit more clarity.

dbfit-java/informix/src/main/java/dbfit/environment/InformixEnvironment.java
```diff
@@ -0,0 +1,309 @@
...
+        super(driverClassName);
+        dbfitToJdbcTransformerFactory.setTransformer(dbfit.util.NormalisedBigDecimal.class, new BigDecimalExactClassTransformer());
+    }
+
+    protected String parseCommandText(String commandText) {
+        commandText = commandText.replaceAll(paramNamePattern, "?");
+        return super.parseCommandText(commandText);
+    }
+
+    private static String paramNamePattern = "[@:]([A-Za-z0-9_]+)";
```
javornikolov: May we move the attributes before the methods.

MMatten: Would you suggest laying out all of the attributes and then all of the methods, or keeping the attributes just before the method that uses them (if only one)?

javornikolov: Both have their merits. In general there are some guidelines e.g. here: http://www.oracle.com/technetwork/java/javase/documentation/codeconventions-141855.html#11684

But if it seems cleaner to have the field close to method using it - perhaps there is no need to be dogmatic about the layout. In the particular case - I think these two are close enough even if we put them in the beginning before all methods.

Another observation: constant values may be good to declare as final (helps for some compiler optimizations sometimes).

MMatten:
> Another observation: constant values may be good to declare as final (helps for some compiler optimizations sometimes).

I can apply this to the data type `List`s then too I guess.

javornikolov:
>> Another observation: constant values may be good to declare as final (helps for some compiler optimizations sometimes).

> I can apply this to the data type Lists then too I guess.

Yes.

MMatten:
Updates applied.

dbfit-java/informix/src/main/java/dbfit/environment/InformixEnvironment.java
```diff
@@ -0,0 +1,317 @@
...
+
+    private static final String paramNamePattern = "[@:]([A-Za-z0-9_]+)";
+
+    @Override
+    protected String parseCommandText(String commandText) {
+        commandText = commandText.replaceAll(paramNamePattern, "?");
+        return super.parseCommandText(commandText);
+    }
+
+    private static Pattern paramRegex = Pattern.compile(paramNamePattern);
```
javornikolov: Can we mark that final too?

MMatten: Ah, yes. Good spot.

MMatten:
Updated.

MMatten:
Worth consolidating this:

```java
    private static String normaliseTypeName(String dataType) {
        dataType = dataType.toUpperCase().trim();
        return dataType;
```

to:

```java
    private static String normaliseTypeName(String dataType) {
        return dataType.toUpperCase().trim();
```
?

javornikolov:
Yes, obviously `dataType.toUpperCase().trim()` seems better. (Breaking into 2 lines doesn't make it shorter nor clearer).

MMatten:
Updated.

dbfit-java/informix/.gitignore
```diff
@@ -0,0 +1 @@
+/bin/
```
javornikolov: BTW - why we need this .gitignore for `/bin`?

MMatten: That's a very good question. I'm not sure how that's sneeked in.

MMatten: > That's a very good question. I'm not sure how that's sneeked in.

Now removed.

dbfit-java/informix/TestDbConnectionDbFitInformixINFORMIX.properties
```diff
@@ -0,0 +1,4 @@
+service=localhost:16466
+username=dbfit
+password=DBFIT
+database=dbfit_ifx
```
javornikolov: I think we should remove these `.properties` from version control too and keep `.default` scripts only? `.gitignore` should be updated (similarly to Oracle props entries which we have now).

MMatten: Right. I'll add the support for the `.custom` properties files for Informix too. They're not there in the `build.gradle` script.

MMatten:
> Right. I'll add the support for the `.custom` properties files for Informix too. They're not there in the `build.gradle` script.

Done.

dbfit-java/informix/src/main/java/dbfit/environment/BigDecimalExactClassTransformer.java
```diff
@@ -0,0 +1,19 @@
...
+    @Override
+    public Object transform(Object o) {
+        if (o == null) {
+            return null;
+        }
+        if (!(o instanceof java.math.BigDecimal)) {
+            throw new UnsupportedOperationException(this.getClass().getSimpleName() +
+                    " cannot transform objects of type " + o.getClass().getName());
+        }
+        java.math.BigDecimal bd = new java.math.BigDecimal(((java.math.BigDecimal) o).toString());
```
javornikolov: I think we can return that right here (no need of the temp var bd).

MMatten:
> I think we can return that right here (no need of the temp var bd).

Updated.

javornikolov:
Cool. That was it. I'm merging the PR :-)

@MMatten thank you very much for this contribution and for your persistence to drive it to completion! Sorry about my comments have been quite scattered, sometimes obscure and coming in too small bits.

MMatten:
> @MMatten thank you very much for this contribution and for your persistence to drive it to completion! Sorry about my comments have been quite scattered, sometimes obscure and coming in too small bits.

@javornikolov - thank you for your reviews, guidance, tutoring and of course infinite patience! I think you would probably have done this yourself in a tenth of the time!

PR #456, opened 2015-12-31
Update FitNesse to 20151230

javornikolov:


javornikolov:
@MMatten , would you be able to test that on the full list of databases? (My environment is a bit limited and I can check the in-memory ones only now).

MMatten:
> @MMatten , would you be able to test that on the full list of databases? (My environment is a bit limited and I can only check the embedded ones now).

Certainly. I do it now.

MMatten:
I've run an `integrationBuild` over all of the DB types - success.

javornikolov:
Perfect. Thanks for checking!

PR #463, opened 2016-01-01
Remove clearParameters from StatementExecution c-tor

javornikolov:
**NOT READY TO BE MERGED YET** (consider other ideas for improving the statement creation API)

 - [x] Remove `clearParameters` parameter from StatementExecution c-tor as it seems superfluous for now
 - [ ] Consider passing statement text instead of PreparedStatement to createStatementExecution?
 - [ ] Consider `DBEnvironment` directly returning `StatementExecution` instead of (or in addition to) raw `PreparedStatement` via `getConnection().prepare{Call|Statement}`
 - [ ] What about moving `StatementExecution` to either `util` or `api` package?


MMatten:
> What about moving `StatementExecution` to either util or api package?

As a replacement for a `java.sql.preparedStatement` being in the `dbfit.fixture` package does seem like the wrong place. As developers might need to know exactly the expected behaviours of a `StatementExecution` would the `api` package be better?

javornikolov:
> As developers might need to know exactly the expected behaviours of a StatementExecution would the api package be better?

Both `api` and `util` seem better. Possibly we can make `StatementExecution` just an interface (especially if moved to `api`).

MMatten:
> Consider passing statement text instead of PreparedStatement to createStatementExecution?

I guess then `createStatementExecution` will need to know if the generated `StatementExecution` needs to contain a `PreparedStatement` or a `CallableStatement`.

javornikolov:
>>  Consider passing statement text instead of PreparedStatement to createStatementExecution?

> I guess then createStatementExecution will need to know if the generated StatementExecution needs to contain a PreparedStatement or a CallableStatement.

Yes... That seems to lead to multiple methods for creating statement executions.

MMatten:
> Yes... That seems to lead to multiple methods for creating statement executions.

Could we get away with something like a `boolean statementIsCallable` param in `createStatementExecution`? Not required on `createFunctionStatementExecution`.

javornikolov:
> Could we get away with something like a boolean statementIsCallable param in createStatementExecution? Not required on createFunctionStatementExecution.

Maybe. I was wondering about the same for function vs non-function sp calls. I just pushed experimental draft commit where raw SQL statements were replaced with `StatementExecution` (using separate the approach with separate methods instead of `boolean` flag). I'm not sure yet if this refactoring is good idea, feedback and other ideas are welcome.

MMatten:
> I'm not sure yet if this refactoring is good idea, feedback and other ideas are welcome.

Off topic a little but...if we're going to move DbFit to 4.0.0 because of API changes coming out of the Informix related PRs then I guess it would be good to sort out any known API factoring issues before releasing. Do you know if anything else needs sorting out?

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
```diff
@@ -121,19 +121,44 @@ public final PreparedStatement createStatementWithBoundFixtureSymbols(
     }
 
     @Override
+    public final StatementExecution createStatementExecutionWithBoundFixtureSymbols(TestHost th, String commandText)
+            throws SQLException {
+        return createStatementExecution(createStatementWithBoundFixtureSymbols(th, commandText));
```
MMatten: We could merge `createStatementWithBoundFixtureSymbols` into this method too I think.

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
```diff
@@ -121,19 +121,44 @@ public final PreparedStatement createStatementWithBoundFixtureSymbols(
...
+    @Override
     public DdlStatementExecution createDdlStatementExecution(String ddl)
             throws SQLException {
         return new DdlStatementExecution(getConnection().createStatement(), ddl);
     }
 
+    /**
+     * Create a procedure statement execution object for the given prepared statement
+     */
+    protected StatementExecution createStatementExecution(PreparedStatement statement) {
```
MMatten: Would it be desirable to have this method accepting the SQL statement text (like `createDdlStatementExecution`)?

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
```diff
@@ -121,19 +121,44 @@ public final PreparedStatement createStatementWithBoundFixtureSymbols(
...
+        return new StatementExecution(statement);
+    }
+
+    /**
+     * Create a function statement execution object for the given prepared statement
+     */
+    protected StatementExecution createFunctionStatementExecution(PreparedStatement statement) {
```
MMatten: Would it be desirable to have this method accepting the SQL statement text (like `createDdlStatementExecution`)?

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
javornikolov: Right, the `PreparedStatement` variants can be made protected or merged into the new versions. So far I removed them from the interface. On the other extreme - there is an option to keep the old versions into the interface, maybe marking them `@Deprecated`.

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
MMatten: Could/should we merge this method into `createFunctionCallExecution`?

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
```diff
@@ -121,19 +121,44 @@ public final PreparedStatement createStatementWithBoundFixtureSymbols(
...
-    public StatementExecution createStatementExecution(PreparedStatement statement, boolean clearParameters) {
-        return new StatementExecution(statement, clearParameters);
+    public final StatementExecution createCallExecution(String commandText) throws SQLException {
+        return createStatementExecution(getConnection().prepareCall(commandText));
     }
 
     @Override
-    public StatementExecution createFunctionStatementExecution(PreparedStatement statement, boolean clearParameters) {
-        return new StatementExecution(statement, clearParameters);
+    public final StatementExecution createFunctionCallExecution(String commandText) throws SQLException {
```
MMatten: What would the value of `commandText` typically look like?

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
```diff
@@ -121,19 +121,44 @@ public final PreparedStatement createStatementWithBoundFixtureSymbols(
...
+
+    @Override
+    public final StatementExecution createStatementExecution(String commandText) throws SQLException {
+        return createStatementExecution(getConnection().prepareStatement(commandText));
+    }
+
     @Override
-    public StatementExecution createStatementExecution(PreparedStatement statement, boolean clearParameters) {
-        return new StatementExecution(statement, clearParameters);
+    public final StatementExecution createCallExecution(String commandText) throws SQLException {
```
MMatten: What would the value of `commandText` typically look like?

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
javornikolov: > Would it be desirable to have this method accepting the SQL statement text (like createDdlStatementExecution)?

Actually I did that - we [have it](https://github.com/dbfit/dbfit/blob/topic/refactor-statement-execution-creation/dbfit-java/core/src/main/java/dbfit/api/DBEnvironment.java#L68).

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
javornikolov: That's currently only used by `DbStoredProcedureCall` which basically builds the sp call - either `{? = call f(?, ?, ...)}` or {call p(?, ?, ...)}`.

dbfit-java/core/src/main/java/dbfit/api/DBEnvironment.java
```diff
@@ -58,21 +56,26 @@ PreparedStatement buildInsertPreparedStatement(String tableName, DbParameterAcce
      */
 
     /**
-     * Create a {@link PreparedStatement} object and binds fixture symbols to
+     * Create a {@link StatementExecution} and binds fixture symbols to
```
MMatten: Should we apply this `{@link StatementExecution}` to the Javadoc for the other methods?

MMatten:
In general I do like the changes as they simplify using the mechanisms. Can we make the jump in the next release or do you think deprecating the old methods initially is better?

javornikolov:
> In general I do like the changes as they simplify using the mechanisms. Can we make the jump in the next release or do you think deprecating the old methods initially is better?

To be honest, I'd rather jump and go with the simpler version right away. If that gets someone in trouble - I think we can easily support either the migration or we can make a release with the deprecated methods back.

javornikolov:
What do you think about `createCallExecution(_)` + `createFunctionCallExecution(_)` vs `createCallExecution(_, boolean isFunction)`?

javornikolov:
> What do you think about createCallExecution(_) + createFunctionCallExecution(_) vs createCallExecution(_, boolean isFunction)?

I just committed a version with `isFunction` flag to see how it looks like.

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
MMatten: Do we still need this method?

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
```diff
@@ -126,14 +127,26 @@ public DdlStatementExecution createDdlStatementExecution(String ddl)
...
+    }
+
     @Override
-    public StatementExecution createStatementExecution(PreparedStatement statement, boolean clearParameters) {
-        return new StatementExecution(statement, clearParameters);
+    public final StatementExecution createStatementExecution(String commandText) throws SQLException {
+        return createStatementExecution(getConnection().prepareStatement(commandText));
+    }
+
+    protected StatementExecution createCallExecution(PreparedStatement statement, boolean isFunction)
```
MMatten: Do we want/need to retain methods that take a `PreparedStatement` or can we replace all of them with methods that take SQL text?

dbfit-java/core/src/main/java/dbfit/api/DbStoredProcedureCall.java
```diff
@@ -52,14 +51,8 @@ void bindParametersTo(StatementExecution cs) throws SQLException {
     }
 
     public StatementExecution toStatementExecution() throws SQLException {
-        String sql = toSqlString();
-        PreparedStatement ps = environment.getConnection().prepareCall(sql);
-        StatementExecution cs;
```
MMatten: May be we should rename `cs` now. Presumably `cs` was once a `CallableStatement`.

dbfit-java/core/src/main/java/dbfit/api/DbTable.java
```diff
@@ -30,8 +30,8 @@ public DbTable(DBEnvironment dbEnvironment, String tableName)
 
     public StatementExecution buildPreparedStatement(
```
MMatten: Should we rename this to `buildStatementExecution`?

dbfit-java/core/src/main/java/dbfit/fixture/StatementExecution.java
```diff
@@ -4,27 +4,20 @@
 
 public class StatementExecution implements AutoCloseable {
     protected PreparedStatement statement;
-    protected int returnValueInd = -1;
 
-    public StatementExecution(PreparedStatement statement, boolean clearParameters) {
+    public StatementExecution(PreparedStatement statement) {
```
MMatten: Should/could we allow this to take the SQL text directly instead of a `PreparedStatement`?

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
javornikolov: > Do we want/need to retain methods that take a PreparedStatement or can we replace all of them with methods that take SQL text?

`createStatementExecution(PreparedStatement)` is currently used by 2 other methods. If e.g. we want to customize something - similarly like we did for Informix: it's good to minimize the number of places we need to touch. (And it's hard to get from-query-text-only version of the bound symbols method).

MMatten:
> What do you think about createCallExecution(_) + createFunctionCallExecution(_) vs createCallExecution(_, boolean isFunction)?

Looking at it now I think I prefer `createCallExecution(_, boolean isFunction)` only because we reduce the number of methods. But other than that I think they're both good. What's your preference?

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
javornikolov: `createCallExecution(PreparedStatement)` - it's used only in one place now and we can leave the `from-text` text version only but that makes it asymmetrical compared to `createStatementExecution`. I thought it would be cleaner if we follow the same pattern for both.

dbfit-java/core/src/main/java/dbfit/api/DbTable.java
javornikolov: > Should we rename this to buildStatementExecution?

Good question about the naming... It's perhaps used in other places, and it's still kind of prepared statement - not so bad name.

Another wondering was about this extra `Execution` - it makes names a bit longer. Would it be cleaner to drop it from method names or it's not worth.

dbfit-java/core/src/main/java/dbfit/fixture/StatementExecution.java
javornikolov: > Should/could we allow this to take the SQL text directly instead of a PreparedStatement?

Well, `StaementExecution` doesn't know how to create statements from SQL; can be solved by passing some extra factory parameter but that won't make things simpler.

javornikolov:
> Looking at it now I think I prefer createCallExecution(_, boolean isFunction) only because we reduce the number of methods. But other than that I think they're both good. What's your preference?

I also think both are viable versions. The `boolean` flag seems a bit better fit for the current moment - less methods, simpler `DbStoredProcedureCall` (at the cost of a bit complication in `InformixEnvironment`).

dbfit-java/core/src/main/java/dbfit/fixture/StatementExecution.java
MMatten: Right. I was thinking that `StatementExecution` could be a direct replacement for `PreparedStatement` and `CallableStatement` but it would need a `DBEnvironment`'s connection to do so.

dbfit-java/core/src/main/java/dbfit/api/DbStoredProcedureCall.java
javornikolov: > May be we should rename cs now. Presumably cs was once a CallableStatement.

It's coming from `createCallExecution` and is indeed backed by `CallableStatement` - so callable statement is so bad association... Yet maybe just `statement` would be a bit more clear (short abbreviations are sometimes obscure).

dbfit-java/core/src/main/java/dbfit/api/DbTable.java
MMatten: I'm not even sure about the `StatementExecution` class name to be honest! The `Execution` bit doesn't quite sound right.

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
MMatten: Ok. Makes sense.

dbfit-java/core/src/main/java/dbfit/api/DbTable.java
javornikolov: Could be `DbStatement` ... to reuse the name of the class we just dropped.

dbfit-java/core/src/main/java/dbfit/api/DbStoredProcedureCall.java
MMatten: Or even something to indicate we're calling a stored code unit.

dbfit-java/core/src/main/java/dbfit/api/DbTable.java
MMatten: Yes. I was thinking the same thing.

MMatten:
Is it also worth reviewing the name of the `DdlStatementExecution` class? This isn't limited to executing DDL but any SQL statement where we don't need to bind params.

javornikolov:
> Is it also worth reviewing the name of the `DdlStatementExecution` class? This isn't limited to executing DDL but any SQL statement where we don't need to bind params.

It was intended for DDL statements which seems to have some peculiarities: e.g. fails to work with `PreparedStatement` for Oracle; looks like it needs commit for `TeraData`.
True - it might be useful for some other things too.

javornikolov:
`StatementExecution` renamed to `DbStatement`.

javornikolov:
@MMatten, what do you think about the current state of this PR - are we good enough to finalize and merge it?

MMatten:
Nothing obvious to comment upon. 

I'm finding it's a lot of stuff for me to try to review and keep in my head as I go. Could it be worth chopping up into smaller PRs?

javornikolov:
Yes, you're right - we got too many changes here. I'll try to slice it into a few smaller ones.

javornikolov:
OK, let's take care about sliced off PRs #475, #476, #477 for a start.

MMatten:
Ok. I can take another look when the others have been merged and this one rebased, etc.

javornikolov:
One more separate PR spinned off: #478

javornikolov:
I rebased this PR - perhaps it's simple enough for review now. Still some wondering about:
 * `api` vs `util` package for `DbStatement`
 * making it interface and invent a name for the implementing class

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
```diff
@@ -117,23 +116,41 @@ public final PreparedStatement createStatementWithBoundFixtureSymbols(
...
             throws SQLException {
-        return new DdlStatementExecution(getConnection().createStatement(), ddl);
+        return new DdlStatement(getConnection().createStatement(), ddl);
+    }
+
+    /**
+     * Create a {@link DbStatement} for the given prepared statement.
+     * This is the method to override if custom implementation needs to be returned
+     */
+    protected DbStatement createPreparedStatement(PreparedStatement statement) {
```
MMatten: This method name still seems strange as it takes a `PreparedStatement` as a parameter.

MMatten: Would `createPreparedDbStatement` be better?

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
```diff
@@ -117,23 +116,41 @@ public final PreparedStatement createStatementWithBoundFixtureSymbols(
...
+        return new DbStatement(statement);
     }
 
     @Override
-    public StatementExecution createStatementExecution(PreparedStatement statement) {
-        return new StatementExecution(statement);
+    public final DbStatement createPreparedStatement(String commandText) throws SQLException {
```
MMatten: ...and so how about `createPreparatedDbStatement` here too?

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
```diff
@@ -117,23 +116,41 @@ public final PreparedStatement createStatementWithBoundFixtureSymbols(
...
+        return createPreparedStatement(getConnection().prepareStatement(commandText));
+    }
+
+    /**
+     * Create a {@link DbStatement} for the given prepared statement.
+     * This is the method to override if custom implementation needs to be returned
+     */
+    protected DbStatement createCallStatement(PreparedStatement statement, boolean isFunction)
```
MMatten: ...and then `createCallDbStatement` (or `createCallableDbStatement`)?

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
```diff
@@ -117,23 +116,41 @@ public final PreparedStatement createStatementWithBoundFixtureSymbols(
...
+            throws SQLException {
+        return new DbStatement(statement);
     }
 
     @Override
-    public StatementExecution createFunctionStatementExecution(PreparedStatement statement) {
-        return new StatementExecution(statement);
+    public final DbStatement createCallStatement(String commandText, boolean isFunction) throws SQLException {
```
MMatten: ...and then `createCallDbStatement` (or `createCallableDbStatement`)?

dbfit-java/core/src/main/java/dbfit/api/DBEnvironment.java
```diff
@@ -58,27 +53,31 @@ PreparedStatement buildInsertPreparedStatement(String tableName, DbParameterAcce
...
     /**
-     * Create a procedure statement execution object for the given command text.
+     * Create a prepared statement for the given command text
      */
-    StatementExecution createStatementExecution(PreparedStatement statement);
+    DbStatement createPreparedStatement(String commandText) throws SQLException;
 
     /**
-     * Create a function statement execution object for the given command text.
+     * Create a stored procedure call {@link DbStatement} for the given command text.
```
MMatten: Should this say "Create a stored code unit call" to include functions?

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
```diff
@@ -104,11 +103,11 @@ protected String parseCommandText(String commandText) {
...
     }
 
-    public final PreparedStatement createStatementWithBoundFixtureSymbols(
+    @Override
+    public final DbStatement createStatementWithBoundFixtureSymbols(
             TestHost testHost, String commandText) throws SQLException {
         String command = Options.isBindSymbols() ? parseCommandText(commandText) : commandText;
-        PreparedStatement cs = getConnection().prepareStatement(
-                command);
+        PreparedStatement cs = getConnection().prepareStatement(command);
```
MMatten: You did explain why `cs` was OK for a name, but I can't quite recall it now. Don't we end up creating a wrapped `PreparedStatement` from this and not a wrapped `CallableStatement`?

dbfit-java/core/src/main/java/dbfit/api/DbObject.java
```diff
@@ -1,12 +1,11 @@
...
 
-import dbfit.fixture.StatementExecution;
 import dbfit.util.DbParameterAccessor;
 import dbfit.util.Direction;
 
 import java.sql.SQLException;
 
 public interface DbObject {
-    public StatementExecution buildPreparedStatement(DbParameterAccessor accessors[]) throws SQLException ;
+    public DbStatement buildPreparedStatement(DbParameterAccessor accessors[]) throws SQLException;
```
MMatten: Should we align names like `buildPreparedStatement` to `buildPreparedDbStatement`?

dbfit-java/core/src/main/java/dbfit/api/DbTable.java
```diff
@@ -28,10 +27,10 @@ public DbTable(DBEnvironment dbEnvironment, String tableName)
         }
     }
 
-    public StatementExecution buildPreparedStatement(
+    public DbStatement buildPreparedStatement(
             DbParameterAccessor[] accessors) throws SQLException {
-        StatementExecution statement = dbEnvironment.createStatementExecution(dbEnvironment
-                .buildInsertPreparedStatement(tableOrViewName, accessors));
+        DbStatement statement =
+            dbEnvironment.buildInsertStatement(tableOrViewName, accessors);
```
MMatten: Should we align names like`buildInsertStatement` to `buildInsertDbStatement`?

dbfit-java/core/src/main/java/dbfit/api/DdlStatement.java
```diff
@@ -1,13 +1,13 @@
-package dbfit.util;
+package dbfit.api;
 
 import java.sql.Statement;
 import java.sql.SQLException;
 
-public class DdlStatementExecution implements AutoCloseable {
+public class DdlStatement implements AutoCloseable {
```
MMatten: Or `DdlDbStatement`?

dbfit-java/core/src/main/java/dbfit/fixture/Clean.java
```diff
@@ -81,10 +81,10 @@ public void deleteTableByIds(String tableName, String whereClause, String column
         }
 
         private void executeQuery(SQLQuery q) throws SQLException {
-            try (PreparedStatement st =
+            try (DbStatement st =
```
MMatten: Worth expanding `st`?

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
javornikolov: `cs` perhaps stands for `callable statement`. We may rename that to `statement` here (and not only here). It's indeed more an exception than a rule to get a callable statement here.

dbfit-java/core/src/main/java/dbfit/fixture/Query.java
```diff
@@ -40,7 +40,7 @@ public MatchableDataTable getDataTable() throws SQLException {
         }
 
         Log.log("Query: '%s'", queryOrSymbol);
-        try (PreparedStatement st =
+        try (DbStatement st =
```
MMatten: Worth expanding `st`?

dbfit-java/core/src/main/java/dbfit/util/DbParameterAccessor.java
```diff
@@ -15,7 +15,7 @@
     private String userDefinedTypeName;
     private Class<?> javaType;
     private int position; // zero-based index of parameter in procedure (-1 for ret value) or column in table
-    protected StatementExecution cs;
+    protected DbStatement cs;
```
MMatten: Should we change `cs` as it might suggest that the `DbStatement` wraps a `CallableStatement`?

dbfit-java/core/src/main/java/dbfit/fixture/Clean.java
javornikolov: Here it doesn't matter so much (it's 4 lines method) but I guess it makes sense to review all abbreviations (`cs`, `st`) and change them to `statement.

MMatten:
> * api vs util package for DbStatement

Do you think `api` as it's a class that adapter developers might want to extend?

dbfit-java/core/src/main/java/dbfit/api/DdlStatement.java
javornikolov: I don't know, `DdlStatement` I think is not so bad. It doesn't even extend `DbStatement`.

MMatten:
> * making it interface and invent a name for the implementing class

What does making `DbStatement` into an interface do for us, or other developers? Just freedom to create their own implementations instead of extending ours?

javornikolov:
> Do you think `api` as it's a class that adapter developers might want to extend?

That's related to purpose of `api` which is maybe a bit unclear. I'm thinking about various options like:

  1. Package containing all what's needed to implement an environment (a custom and/or a built-in one)
  2. All needed to implement dbfit fixtures (custom and/or built-in ones)
  3. Both of above
  5. Lower-level of services to be used for integration tests - e.g. without DbFit UI
  6. Interfaces and abstract classes without a real implementation. (It's just an API, no implementation - hence the `api` name of the package)

Variation of above (e.g. for 1-3) is `main/major components` instead of `all needed`.

Currently it's hard to imagine achieving 1-3 without some `util` classes like `DbParameterAccessors`.

javornikolov:
> What does making DbStatement into an interface do for us, or other developers? Just freedom to create their own implementations instead of extending ours?

Here is [an article](http://www.javaworld.com/article/2073649/core-java/why-extends-is-evil.html) discussing classes vs interfaces. But as you're questioning - maybe right now there is no obvious/expected use case where splitting out an interface would be beneficial.

dbfit-java/core/src/main/java/dbfit/api/DBEnvironment.java
javornikolov: > Should this say "Create a stored code unit call" to include functions?

I think to say it's a `stored subroutine`.

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
javornikolov: > ...and then createCallDbStatement (or createCallableDbStatement)?

I had some hesitation about extra particles (`Execution`, `Db`) since they're making names longer. Maybe `Db` is not so huge trouble.

javornikolov:
I think to move `DbStatement` to `util` package, rename abbreviations to `statement` and maybe add the `Db` to the method names for better alignment with the class name.

dbfit-java/core/src/main/java/dbfit/api/DBEnvironment.java
javornikolov: > > Should this say "Create a stored code unit call" to include functions?

> I think to say it as a stored subroutine.

But in general, at least as per my understanding, `stored procedure` is already kind of aggregating term. I would expect functions to be part of this unless it's clear from the context that we're talking about technology specific `procedures`. (E.g. there are two types of stored procedures in Oracle: functions and procedures).

javornikolov:
I pushed another experimental change:
* interfaces `RunnableStatement`  and `QueryStatement` (1st - just execute, 2nd one returns `ResultSet`)
* `PreparedDbStatement` (new name for `DbStatement`) implements both
* `DdlStatement` implements `QueryStatement`
* Fixtures mostly depend on interfaces only (at least no direct reference to class names). Exception is `Update` because it builds its statement itself.

MMatten:
> That's related to purpose of api which is maybe a bit unclear.

I must admit that I still don't quite understand the purpose of the DbFit API. It's not like a set of methods that can be invoked by other services to perform some actions. Was it established by @gojko as part of the GreenPepper support?

javornikolov:
I'm not sure about the history. 'api' package is already present in the 1st commit of the current repository.

An approach with `api` package is used e.g. in [junit-lambda](https://github.com/junit-team/junit-lambda). There are samples how it's used: [junit-samples](https://github.com/junit-team/junit5-samples).

In our case it's maybe close to the point I mentioned - API (or we may say "database access API") to be used in implementing fixtures. Currently we have `Fit Fixtures', but in theory we may add/migrate to SLIM or other testing framework.

dbfit-java/core/src/main/java/dbfit/api/QueryStatement.java
```diff
@@ -0,0 +1,10 @@
+package dbfit.api;
+
+import java.sql.ResultSet;
+import java.sql.SQLException;
+
+public interface QueryStatement extends AutoCloseable {
```
MMatten: I guess we need some Javadoc text for any new interfaces and methods, etc.

dbfit-java/core/src/main/java/dbfit/api/RunnableStatement.java
```diff
@@ -0,0 +1,9 @@
+package dbfit.api;
+
+import java.sql.SQLException;
+
+public interface RunnableStatement extends AutoCloseable {
```
MMatten: As all statements are runnable, is this a good name?

MMatten:
May be it will be hard to predict what the API needs to look like until we have SLIM or other front-ends like Robot or GreenPepper.

I'm a bit out of my depth here I think. I'll defer to your experience and judgement.

dbfit-java/core/src/main/java/dbfit/api/RunnableStatement.java
javornikolov: > As all statements are runnable, is this a good name?

Good question, the name came from the `run` method... Basically it's statement where we don't have result returned (unlike `QueryStatement` where we have). I wondered also about `ExecutionStatement` as alternative name but it suffers from the same problem (all statements are executable).

javornikolov:
> May be it will be hard to predict what the API needs to look like until we have SLIM or other front-ends like Robot or GreenPepper.

Maybe leads to another point of `api` possible purpose - more or less stable and generic API which is agnostic to the specifics of a particular testing framework.

MMatten:
> Maybe leads to another point of api possible purpose - more or less stable and generic API which is agnostic to the specifics of a particular testing framework.

That sounds like a good case to me. But with only one example of a consumer it's hard to say what the generalisations should be.

MMatten:
> Good question, the name came from the run method... Basically it's statement where we don't have result returned (unlike QueryStatement where we have)

I don't think there are any other good alternatives. I can't think of any.

javornikolov:
>> Good question, the name came from the run method... Basically it's statement where we don't have result returned (unlike QueryStatement where we have)

> I don't think there are any other good alternatives. I can't think of any.

In that respect maybe `StatementExecution` had its merits. Other ideas so far:
* `DbCommand` + `DbCommandWithResult` (or `DbQueryCommand`, `DbQuery`)
* `DbExecution` + `DbExecutionWithResult` or `[Db]QueryExecution`
Maybe use `execute()` instead of `run()` - to keep similar names to JDBC; or at least have some alignment between the 2 versions (e.g. `run` + `runQuery`)

MMatten:
> In that respect maybe StatementExecution had its merits.

This name always suggested to me it was an instance of an execution of a statement, not an executable statement.

MMatten:
Would it make any sense or us to have something like `DbStatement` and then `DbQueryStatement` that extends it? Would there be anything in `DbStatement` that wouldn't be relevant to a `DbQueryStatement`?

MMatten:
> Maybe use execute() instead of run() - to keep similar names to JDBC; or at least have some alignment between the 2 versions (e.g. run + runQuery)

JDBC has `execute()` and `executeQuery()` if we want to stay close to those.

javornikolov:
> Would it make any sense or us to have something like interfaces DbStatement and then DbQueryStatement that extends it? Would there be anything in DbStatement that wouldn't be relevant to a DbQueryStatement?

I pushed a version with `DbCommand` and `DbQuery`. While working on it same question emerged (`DbQueryCommand` extending `DbCommand`). We have cases (maybe most or all cases) when we just need the `runQuery` version returning `ResultSet` but we don't need the `void run()` one.

That's why I decided the Query interface not to extend from Command for now. But the question is still valid. If we go on with extending, and in order to avoid the burden of unused `run` method - either abstract class may be created (not a perfect complication); or use interface *default method* (needs Java 8).

javornikolov:
More about the `DbCommand`, `DbQuery` experiments so far and some other general comments:
  - It avoids collisions with `Statement`
  - Fixtures are usually based on an `api` layer (`DbObjects', `DbTable`, etc.). Direct reference to `DBEnvironment` is minimal. (I moved some of the `Update` logic to `DbTable`).
  - I also created `PreparedDbCommand` interface (fixtures are not using it directly, so far I placed it in `util` package)
  - I did some refactoring of `Insert`, `Update` fixtures to see how things are going with these Command interfaces and package separation.

  - In `util` package we have a mix of things supporting *db environment* implementation; and other things supporting *fixtures*, including `FIT` specific stuff. Perhaps it's good to separate those (int separate packages - below or next to `util`, I'm not sure yet). Some common, generic stuff may remain in a commonly used package (could remain in `util` I guess) - but as a rule should be avoided when possible. In any case - mixed packages seem to increase complicate dependencies/increase coupling.
  - The above point maybe applies to `api` too.
  - In particular, `DbParameterAccessor` - maybe should rather be moved either to `api` (or rather an interface extracted out of it)
  - `AbstractDbEnvironment` - I have a feeling that it doesn't quite belong to `api`. (Maybe rather to a package supporting environment implementation. Could be `environment` package in `core` project - this way we can remove public visibility of some supporting classes to keep them local).

javornikolov:
> JDBC has execute() and executeQuery() if we want to stay close to those.

After a few renaming exercises I tend to like 'run` and `runQuery` in such scenarios: since it's easier to distinguish from JDBC and mass-update it...

MMatten:
> AbstractDbEnvironment - I have a feeling that it doesn't quite belong to api.

I agree. It's not directly usable. May be `environment` is better.

MMatten:
> Perhaps it's good to separate those (int separate packages - below or next to util, I'm not sure yet).
...
> In particular, DbParameterAccessor - maybe should rather be moved either to api (or rather an interface extracted out of it)

I think I'm struggling to get to grips with this stuff still. Perhaps some kind of statement, even a picture (an extension to the SUT part of http://www.fitnesse.org/FitNesse.UserGuide.WritingAcceptanceTests.TestSystems) of desired logical layers would help.

javornikolov:
I encountered some general guidelines and discussions about API design:
http://theamiableapi.com/2012/01/16/java-api-design-checklist/
http://stackoverflow.com/questions/2954372/difference-between-spi-and-api

In our case:
* `DBEnvironment`, `AbstractDbEnvironment` and other supporting stuff (which tends to go to `environment` package I guess) is kind of SPI (Service Provider Interface) for implementing database-specific environments.
* `api`, `DbObject`, `DbTable`, (and DBEnvironment again) - provides interface for running database commands. (Used to implement DbFit fixtures in our case, in `fixture` package).
* Environment implementations - we have these in separate projects, so far in `environment` package. Might make sense to also move down to sub-packages (e.g. `dbfit.environment.oracle`).

dbfit-java/core/src/main/java/dbfit/api/DBEnvironment.java
```diff
@@ -35,10 +32,15 @@
...
+     * Create an insert command that will be used to populate new rows in a table
      */
-    PreparedStatement buildInsertPreparedStatement(String tableName, DbParameterAccessor[] accessors)
+    PreparedDbCommand buildInsertStatement(String tableName, DbParameterAccessor[] accessors)
+            throws SQLException;
+
+    /**
+     * Build a call command for the given stored procedure name and parameters
+     */
+    DbCommand buildStoredProcedureCall(String name, DbParameterAccessor[] accessors)
```
javornikolov: I think we may stick to following convention:
* `buildXXX` - building command out of object name, accessors, etc.
* `createXXX` - creating command out of plain text statement.

javornikolov:
@MMatten, what do you think about the current state of this PR? (I created issue #483 to keep track ideas/questions so that we can reduce the further scope creep here).

MMatten:
@javornikolov, sorry for dropping out of the discussion recently. I won't be able to do much on this for another week. I need to read those articles firstly. Do you think you have a clear idea of where we should take the API or repackaging?

javornikolov:
@MMatten, the ideas are still no crystal clear and I think we'll need to refine both API and package further in several iterations (separate PRs). Some general points so far:
 - Simplify interfaces between modules, don't expose excessive details. (In that respect - e.g. prefer interfaces over concrete classes).
 - Avoid mixing components with different purpose in one place (same class, or package, etc.). E.g.:
   - Environment implementations have different needs than fixture implementations.
   - Or: overloading a single module with too much responsibility. (I removed building `update` statement from `Update` fixture, it didn't seem to be the right place).
 - Clear naming;

MMatten:
That's a good discussion on stackoverflow. The other info from the API design site is a bit overwhelming right now. Perhaps we can pick out a few key guidelines. 

If we were to map what we have now (in the master branch) to the concepts in the Service Provider Framework then perhaps we have: -
* Service Interface: This is the unclear but I think. We're not sure what our API is.
* Provider Registration API: I don't think we have this as we use annotations/reflection to discover the available implementations. 
* Service Access API: DbEnvironmentFactory.
* Service Provider API: DBEnvironment, extending AbstractDbEnvironment, DdlStatementExecution, StatementExecution - the stuff Service Providers (adapter/environment developers) need to provide.

Is this how you understood it?

May be it would help for us to clarify what our services are first before doing more work on providing the APIs for those services. A diagram might be good too. 





Sent from my iPad

javornikolov:
* Service Interface: this is `DBEnvironment`
* Provider Registration API: this is `DbEnvironmentFactory.registerEnv` but as you noted - it's used only internally.

* SPI
> ... the stuff Service Providers (adapter/environment developers) need to provide

Not exactly... It's the stuff which is provided **to** them to help them to support their implementation of `DBEnvironment`. (What the need to provide is an implementation of the interface `DBEnvironment`).

The rest is as I understand it, yes.

MMatten:
> Not exactly... It's the stuff which is provided *to* them to help them implement `DBEnvironment`. (What they need to provide is an implementation of the interface `DBEnvironment`).

Right. That what I'd really meant to say.

MMatten:
> Perhaps we can pick out a few key guidelines.

Which I guess your previous points: -
* Simplify interfaces between modules, don't expose excessive details. (In that respect - e.g. prefer interfaces over concrete classes).
* Avoid mixing components with different purpose in one place (same class, or package, etc.). E.g.:
  * Environment implementations have different needs than fixture implementations.
  * Or: overloading a single module with too much responsibility.
* Clear naming;

MMatten:
> Service Interface: this is `DBEnvironment`

I see. So the existing `DBEnvironment` is both consumer/API and provider/SPI.

javornikolov:
> So the existing `DBEnvironment` is both consumer/API and provider/SPI.

Kind-of - provider is implementing the interface, the consumer is using it. There is a note about the API/SPI shared stuff: [Do not place API and SPI into the same package](http://theamiableapi.com/java-cl-section-1-1/#cl.item.1.1.10).

MMatten:
> Kind-of - provider is implementing the interface, the consumer is using it.

Again, that's what I really meant. I'm not having a good day with words today...!

dbfit-java/core/src/main/java/dbfit/util/DbParameterAccessors.java
```diff
@@ -12,15 +12,34 @@
 public class DbParameterAccessors implements Iterable<DbParameterAccessor> {
```
MMatten: Could it be worth renaming this class to something like `DbAttributeAccessor` as it's not only used for stored procedure parameters?

javornikolov: Good question... I'm not sure we can cay that a parameter is kind of `Attribute`. In `OracleEnvironment` we have some nested classes named like `DbParameterOrColumnInfo `, `DbDictionaryParamsOrColumnsIterator`, etc.

Technically `DbParameterAccessor` is used for accessing parameters of database queries (sp calls, insert, update).

I'm adding a note about that in #483.

MMatten:
Should we rename `InformixFunctionStatement` to `InformixFunctionPreparedDbStatement`?

javornikolov:
> Should we rename `InformixFunctionStatement` to `InformixFunctionPreparedDbStatement`?

Could rather be something like `InformixFunctionCallCommand`? Maybe extending from `PreparedDbStatement` is not so good idea, what we need is to comply with the interface (`PreparedDbCommand`).

I'm also wondering about the name of `PreparedDbStatement` - should we keep it or rename (to `DefaultPreparedDbCommand` or something better).

dbfit-java/core/src/main/java/dbfit/api/DbQuery.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.sql.ResultSet;
+import java.sql.SQLException;
+
+/**
+ * An object representing an executable database operation.
+ * Executing the operation returns a ResultSet result.
+ *
+ */
+public interface DbQuery extends AutoCloseable {
```
MMatten: As this is only used by `PreparedDbCommand` is it worth having this as an independent interface?

MMatten:
For my own understanding I'm starting to create a basic diagram of what we have.

image

dbfit-java/core/src/main/java/dbfit/api/DbQuery.java
javornikolov: > As this is only used by PreparedDbCommand is it worth having this as an independent interface?

We have it in `Inspect`, `Query`, `QueryStats`, `StoreQuery`.

MMatten: What are the benefits of having separate interfaces vs a more complex API?

MMatten:
> I'm also wondering about the name of `PreparedDbStatement `- should we keep it or rename (to `DefaultPreparedDbCommand` or something better).

I think it would be good to have the name aligned with the originating interface. Could we even use the same name (e.g. `util.PreparedDbCommand`) or is that a recipe for confusion?

dbfit-java/core/src/main/java/dbfit/api/DbQuery.java
javornikolov: > What are the benefits of having separate interfaces vs a more complex API?

It's a general principle, some description and examples here: https://springframework.guru/principles-of-object-oriented-design/interface-segregation-principle. But it's a good question - are we on the right balance with these interfaces?

javornikolov:
> Could we even use the same name (e.g. `util.PreparedDbCommand`) or is that a recipe for confusion?

In theory it's possible but it may be a bit confusing indeed. I think if we see what's distinguishing for the implementing class - we can come up with some class-name prefix (`Generic`, `Standard`, ...). E.g. - now we have two implementations - the common one and one for Informix functions: how can we describe what's specific in both cases?

dbfit-java/core/src/main/java/dbfit/api/PreparedDbCommand.java
```diff
@@ -0,0 +1,18 @@
+package dbfit.api;
+
+import java.sql.SQLException;
+
+/**
+ * Represents a prepared database statement with ability to bind input parameters and
+ * to register and access output ones including return values.
+ */
+public interface PreparedDbCommand extends DbCommand, DbQuery {
```
MMatten: By including `Prepared` in the name do we tie ourselves to the JDBC `PreparedStatement` or reveal clues about implementation details? 

Would something like `ParameterisedDbCommand` be an alternative?

javornikolov: We don't tie to JDBC details - just borrowing the same convention. I think it's worth considering `ParameterisedDbCommand`. It raises a question about `getGeneratedKey` - as it doesn't seem to be related to the parameterization.

dbfit-java/core/src/main/java/dbfit/api/DbQuery.java
MMatten: Ok. So it looks like a good idea to keep `DbCommand` and `DbQuery` separate (segregated) and we couldn't even use the `execute()` method for statements that return a `ResultSet` and those that don't because of the different method signatures.

dbfit-java/core/src/main/java/dbfit/api/DbObject.java
```diff
@@ -1,12 +1,11 @@
...
 
-import dbfit.fixture.StatementExecution;
 import dbfit.util.DbParameterAccessor;
 import dbfit.util.Direction;
 
 import java.sql.SQLException;
 
 public interface DbObject {
-    public StatementExecution buildPreparedStatement(DbParameterAccessor accessors[]) throws SQLException ;
+    public DbCommand buildPreparedStatement(DbParameterAccessor accessors[]) throws SQLException;
```
MMatten: Should we align this with the returned type? E.g. `buildDbCommand`.

javornikolov: > Should we align this with the returned type? E.g. `buildDbCommand`.

Not necessarily aligned with the type name but should be aligned with what it does and `buildDbCommand` sounds to me a better fit for that purpose too. So I renamed it.

dbfit-java/core/src/main/java/dbfit/api/PreparedDbCommand.java
MMatten: Would it make any sense to try to segregate `PreparedStatement` further? E.g. slice off `getGeneratedKey()` as we're only using it for `INSERT`s.

I've not really thought it through yet. Do you think it would make the statement creation much more complex?

dbfit-java/informix/src/main/java/dbfit/environment/InformixFunctionStatement.java
```diff
@@ -2,19 +2,19 @@
 
 import java.sql.*;
 
-import dbfit.fixture.StatementExecution;
+import dbfit.util.PreparedDbStatement;
 
-public class InformixFunctionStatementExecution extends StatementExecution {
+public class InformixFunctionStatement extends PreparedDbStatement {
```
MMatten: Do we have any standard for naming subtypes? Should we name this `InformixFunctionDbStatement`?

MMatten:
> Could rather be something like InformixFunctionCallCommand? Maybe extending from PreparedDbStatement is not so good idea, what we need is to comply with the interface (PreparedDbCommand).

Why do you think it's not a good idea? I guess if we're overriding lots of stuff then a new class could be better.

MMatten:
> I'm also wondering about the name of `PreparedDbStatement` - should we keep it or rename (to `DefaultPreparedDbCommand` or something better).

I'm still not sure about having the `Prepared` part of the name. What about `DefaultDbCommand` or `BasicDbCommand`?

MMatten:
I'll have another pass over the changes tomorrow too.

dbfit-java/informix/src/main/java/dbfit/environment/InformixFunctionStatement.java
MMatten: >  Should we name this `InformixFunctionDbStatement` or `InformixFunctionPreparedDbStatement`? (A little long winded perhaps).

Ah, we've already covered this one off.

javornikolov:
>> Could rather be something like InformixFunctionCallCommand? Maybe extending from PreparedDbStatement is not so good idea, what we need is to comply with the interface (PreparedDbCommand).

> Why do you think it's not a good idea? I guess if we're overriding lots of stuff then a new class could be better.

I mean - we can still have a new class but use composition instead of inheritance. The fact that we're using inheritance is implementation detail - should it influence our naming then.

dbfit-java/core/src/main/java/dbfit/api/PreparedDbCommand.java
javornikolov: > Would it make any sense to try to segregate PreparedStatement further? E.g. slice off getGeneratedKey() as we're only using it for INSERTs.
> 
> I've not really thought it through yet. Do you think it would make the statement creation much more complex?

One issue is when we create a statement and we want to ask it for both `execute()` and `getGeneratedKey()`.

MMatten: > One issue is when we create a statement and we want to ask it for both `execute()` and `getGeneratedKey()`.

We'd have to know what type of statement we're building and we'd need more combinations (extensions) of interfaces in addition to `PreparedDbCommand` (which would be suitable for `INSERT`s, but not `DELETE`/`UPDATE`/`SELECT`/`CALL` for us at the moment).

MMatten: So, perhaps we'd only have to know if we're creating an `INSERT` statement or not, but more specifically whether we, potentially, need to get at any generated keys and this would only be where we're building an `INSERT` statement for the `Insert` fixture I think.

MMatten:
> I mean - we can still have a new class but use composition instead of inheritance. The fact that we're using inheritance is implementation detail - should it influence our naming then.

By composition do you mean, for example, wrapping `PreparedDbStatement` (and implementing `PreparedDbCommand`) instead or extending it?

I don't know about the conventions for naming. Whatever the implementation it's still going to be a specialised type of `PreparedDbCommand` for Informix function execution.

javornikolov:
> By composition do you mean, for example, wrapping PreparedDbStatement (and implementing PreparedDbCommand) instead or extending it?

Yes, exactly. The name should indicate what it is. I'm in the middle of getting rid that `Prepared` prefix now to see how would it look like. I'll push it soon.

dbfit-java/core/src/main/java/dbfit/api/PreparedDbCommand.java
javornikolov: > So, perhaps we'd only have to know if we're creating an INSERT statement or not, but more specifically whether we, potentially, need to get at any generated keys and this would only be where we're building an INSERT statement for the Insert fixture I think.

Right now `getGeneratedKeys` is accessed via `DbParameterAccessor` - the database statement member in it (which  in turn is being set via `bindTo`).

In general - I think we should be aiming to easy extension and maintenance in terms of:
  1. New environment implementations (and let's say - some of these may be proprietary). Things should be simple but also stable for environment authors. (Adding more methods to `DBEnvironment` interface is in general a pain for environment implementations: it may break them /it's forcing them to be implemented/). So in such situations "fat" interfaces might be a sane choice (e.g. JDBC API is like that).
  2. New test systems (in addition to Fit), maybe new fixtures, etc. The API used by the fixtures should be thin and simple (yet, adding more methods here is not a pain /removing ones would be/). By providing a bridging layer between `DBEnvironment` and the fixtures - we can avoid exposing excessive details to fixtures.

These are just general comments. It might be worthwhile to see if we can come up with something better about auto-generated keys access.

javornikolov:
> I'm still not sure about having the Prepared part of the name. What about `DefaultDbCommand` or `BasicDbCommand`?

Renamed (alongside related classes and methods).

javornikolov:
@MMatten, I'm considering to merge this PR as we already got quite many changes here. It's not all polished and perfect but I believe we can handle further improvements in separate PRs (or even next releases).

MMatten:
@javornikolov, cool. There were a few loose ends in here but it's good to keep moving forwards. Should we do a sweep up of those (some are buried in old diff comments - e.g. the fat API points) an carry them on to a new issue?

We can get another release out too perhaps (4.0?).

javornikolov:
@MMatten , thanks for your update. We have #483 linked here and I believe we can dig into the old comments when needed (no rush right now to open new issues for all uncovered topics and carry on related comments there).

Yes, version we can bump up to `4.0` but I'm thinking to do it in a separate PR.

So I'll proceed with merging this PR...

PR #466, opened 2016-01-02
Fix and refactor Informix DateTime types lists

### LINKED ISSUE #464: Simplify InformixEnvironment timestampTypes list initialization
Currently timestampTypes list is quite big. Consider simplifying the initialization - e.g. generate the list with some helper function.

### LINKED ISSUE #465: Bad timestamp types list for Informix
In `InformixEnvironment.timestampTypes` list we have `HOUR TO FRACTION(5)` twice and we're missing `MINUTE TO FRACTION(5)`

javornikolov:
resolves #464 resolves #465


javornikolov:
@MMatten, do DateTime lists look OK to you now? I'm not quite certain what about the `Time` vs `Timestamp` distinction (as I read Informix is quite pedantic and won't like using Timestamp instead of Time).

MMatten:
I can take a look later on but I think we must be missing some acceptance tests too.

javornikolov:
> I think we must be missing some acceptance tests too.

Exactly...

MMatten:
Does this change make it all more cryptic? If we have a unit tests for this then the expected output would be the original format list won't it?

MMatten:
>  I'm not quite certain what about the Time vs Timestamp distinction

I could really have just put them all into a single list. It was just to make the number of `DATETIME <high> TO <low>` list entries more manageable.

javornikolov:
> Does this change make it all more cryptic?

To some extent yes, indeed... Yet - seems easier to check whether the list is correct.

> If we have a unit tests for this then the expected output would be the original format list won't it?

I started with a unit test and that's how I discovered that the original list was wrong! (The fix is in ae5733776d310444d32de9192d0552b6f6f60c83). Looks like it may be good to get that unit test (or reverse version of it) committed - I'll get check it in.

javornikolov:
I extracted hard-coded pre-populated lists of date/time dates into a helper class:
 - allowed adding a unit test (without exposing too much `InformixEnvironment` internals)
 - avoids bloating `InformixEnvironment`
 - neater formatting with one item on line is not a problem

MMatten:
I've revised the acceptance test page for the `DATETIME` types. I can push it to your branch if that's ok?

javornikolov:
> I've revised the acceptance test page for the DATETIME types. I can push it to your branch if that's ok?

Yes, sure - you can go ahead and push your changes here.

MMatten:
Pushed it now. I've reformatted it to make it easier to read (and spot issues).

javornikolov:
Thanks! Looks good to me :+1:. @MMatten, unless you have other ideas for changes here - I'm planning to squash last 3 commits and merge the PR.

MMatten:
Nothing to add. All looking good to me.

PR #467, opened 2016-01-02
Set MSSQL output parameters as INPUT_OUTPUT

javornikolov:
* Change `SqlServerEnvironment.getAllProcedureParameters` to handle `output` parameter as `INPUT_OUTPUT`
* Tidy up a quirk in DbStoredProcedure (it used to handle `OUTPUT` parameters as `INPUT_OUTPUT` for SQL Server) - shouldn't be needed any more

### DIFF
--- a/FitNesseRoot/DbFit/AcceptanceTests/DotNetTests/SqlServerTests/FlowMode/SimpleStoredProc/content.txt
+++ b/FitNesseRoot/DbFit/AcceptanceTests/DotNetTests/SqlServerTests/FlowMode/SimpleStoredProc/content.txt
@@ -21,3 +21,21 @@ Caputure the implicit return value
 !|Execute Procedure|CalcLength_P|
 |name|str length?|?|
 |mika|4|0|
+
+SQL Server OUTPUT parameters behave like INOUT
+{{{
+CREATE PROCEDURE Increment_P
+	@counter INT OUTPUT
+AS
+BEGIN
+	SET @counter = ISNULL(@counter, 1) + 1
+END;
+}}}
+
+!|Execute Procedure|Increment_P|
+|counter?|
+|2|
+
+!|Execute Procedure|Increment_P|
+|counter|counter?|
+|7|8|
--- a/dbfit-java/sqlserver/src/integration-test/resources/acceptancescripts-SqlServer.sql
+++ b/dbfit-java/sqlserver/src/integration-test/resources/acceptancescripts-SqlServer.sql
@@ -82,6 +82,22 @@ SET ANSI_NULLS ON
 GO
 SET QUOTED_IDENTIFIER ON
 GO
+IF NOT EXISTS (SELECT * FROM sys.objects WHERE object_id = OBJECT_ID(N'[dbo].[Increment_P]') AND type in (N'P', N'PC'))
+BEGIN
+EXEC dbo.sp_executesql @statement = N'CREATE PROCEDURE [dbo].[Increment_P]
+	@counter INT OUTPUT
+AS
+BEGIN
+	SET @counter = ISNULL(@counter, 1) + 1
+END;
+'
+END
+GO
+
+SET ANSI_NULLS ON
+GO
+SET QUOTED_IDENTIFIER ON
+GO
 IF NOT EXISTS (SELECT * FROM sys.objects WHERE object_id = OBJECT_ID(N'[dbo].[ReturnUserTable_F]') AND type in (N'FN', N'IF', N'TF', N'FS', N'FT'))
 BEGIN
 execute dbo.sp_executesql @statement = N'-- =============================================
--- a/dbfit-java/core/src/main/java/dbfit/api/DbStoredProcedure.java
+++ b/dbfit-java/core/src/main/java/dbfit/api/DbStoredProcedure.java
@@ -35,12 +35,6 @@ public class DbStoredProcedure implements DbObject {
             parameter = parameter.clone();
             parameter.setDirection(expectedDirection);
         }
-        // sql server quirk. if output parameter is used in an input column,
-        // then the param should be cloned and remapped to IN/OUT
-        if (expectedDirection!=Direction.OUTPUT && parameter.hasDirection(Direction.OUTPUT)) {
-            parameter = parameter.clone();
-            parameter.setDirection(Direction.INPUT);
-        }
         return parameter;
     }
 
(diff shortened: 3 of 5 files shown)


MMatten:
What's the rationale for this change?

javornikolov:
Basically the goal is to simplify `DbStoredProcedure` and have the correct metadata in parameter accessors. Also to avoid possible side effects on non non-mssqlserver databases.

I got into that while looking into `DbStoredProcedure`, there is some duplicated code there. And that extra code seems to be due to the not quite precise direction metadata in SQL Server parameter accessors.

javornikolov:
Hi @MMatten, may you check that too - that it doesn't break acceptance tests? Looks like MSSQL Server can always get an input for a procedure parameter (the `output` parameters are effectively `inout`) so I think this change sets the direction correctly.

MMatten:
Ok, will test this too now.

MMatten:
All existing acceptance tests have passed. But do we have acceptance tests to cover this change?

javornikolov:
Yes, we have new one here inside the pull request - added to `SimpleStoredProc` test. But it needs  a new stored procedure to be created in the database (`Increment_P`) - have you ran that test too?

MMatten:
Ah. I hadn't created the new procedure.

Something strange (and worrying) is happening. The `dbfit-java:sqlserver:integrationTest` gradle task runs OK (passes). When I start FitNesse and run the suite then the `SimpleStoredProc` tests fails with the exceptions for the missing stored proc!

javornikolov:
> Ah. I hadn't created the new procedure.
> 
> Something strange (and worrying) is happening. The dbfit-java:sqlserver:integrationTest gradle task runs OK (passes). When I start FitNesse and run the suite then the SimpleStoredProc tests fails with the exceptions for the missing stored proc!

Hmm, I wonder if we have that test properly included. (I think the FitNesse links don't report errors if link is pointing to a bad path...)

MMatten:
In the HTML files in `/var/dbfit/tmp' on the test VM I can see that the test has failed though.

MMatten:
> Ah. I hadn't created the new procedure.

May be we can use `dbdeploy` for the SQL Server script some time.

dbfit-java/sqlserver/src/integration-test/resources/acceptancescripts-SqlServer.sql
```diff
@@ -82,6 +82,22 @@ SET ANSI_NULLS ON
 GO
 SET QUOTED_IDENTIFIER ON
 GO
+IF NOT EXISTS (SELECT * FROM sys.objects WHERE object_id = OBJECT_ID(N'[dbo].[Increment_P]') AND type in (N'P', N'PC'))
+BEGIN
+EXEC dbo.sp_executesql @statement = N'CREATE PROCEDURE [dbo].[Increment_P]
+	@counter INT OUTPUT
+AS
+BEGIN
+	set @counter = isnull(@counter, 1) + 1
```
MMatten: We have no defined standard style for SQL code but elsewhere we have the keyword `SET` and built-in functions in UPPER case.

FitNesseRoot/DbFit/AcceptanceTests/DotNetTests/SqlServerTests/FlowMode/SimpleStoredProc/content.txt
```diff
@@ -21,3 +21,22 @@ Caputure the implicit return value
...
+{{{
+CREATE PROCEDURE Increment_P
+	@counter INT OUTPUT
+AS
+BEGIN
+	set @counter = isnull(@counter, 1) + 1
+END;
+}}}
+
+
```
MMatten: 2 blank lines intentional?

javornikolov:
Good spots - I'll fix these formatting bits and I'll force push.

javornikolov:
> Good spots - I'll fix these formatting bits and I'll force push.

Done.

dbfit-java/sqlserver/src/integration-test/resources/acceptancescripts-SqlServer.sql
```diff
@@ -82,6 +82,22 @@ SET ANSI_NULLS ON
 GO
 SET QUOTED_IDENTIFIER ON
 GO
+IF NOT EXISTS (SELECT * FROM sys.objects WHERE object_id = OBJECT_ID(N'[dbo].[Increment_P]') AND type in (N'P', N'PC'))
+BEGIN
+EXEC dbo.sp_executesql @statement = N'CREATE PROCEDURE [dbo].[Increment_P]
+	@counter INT OUTPUT
+AS
+BEGIN
+	SET @counter = isnull(@counter, 1) + 1
```
MMatten: What about `isnull`? We have `OBJECT_ID()` and `CAST()` elsewhere.

javornikolov: > What about isnull? We have OBJECT_ID() and CAST() elsewhere.

I don't know - for the user-defined functions and procedures we're using `CamelCase`. Maybe SQL Server preserves the case - as the object has been created, does it? (If yes - that may be the general convention). Anyway - in the meanwhile I'll change that to uppercase too.

javornikolov: > Anyway - in the meanwhile I'll change that to uppercase too.

Done (`isnull` -> `ISNULL`)

MMatten: With SQL Server case sensitivity depends upon the collation settings and I think they can be specified at different levels too (there was a PR as part of 3.2.0 related to this). Our test DBs are defined to be case insensitive anyway.

MMatten:
I've run all of the acceptance tests via the FitNesse UI and they've all passed.
That gradle behaviour is a concern. I take it FitNesse would report tests with only exceptions, not test failures, as errors and fail the task...?

javornikolov:
> That gradle behaviour is a concern. I take it FitNesse would report tests with only exceptions, not test failures, as errors and fail the task...?

I would expect exceptions to be treated as errors and fail the test. Maybe we should open an issue for that problem in order to examine it closer.

javornikolov:
> I've run all of the acceptance tests via the FitNesse UI and they've all passed.

Great, thanks a lot for testing and and for your feedback! I'm merging...

PR #468, opened 2016-01-02
DbAutoGeneratedKeyAccessor: use OUTPUT instead of RETURN_VALUE direction

javornikolov:
In general `RETURN_VALUE` direction is used for functions - it's used to check the stored procedure type (function or procedure) in some places. At first glimpse now - `OUTPUT` seems a better choice if it works.

 - [ ] TODO: run acceptance tests and check if this works OK with `OUTPUT` direction


javornikolov:
Accidentally this got in the way of some hardening checks for #470. I'm going to cherry-pick this there.

PR #470, opened 2016-01-03
Reorg/fix DbParameterAccessors bindParameters

javornikolov:
Something seems suspicious when registering RETURN_VALUE parameter:

``` java
        for (DbParameterAccessor ac : accessors) {
            int realindex = accessorNames.indexOf(ac.getName());
            ac.bindTo(statement, realindex + 1); // bindTo called 1st time
            if (ac.hasDirection(RETURN_VALUE)) {
                ac.bindTo(statement, Math.abs(ac.getPosition())); // 2nd time
            }
        }
```

* If metadata is correct: position is expected to be `-1` for return values. In such a case it would be sorted 1st, realIndex == 0 => both bindTo calls will be invoked with `ind = 1`.
* If return value position is not correct (not `-1`) - seems a broken case.

I'm considering to remove that special case with the 2nd bindTo - the 1st one should be enough.

### DIFF
--- a/dbfit-java/core/src/main/java/dbfit/util/DbParameterAccessor.java
+++ b/dbfit-java/core/src/main/java/dbfit/util/DbParameterAccessor.java
@@ -52,6 +52,10 @@ public class DbParameterAccessor {
             throw new IllegalArgumentException("Position of return value should be -1");
         }
 
+        if ((name == null) || (direction != RETURN_VALUE && name.isEmpty())) {
+            throw new IllegalArgumentException("Missing column or procedure parameter name");
+        }
+
         this.name = name;
         this.direction = direction;
         this.sqlType = sqlType;
--- a/dbfit-java/core/src/main/java/dbfit/util/DbParameterAccessors.java
+++ b/dbfit-java/core/src/main/java/dbfit/util/DbParameterAccessors.java
@@ -25,9 +25,6 @@ public class DbParameterAccessors {
         for (DbParameterAccessor ac : accessors) {
             int realindex = accessorNames.indexOf(ac.getName());
             ac.bindTo(statement, realindex + 1); // jdbc params are 1-based
-            if (ac.hasDirection(Direction.RETURN_VALUE)) {
-                ac.bindTo(statement, Math.abs(ac.getPosition()));
-            }
         }
     }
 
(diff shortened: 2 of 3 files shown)


javornikolov:
@MMatten, do you know what's the logic of `paramid` in `sysproccolumns` for Informix? Basically the expectation is to have zero-based index for all non-return value parameters.

MMatten:
Looking at what's in `sysproccolumns` for `FuncWithOutParams`, the return value has paramid 0, `instring` is 1 and `outstring` is 2.

javornikolov:
What about procedures (i.e. when no return value is present)?

javornikolov:
> the return value has paramid 0, `instring` is 1 and `outstring` is 2.

In anyways - it doesn't seem quite right to use that value directly in accessors. I modified the implementation to generate the position (zero-based) using the query sorting order (like we do in the adapters for the other databases).

MMatten:
It might be good to implement `DbParameterAccessorsMapBuilder` in all of the other adaptors too for consistency. I think we only have it in SQL Server and now Informix.

MMatten:
For procedures the first param has `paramid` of 0. E.g. for `CalcLength` then `InString` is 0 and `StrLength` is 1.

javornikolov:
> It might be good to implement `DbParameterAccessorsMapBuilder` in all of the other adaptors too for consistency. I think we only have it in SQL Server and now Informix.

Yes, we can think of that (it's a separate topic though).

javornikolov:
> For procedures the first param has paramid of 0. E.g. for CalcLength then InString is 0 and StrLength is 1.

OK - so it should have been right before for procedures.

@MMatten, may you test if the version in this PR is working fine?

MMatten:
>> It might be good to implement  DbParameterAccessorsMapBuilder  in all of the other adaptors too for consistency. I think we only have it in SQL Server and now Informix.

> Yes, we can think of that (it's a separate topic though).

Draft/placeholder created in #471.

MMatten:
> @MMatten, may you test if the version in this PR is working fine?

Yes. I can do this later today.

MMatten:
>> @MMatten, may you test if the version in this PR is working fine?

> Yes. I can do this later today.

All acceptance tests for all adapters have passed.

javornikolov:
Nice :-) Thanks for checking that, @MMatten!
I'm merging the PR...

PR #475, opened 2016-01-05
Move returnInd to InformixFunctionStatementExecution

javornikolov:
Refactoring: move returnInd field from StatementExecution to InformixFunctionStatementExecution as it's currently specific for Informix only


MMatten:
Easy to review now and looking good to me :+1:

PR #477, opened 2016-01-05
Remove clearParameters from StatementExecution c-tor

javornikolov:
Remove `clearParameters` parameter from `StatementExecution` and related factory methods. Leave the behavior of `clearParameters = *false*`.

In the existing code all cases with `clearParameters=true` the passed statement is clean anyway (it doesn't  ave any parameters) - so explicit `clearParameters` seems superfluous.

If in future a need for explicit `cleanParameters` emerges: it could be rather implemented as a `StatementExecution` method.


MMatten:
Easy to review now and looking good to me :+1:

Issue #215, opened 2014-01-20
Harden coding conventions check


javornikolov:
Being more explicit on desired coding conventions may avoid some unnecessary rework and discussions around pull requests.
- [ ] Add a note about coding conventions on contributor's doc
- [ ] Harden checkstyle checks
- [ ] Cleanup existing code (to ease working by example and to reduce "broken windows" syndrome)

PR #480, opened 2016-01-09
Enforce Java whitespace style

MMatten:
* Remove horizontal tabs from Java code and replace with 4 spaces.
* Remove trailing spaces from Java code.
* Configure checkstyle to detect horizontal tab characters and trailing spaces.


javornikolov:
Wonderful! (Partly addresses #215).
I took a look at some other checkstyle options and I think we can include a few more whitespspace checks here (trailing whitespace, newline at end of file) - I'm pushing the related commits.

dbfit-java/teradata/.gitignore
```diff
@@ -0,0 +1 @@
+/bin/
```
javornikolov: Why do we need that .gitignore?

MMatten: That's my bad (wild card used in `git add`).

MMatten:
> I took a look at some other checkstyle options and I think we can include a few more whitespspace checks here (trailing whitespace, newline at end of file) 

Cool. I was just getting the ball rolling really with this piece but we might as well get it all into this PR.

dbfit-java/teradata/.gitignore
javornikolov: > That's my bad (wild card used in git add).

I think there's a way to configure your own ignore settings (in `.git/info/exclude or something)

javornikolov:
> Cool. I was just getting the ball rolling really with this piece but we might as well get it all into this PR.

I'm ready with cleaning up trailing whitespaces. It got into several commits, we'll need to squash (maybe just 1 commit for the fixes and 1 for the checkstyle.xml). Besides that, I think we're ready with that PR.

MMatten:
Great. I'll squash them now.

In another PR perhaps we can deal with braces. I can see that we can also specify that braces are always required around blocks, but I'm not sure if we can specify the required position, etc.

javornikolov:
> In another PR perhaps we can deal with braces. I can see that we can also specify that braces are always required around blocks, but I'm not sure if we can specify the required position, etc.

Yes, I saw some check-style options: about space before/after braces and brackets... Hopefully there should be something for placement too. We should pay attention what happens with `try-with-resource` cases as it's relatively new syntax.

MMatten:
Squashed.

javornikolov:
Thanks! May we mention in the comments about the `NewlineAtEndOfFile` (we have 3 new rules in checkstyle config). I would also suggest moving the formatting fixes commit before `checkstyle.xml` (not that it matters much but this way gradle build won't fail if we build from any of the commits).

MMatten:
Good points. I'll update now.

javornikolov:
Thank you @MMatten! It looks good now :+1: Merging...

PR #496, opened 2016-05-17
more MSSQL notes

jeremysimmons:
add note about SQL Browser service requirement when using sql express
also changed previous instruction about enabling tcp/ip protocol. I'm not sure if there's a better way to indicate you could have multiple engine instances installed, and you need to enable for the one you're working with.

### DIFF
--- a/website/docs/database-specific-information.md
+++ b/website/docs/database-specific-information.md
@@ -110,9 +110,13 @@ There are two ways around this:
 
 ### Database Engine configuration 
 
-* Make sure the engine is configured to allow TCP/IP connections. Launch SQL Server Configuration. SQL Server Network Configuration > SQL Server Express > Right click TCP/IP and choose Enable. 
+* Make sure the engine is configured to allow TCP/IP connections. Launch SQL Server Configuration. SQL Server Network Configuration > SQL Server > Right click TCP/IP and choose Enable. 
 * Make sure you restart your engine after you enable TCP/IP connections.
 * Make sure you allow mixed-mode authentication so you can use Logins created to use SQL Authentication.
+
+#### SQL Express Specific 
+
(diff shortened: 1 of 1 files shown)


MMatten:
Is it always port 1434?

Can you double check this requirement? I've not had to do this myself before.

jeremysimmons:
yes. absolutely sure @MMatten 
Documentation here: https://technet.microsoft.com/en-us/library/ms181087(v=sql.105).aspx

What are the details of the !Connect fixture you're using?

When I use this !Connect fixture, with SQL Browser Service in Stopped, I get an exception.
!|Connect|jdbc:sqlserver://localhost\sqlexpress;user=FitNesseUser;password=FitNesseUser;databaseName=FitNesseTestDB|

com.microsoft.sqlserver.jdbc.SQLServerException: The connection to the host localhost, named instance sqlexpress failed. Error: "java.net.SocketTimeoutException: Receive timed out". Verify the server and instance names and check that no firewall is blocking UDP traffic to port 1434.  For SQL Server 2005 or later, verify that the SQL Server Browser Service is running on the host.

MMatten:
What happens if you connect using the port number instead of the instance name with the browser service disabled?

E.g. 

`!|Connect|jdbc:sqlserver://localhost:1433;user=FitNesseUser;password=FitNesseUser;databaseName=FitNesseTestDB|`

jeremysimmons:
@MMatten sql express binds to a random port by default. The SQL Server Client requires the SQL Browser service which operates on a single well-known port to provide resolution of Instance-Name to Port. Think of it like DNS... you trade a DomainName for IP address...

You CAN force SQL Express to use Port 1433, but it's not the default.
https://msdn.microsoft.com/en-us/library/ms177440.aspx

javornikolov:
Thank you @jeremysimmons!

PR #499, opened 2016-06-19
Bump up FitNesse to 20160618 (RowSetFixture updated accordingly)

javornikolov:
Align DbFit with current FitNesse version.

Makes DbFit incompatible with older FitNesse versions.

### DIFF
--- a/dbfit-java/core/src/main/java/dbfit/fixture/RowSetFixture.java
+++ b/dbfit-java/core/src/main/java/dbfit/fixture/RowSetFixture.java
@@ -64,7 +64,7 @@ public abstract class RowSetFixture extends ColumnFixture {
     }
 
     @Override
-    protected void bind(Parse heads) {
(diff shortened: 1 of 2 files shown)


dbfit-java/core/src/main/java/dbfit/fixture/RowSetFixture.java
```diff
@@ -64,7 +64,7 @@ private int findColumn(String name) throws Exception {
     }
 
     @Override
-    protected void bind(Parse heads) {
+    protected void bindColumnHeadersToMethodsAndFields(Parse heads) {
```
MMatten: What triggered this change?

javornikolov: The method in the parent `ColumnFixture` got renamed unclebob/fitnesse/issues/938

javornikolov:
Merging that since as per the conversation at FitNesse side - seems the Fit API change is going to stay.

PR #513, opened 2016-10-20
Test VM: updated base box (centos/7), cookbooks and packages

MMatten:
Update Vagrant VB base box, cookbook versions and packages versions where appropriate:

* box based on CentOS 7.2. changed VG/LV names.
* VBox disk expansion now handled via vagrant-newdisk plugin
* mysql 5.7
* postgresql 9.6
* git 2.10.1
* eliminate resource cloning in mysql and  postgresql cookbooks (CHEF-3694)
* swap, postgresql - switch back to upstream cookbooks
* update .gitignore list (for the extra VBox disk, vagrant temp files, informix installer)
* disable selinux and firewalls
* fix vbox shared mountpoint options (nfs v4 requires udp_off)

TODO:
- [x] Check if we can switch `swap` cookbook to upstream (instead of dbfit fork)
- [x] Try to deal with the CHEF-3694 in our mysql cookbook
- [x] Try to deal with the CHEF-3694 in our postgres cookbook
- [x] Tidy up the commits in this PR
- [x] Tidy up the PR title and description  (**IN PROGRESS** [Yavor])


test_vm/vendor/cookbooks/dbfit_test/recipes/default.rb
```diff
@@ -1,6 +1,8 @@
 include_recipe "dbfit_test::diskspace"
-include_recipe "dbfit_test::firewalld"
+include_recipe 'firewalld::disable'
```
javornikolov: Hmm. Is that firewall change necessary?
If we want to disable firewall - maybe we should also remove dbfit `firewalld.rb` recipe. Or alternatively we can tweak `firewalld.rb`, open more ports, etc. instead of disabling it.

MMatten: Good spot.

MMatten:
Actually regarding the `firewalld` service - it looks like port forwarding of 8085 to the host still isn't quite working. After turning off the `iptables` service I was able to connect to FitNesse on the guest from the host.

I could turn off this service as well (with https://supermarket.chef.io/cookbooks/iptables). What do you think?

javornikolov:
OK, let's disable the firewalls. In general it depends if VM network interfaces are exposed to external world (depends how the VM box is configured).

javornikolov:
So... I managed to arrange a cloud box on https://zrh.cloudsigma.com to try the VM setup for a few days :-) I'm getting the following error on `vagrant up` from this branch:

```
==> default: Running 'pre-boot' VM customizations...
A customization command failed:

["storageattach", :id, "--storagectl", "IDE Controller", "--port", 1, "--device", 0, "--type", "hdd", "--medium", "./disk1.vdi"]

The following error was experienced:

#<Vagrant::Errors::VBoxManageError: There was an error while executing `VBoxManage`, a CLI used by Vagrant
for controlling VirtualBox. The command and stderr is shown below.

Command: ["storageattach", "acdec18e-b21f-47e2-9349-89ccfa2500ff", "--storagectl", "IDE Controller", "--port", "1", "--device", "0", "--type", "hdd", "--medium", "./disk1.vdi"]

Stderr: VBoxManage: error: Could not find a controller named 'IDE Controller'
>

Please fix this customization and try again.
```

MMatten:
That's strange. Let me double check I'm up to date with VB.

MMatten:
> In general it depends if VM network interfaces are exposed to external world (depends how the VM box is configured).

Isn't that ultimately governed by the host's firewall?

javornikolov:
> That's strange. Let me double check I'm up to date with VB.

That indeed depends on VBox version - controller name is `IDE` for 5.1.8 and `IDE Controller` For VBox `5.0`. Similar topic is discussed here: https://www.virtualbox.org/ticket/5994.

MMatten:
> That indeed depends on VBox version - controller name is IDE for 5.1.8 and IDE Controller For VBox 5.0. Similar topic is discussed here: https://www.virtualbox.org/ticket/5994.

That's nasty. But this ticket is ancient. 

Have you tried with both recent VB versions?

javornikolov:
> Have you tried with both recent VB versions?

Versions were 5.1.8 and 5.0.28.

MMatten:
I'm actually on actually 5.0.25 for VB. I might have installed from a public repo earlier instead of installing the download that I'd got.

So we could perhaps add our own disk controller to avoid name variations in VB versions. Can you try the latest commit? I think it might need a better name though.

javornikolov:
I switched to VB 5.0.10 - there are some bugs when running newer versions of VB inside KVM virtual machine...

The last commit is a great idea :+1: Looks like it would be compatible with the different VB versions. I just tested it - looks working fine with `VBox 5.0.10`.

MMatten:
> The last commit is a great idea :+1: Looks like it would be compatible with the different VB versions. I just tested it - looks working fine with VBox 5.0.10.

This is weird. Today, the additional disk now appears as `/dev/sda` and the `diskpace` recipe fails.

May be the device names are unpredictable with different controller types.

javornikolov:
Hmm, I also had some issue with the device name but I thought it was due to running `vagrant provision` while the vagrant box is already up (versus creating the box from scratch via `vagrant up`).

javornikolov:
Port and device numbers maybe have some influence: `'--port', 1, '--device', 0`

Another interesting question is - which is the controller of the disk which already exists in the vm box?

MMatten:
> Port and device numbers maybe have some influence: '--port', 1, '--device', 0

I'll try playing with these.

MMatten:
> Another interesting question is - which is the controller of the disk which already exists in the vm box?

The base box comes up with "IDE Controller" on my VB version. Do you mean port/device numbers? There's nothing visible in the VB UI.

MMatten:
But I can try these options: http://www.virtualbox.org/manual/ch08.html#vboxmanage-storageattach

javornikolov:
Once upon a time IDE controllers used to lead to `hdX*` device names instead of `sdX`. However that changed at some point...

I guess the way to deal with that is to exploit `uuid` options in some of the `storageattach` options.

javornikolov:
Here is an idea: if the new controller is of the same type as the existing one (`IDE`) - maybe it's more likely to come on 2nd place and get named as `/sdb*`.

javornikolov:
> Here is an idea: if the new controller is of the same type as the existing one (IDE) - maybe it's more likely to come on 2nd place and get named as /sdb*.

That unfortunately fails (there can be only one IDE controller):
```
Stderr: VBoxManage: error: Too many storage controllers of this type
```

MMatten:
> Here is an idea: if the new controller is of the same type as the existing one (IDE) - maybe it's more likely to come on 2nd place and get named as /sdb

I tried that too.

MMatten:
I'll try adding a SCSI controller.

javornikolov:
`SCSI` controllers naming is also `/dev/sd*`.
I'm already in a way of writing a Vagrant plugin :-) I'm just exploring https://github.com/sprotheroe/vagrant-disksize as baseline. The idea is to create something similar with following syntax:
``` ruby
config.newdisk.size 4000 # size in MB
config.newdisk.path file_to_disk
```

Within plugin there is an access to the VM internals.

MMatten:
Well, the additional disk seems to now consistently appear at `/dev/sda`. May be this is actually reliable after all.

What was the exact device name issue you observed?

MMatten:
OK. I've noticed that `vagrant destroy` doesn't always delete the `disk`.vdi` file. I wonder if that what happened on my env yesterday and if the `diskspace` recipe was actually executed against the built in disk device at `/dev/sdb`. But I even doubt this as if I run `pvcreate /dev/sdb` I get the error 

`Device /dev/sdb not found (or ignored by filtering).`.

It would be nice to get this working but if it's too much hassle we can just abandon it as the new base box comes with a 40GB disk and it's mostly unused initially.

javornikolov:
Same thing with leftover `disk.vdi` - while experimenting I often abort the provisioning (not giving it opportunity to clean up the disk).

To me usually the new disk appears as `/dev/sda`. It would be good the original disk to always be `/dev/sda`, otherwise things are too brittle...

It doesn't look very hard to do the Vagrant plugin at first glimpse.

MMatten:
Is it worth slicing off the new Vagrant plugin work onto a different PR to get something usable up and running first?

Any comments on the MySQL version specification changes? I could try getting 5.7 working although it's just a dev milestone release.

javornikolov:
The `vagrant-newdisk` plugin is ready: https://github.com/dbfit/vagrant-newdisk
Version 0.0.2 is published at [rubygems](https://rubygems.org/gems/vagrant-newdisk/). Currently `vagrant plugin install vagrant-newdisk` is unable to find it for some reason, but the gem can be downloaded and manually installed.

javornikolov:
> Any comments on the MySQL version specification changes? I could try getting 5.7 working although it's just a dev milestone release.

I have some insights about that. Basically - I think the way is to declare `depends 'mysql', '< 8.0'` in `metadata.rb`. There is some mess in mysql recipe since version `8.0`. After that - it's not needed to manually install the devel package, version `5.7` pulls the latest 5.7 (no need to specify the specific version like 5.7.16-1.el7). Maybe `yum-community-package` won't be needed either but I haven't thoroughly checked that yet.

javornikolov:
`vagrant plugin install vagrant-newdisk` works fine now from Internet. I just pushed a few commits here to add the disk using the new plugin.

CONTRIBUTING.md
```diff
@@ -75,10 +75,12 @@ The VM doesn't include:
    [vagrant-librarian-chef](https://github.com/jimmycuadra/vagrant-librarian-chef),
    [vagrant-omnibus](https://github.com/schisamo/vagrant-omnibus),
    [vagrant-vbguest](https://github.com/dotless-de/vagrant-vbguest):
+   [vagrant-newdisk](https://github.com/dbfit/vagrant-newdisk):
 
         vagrant plugin install vagrant-librarian-chef
         vagrant plugin install vagrant-omnibus
         vagrant plugin install vagrant-vbguest
+        vagrant plugin install vagrant-newdisk
```
MMatten: Ok. This makes things easy for now at least. Do some controller types have a maximum number of attached disks? E.g. could be end up with a base box in the future with multiple disks already attached and then this could fail?

javornikolov: Yes, I think there are certain limitations. I think max number of `ports` per IDE controller is 2. I'm not sure about devices (I guess it should be OK to add at least 10). We'll think about it if we reach the limit.
Also in case of more disks it's likely the new device not to be `/dev/sdb`. We should check that when we change the vagrant base box.

MMatten:
I'll rebuild my VM and run all the tests for every DB type. If they all pass then are we happy to merge?

javornikolov:
Some other things TODO:
- [ ] Check if we can switch `swap` cookbook to upstream (instead of dbfit fork)
- [ ] Tidy up the commits in this PR
- [ ] Try to deal with the CHEF-3694 in our mysql cookbook [I'm investigating that right now]

MMatten:
This is annoying - the Netezza emulator is only available natively for Windows / MacOS via VMWare. I'll try getting it running on KVM (as on here: http://tesla.duckdns.org/netezza-emulator-kvm/) so I can complete the testing.

test_vm/vendor/cookbooks/dbfit_test/recipes/mysql.rb
```diff
@@ -23,14 +23,18 @@
...
     connection mysql_connection_info
+    username username
     password password
     action :create
   end
 
   %w{localhost 127.0.0.1}.each do |hostname|
-    mysql_database_user username do
+    mysql_database_user "grans on dbfit.* to #{username}@#{hostname}" do
+      connection mysql_connection_info
```
MMatten: what's this do? Is `grans` a typo?

javornikolov: Yes, it's a typo (thanks for spotting it!). I just pushed a fix-up. It grants privileges for the test users to the test database. There are some warnings when having several `mysql_database_user <name>` with the same name (as it leads to `resource cloning`).

javornikolov:
@MMatten, I'm going to tidy up the commit history in the PR and then I'll force-push.

javornikolov:
One thing I noticed is that the [vagrant-librarian-chef](https://github.com/jimmycuadra/vagrant-librarian-chef) plugin is not maintained any more. I gave a quick experiment with the berkshelf alternative but for some reason it results in pulling different (lower) versions for some packages. So far `vagrant-librarian-chef` works fine for us so let's stick to it until we have a good enough replacement.

MMatten:
Nothing else to add. I've tested with everything apart from Teradata and Netezza and as those are nit gusted in our VM I'm guessing no issues there.

javornikolov:
Great :-) Thank you @MMatten!
I'm merging...

I have some progress with the librarian-chef -> bershelf migration but it will be a subject of separate PR.

PR #515, opened 2016-10-20
Exclude test VM optional installation DB types from default integrationBuild task

MMatten:
Exclude test VM optional installation DB types from default integrationBuild task

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,0 +1,1 @@
+excludeIntegrationTests=db2,informix,netezza,oracle,sqlserver,teradata


javornikolov:
Why do we want that for the default build? We're already excluding certain int builds from `fastbuild`.

One issue with adding `gradle.properties` under version control is that it may have user specific customization (in my case I have) - related to publishing to maven or other stuff. This is why that file is in `.gitignore` list.

MMatten:
My thoughts on this are to make the default `integrationBuild` config match the default VM provisioning. I.e. no Oracle/DB2/Informix/Teradata/Netezza.

In order to use the default `integrationBuild` task with the default VM build I am forced to do something - either exclude integration test types or install additional software (DBs).

Can't users base their customisations upon a new standard `gradle.properties`?

I think the main downside of this proposal is that we could miss out integration tests when regression testing changes but seeing as devs/users currently need a big machine to concurrently host all of the DB types I think they'll always have to disable certain builds anyway and this risk will always exist.

javornikolov:
I agree that it's good to do something to avoid confusion/frustration when running the build without the extra databases installed on the VM.

> Can't users base their customisations upon a new standard gradle.properties?

Good question, I'm not sure.

Another option I can think of is adding a few notes to the contribution notes + `gradle.properties.samlpe`.

MMatten:
> Another option I can think of is adding a few notes to the contribution notes + gradle.properties.samlpe

True. But them I'm back to having to do something manual.

May be the extra DB provisioning could update the `gradle.properties` to enable the tests upon successful installation but that's starting to get complex and a little obscure.

javornikolov:
BTW, `build.properties` might be located also in user home directory: https://docs.gradle.org/current/userguide/build_environment.html

It's possible one to want running tests with certain configuration when doing that inside the VM and with other settings when build is started outside the test VM.

MMatten:
It looks like a properties file in $HOME overrides the file in the project directory then.

So can we leave the one it the source repo as per the PR and override via the $HOME one (but that then sounds like a global override for any project).

javornikolov:
> but that then sounds like a global override for any project

Yes, it's not nice (even if unlikely anyone to run other projects *in the dbfit test VM*).

Personally I don't have the VM right now and my settings are: `excludeIntegrationTests=netezza,sqlserver,teradata,oracle,db2,postgres`. I'm also excluding postgres. Quite likely someone working on windows would like to keep `sqlserver` only. Others may want the built-in embedded databases only... At the same time - the ideal goal is to run tests on all the databases.


Imagine that we commit this `gradle.properties` now with the list of tests to exclude. I want to customize it - what can I do. I don't want to affect any other project; touching any version-controlled file is not an option since it's my personal settings.

javornikolov:
If we want to align list of tests with the test-vm: we can have a helper script which does `vagrant up` + optional parameters to init the list of excluded tests list (it can automatically check which optional software package are present for the test VM - and omit if them from the exclusion list).

Or an `init` help script can prepare the right settings (appending them to `gradle.properties` or something). It can accept certain parameters - explicit list of tests to accept or to skip; maybe something like profile name (full, simple, testvm-default, ...), we may have an option to setup and provision the test vm.

Or we can have a helper script inside the VM which to pass custom parameters for the build, relevant inside the vm: `./gradlew -DexcludeIntegrationTests="..." ...`

MMatten:
> I want to customize it - what can I do? I don't want to affect any other project; touching any version-controlled file is not an option since it's my personal settings.

You can. 

Note that I'm not proposing that we remove this file from `.gitignore`. I just wanted to get the `integrationBuild` task usable out of the box.

MMatten:
So, picking up this thread again.

> Quite likely someone working on windows would like to keep sqlserver only. 

They can run the specific test suite. e.g. `./gradlew :dbfit-java:sqlserver:integrationTest`.

> We're already excluding certain int builds from fastbuild.
...
> Others may want the built-in embedded databases only... At the same time - the ideal goal is to run tests on all the databases.

The `fastbuild` is the Java-only (unit tests) build with the embedded Java DB types integration tests. It's fast!

Providing a starting point / default `gradle.properties` and leaving it in `.gitignore` means that the build tasks are usable out of the box and provide a basis for further gradle customisations. Any more files would starting to eat into the KISS principle to me.

javornikolov:
Ah, look what I've found that we have here :-) [`test_vm/vendor/cookbooks/dbfit_test/recipes/gradle.rb`](https://github.com/dbfit/dbfit/blob/master/test_vm/vendor/cookbooks/dbfit_test/recipes/gradle.rb)

MMatten:
Why is this better than creating the file from the git repo? It's to control the build rather than provision the VM so why use Chef?

javornikolov:
> Why is this better than creating the file from the git repo?

In general - to me seems a bit tricky and obscure to commit a version-ignored file. It works but not very obvious. Actually I see drawback in both - of violation of SRP principle: other settings are relevant for `gradle.properties` with different reason to change than the list of tests to skip.

> It's to control the build rather than provision the VM so why use Chef?

The PR title is about Test VM. But you're right - when we run Test VM provisioning it suddenly may affect the configuration for builds within the host OS too. Maybe that's not very good.

Also I'm still not sure about the general default outside Test VM. The current default (empty list of exclusions) is something I'd like to consciously decide about and customize. (I'm doing that customization pretty rarely so I haven't thought so much about automating some default setup).

--
In direction about options for easy to run out-of-the box usage for typical cases - I'm wondering about alternatives:
* `fastbuild` - fast build of embedded databases
* `typical`, `standard` - free & open source stuff which is easily arranged
* `full`, `default`, `extended`
* `mycustom1`, `custom2`, ...

Some wondering about the syntax: `gradlew typicalBuild` vs `gradlew build -Pprofile=typical` (maybe the latter allows more options for customization).
Perhaps still a way to control the default behavior (when no profile is specified).


Maybe none of these is better than your proposal. I'm just trying to share some gut feelings and think about alternatives.

MMatten:
> SRP principle

I guess perhaps we can't be rid of potentially violating this as we could end up moving this configuration into the `build.gradle` files instead and have the same problem. 

May be new shell/batch scripts to wrap `gradlew` but that feels like we're introducing single line modules for the sake of it.

> Maybe none of these is better than your proposal. I'm just trying to share some gut feelings and think about alternatives.

No, of course. It's good to explore and challenge each others' thinking :+1: 

Perhaps a new build task, like `vmdefault`, would be best.

PR #516, opened 2016-10-23
test_vm: migrate from Librarian-Chef to Berkshelf

javornikolov:
* migrate test VM cookbook management from librarian-chef to
  berkshelf as vagrant-librarian-chef plugin is not supported
  any more
* postgresql dependency version is set to `>= 4` in order to
  get priority over openssl (otherwise Berkshelf picks a lower
  version of postgresql in favour of higher one for openssl)


javornikolov:
Functionally this should behave identically - just the cookbooks are maintained elsewhere (the folders are not under the project). At first glimpse the cookbooks are downloaded a bit faster with Berkshelf and with a bit more detailed logging.

MMatten:
Looks great.

I don't quite understand the comment about postgres vs OpenSSL. Can you explain again?

javornikolov:
> I don't quite understand the comment about postgres vs OpenSSL. Can you explain again?

The latest 4.x version of postgresql is not compatible with latest version of openssl 6.x. So the `highest-possible-version` method for selecting cookbook versions is not deterministic: it could be either postgresql 3.x and openssl 6.x or postgresql 4.x and openssl 4.x. Librarian-chef used to pick the latter option but Berkshelf - the former one. By adding the version constraint for postgresql I managed to make this choice more deterministic (and postgresql is more important to us than openssl).

MMatten:
Got it. Looks good to me. 

It will be good to move onto a new supported toy (Berkshelf) :)

javornikolov:
Cool. I'm merging...

Issue #490, opened 2016-02-23
Execute Procedure Expect Exception fails for procedures with no parameters


javornikolov:
The problem has been reported on the mailing list. Stack trace of a failing example as following;

```
!| Execute Procedure Expect Exception| DBFIT_delete_test_proc |
```

```
com.microsoft.sqlserver.jdbc.SQLServerException: Hard deletes of the dbo.test_table table prohibited.  Set deletedsw=1 instead.
	at com.microsoft.sqlserver.jdbc.SQLServerException.makeFromDatabaseError(SQLServerException.java:216)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.getNextResult(SQLServerStatement.java:1515)
	at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement.doExecutePreparedStatement(SQLServerPreparedStatement.java:404)
	at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement$PrepStmtExecCmd.doExecute(SQLServerPreparedStatement.java:350)
	at com.microsoft.sqlserver.jdbc.TDSCommand.execute(IOBuffer.java:5696)
	at com.microsoft.sqlserver.jdbc.SQLServerConnection.executeCommand(SQLServerConnection.java:1715)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.executeCommand(SQLServerStatement.java:180)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.executeStatement(SQLServerStatement.java:155)
	at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement.execute(SQLServerPreparedStatement.java:332)
	at dbfit.fixture.StatementExecution.run(StatementExecution.java:24)
	at dbfit.fixture.DbObjectExecutionFixture.doRows(DbObjectExecutionFixture.java:66)
	at fit.Fixture.doTable(Fixture.java:155)
	at fitlibrary.traverse.AlienTraverseHandler.doTable(AlienTraverseHandler.java:18)
	at fitlibrary.traverse.workflow.DoTraverseInterpreter.interpretWholeTable(DoTraverseInterpreter.java:99)
	at fitlibrary.traverse.workflow.DoTraverseInterpreter.interpretWholeTable(DoTraverseInterpreter.java:87)
	at fitlibrary.DoFixture.interpretWholeTable(DoFixture.java:69)
[...]

PR #522, opened 2016-11-26
Support stored procedure and function parameters in Derby

MMatten:
Support for executing stored procedures and functions in Derby.

Part of preparation for global support of `Execute Procedure Expect Exception` (part of fix to #490).

### DIFF
--- a/dbfit-java/derby/src/integration-test/java/dbfit/environment/DerbyRegressionSupportStoredProcs.java
+++ b/dbfit-java/derby/src/integration-test/java/dbfit/environment/DerbyRegressionSupportStoredProcs.java
@@ -18,4 +18,12 @@ public class DerbyRegressionSupportStoredProcs {
             stmt.execute(sql);
         }
     }
+
+    public static void calcLength(String name, int[] strLength) throws SQLException {
+        strLength[0] = name.length();
+    }
+
+    public static double multiply(double n1, double n2) throws SQLException {
+        return n1 * n2;
+    }
 }
--- a/dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
+++ b/dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
@@ -4,8 +4,10 @@ import dbfit.annotations.DatabaseEnvironment;
 import dbfit.api.AbstractDbEnvironment;
 import dbfit.util.DbParameterAccessor;
 import dbfit.util.NameNormaliser;
+import dbfit.util.Direction;
 
 import java.math.BigDecimal;
+import java.sql.DatabaseMetaData;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
@@ -16,7 +18,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.regex.Pattern;
 
-import dbfit.util.Direction;
+import static org.apache.commons.lang3.ObjectUtils.defaultIfNull;
 
 /**
  * Encapsulates support for the Derby database (also known as JavaDB). Operates
@@ -53,38 +55,131 @@ public class DerbyEnvironment extends AbstractDbEnvironment {
 
     public Map<String, DbParameterAccessor> getAllColumns(
             final String tableOrViewName) throws SQLException {
-        String qry = "SELECT COLUMNNAME, COLUMNDATATYPE "
+        String qry = "SELECT COLUMNNAME AS COLUMN_NAME, COLUMNDATATYPE AS TYPE_NAME, "
+                + " ROW_NUMBER() OVER() AS ORDINAL_POSITION, " +
+                + DatabaseMetaData.procedureColumnIn + " AS COLUMN_TYPE "
                 + "FROM SYS.SYSCOLUMNS WHERE REFERENCEID = "
-                + "(SELECT TABLEID FROM SYS.SYSTABLES WHERE TABLENAME = ?)";
-        return readIntoParams(tableOrViewName.toUpperCase(), qry);
+                + "(SELECT TABLEID FROM SYS.SYSTABLES WHERE TABLENAME = ?)"
+                + "ORDER BY COLUMNNUMBER";
+        return fetchIntoParams(tableOrViewName.toUpperCase(), qry);
     }
 
-    private Map<String, DbParameterAccessor> readIntoParams(
-            String tableOrViewName, String query) throws SQLException {
+    private Map<String, DbParameterAccessor> readIntoParams(ResultSet rs)
+                                                      throws SQLException {
+        Map<String, DbParameterAccessor> allParams = new HashMap<>();
+        while (rs.next()) {
+            String columnName = rs.getString("COLUMN_NAME");
+            DbParameterAccessor dbp = createObjectDbParamAccessor(columnName, rs, false);
+            allParams.put(NameNormaliser.normaliseName(columnName), dbp);
+        }
+        return allParams;
+    }
+
+    private Map<String, DbParameterAccessor> fetchIntoParams(String tableOrViewName, String query)
+                                                                               throws SQLException {
         try (PreparedStatement dc = getConnection().prepareStatement(query)) {
             dc.setString(1, tableOrViewName);
-
-            ResultSet rs = dc.executeQuery();
-            Map<String, DbParameterAccessor> allParams = new HashMap<String, DbParameterAccessor>();
-            int position = 0;
-            while (rs.next()) {
-                String columnName = rs.getString(1);
-                String dataType = rs.getString(2);
-                DbParameterAccessor dbp = createDbParameterAccessor(
-                        columnName,
-                        Direction.INPUT,
-                        typeMapper.getJDBCSQLTypeForDBType(dataType),
-                        getJavaClass(dataType), position++);
-                allParams.put(NameNormaliser.normaliseName(columnName), dbp);
+            try(ResultSet rs = dc.executeQuery()) {
+                return readIntoParams(rs);
             }
-            rs.close();
-            return allParams;
         }
     }
 
-    public Map<String, DbParameterAccessor> getAllProcedureParameters(
-            String procName) throws SQLException {
-        throw new UnsupportedOperationException();
+    private boolean isProcedureOrFunction(DbObjectName qn, boolean checkForFunction)
+                                                                  throws SQLException {
+        DatabaseMetaData dbmd = getConnection().getMetaData();
+        try (ResultSet rs = checkForFunction ? dbmd.getFunctions(null, qn.getSchemaName(), qn.getObjectName()) :
+                                               dbmd.getProcedures(null, qn.getSchemaName(), qn.getObjectName())) {
+            return rs.next();
+        }
+    }
+
+    private Direction getParamDirection(int columnType, boolean isFunction) {
+        if (isFunction) {
+            switch (columnType) {
+                case DatabaseMetaData.functionColumnIn:
+                    return Direction.INPUT;
+                case DatabaseMetaData.functionReturn:
+                    return Direction.RETURN_VALUE;
+            }
+        } else {
+            switch (columnType) {
+                case DatabaseMetaData.procedureColumnIn:
+                    return Direction.INPUT;
+                case DatabaseMetaData.procedureColumnInOut:
+                    return Direction.INPUT_OUTPUT;
+                case DatabaseMetaData.procedureColumnOut:
+                    return Direction.OUTPUT;
+            }
+        }
+        throw new UnsupportedOperationException("Type " + columnType
+                      + " is not a supported type for "
+                      + (isFunction ? "function" : "procedure") + " parameters");
+    }
+
+    private DbParameterAccessor createObjectDbParamAccessor(String paramName,
+                                                            ResultSet rs,
+                                                            boolean isFunction)
+                                                            throws SQLException {
+        Direction paramDirection = getParamDirection(rs.getInt("COLUMN_TYPE"), isFunction);
+        String paramTypeName = rs.getString("TYPE_NAME");
+        int position = rs.getInt("ORDINAL_POSITION"); // We rely upon Derby returning 0 for function return value.
+        return createDbParameterAccessor(paramName, paramDirection,
+                   typeMapper.getJDBCSQLTypeForDBType(paramTypeName), getJavaClass(paramTypeName),
+                   position - 1);
+    }
+
+    private ResultSet getObjectParameters(DbObjectName qn, boolean isFunction)
+                                                            throws SQLException {
+        DatabaseMetaData dbmd = getConnection().getMetaData();
+        return isFunction ? dbmd.getFunctionColumns(null, qn.getSchemaName(), qn.getObjectName(), "%")
+                          : dbmd.getProcedureColumns(null, qn.getSchemaName(), qn.getObjectName(), "%");
+    }
+
+    private Map<String, DbParameterAccessor> getParamMap(DbObjectName qn,
+                                                         boolean isFunction)
+                                                         throws SQLException {
+        Map<String, DbParameterAccessor> allParams = new HashMap<>();
+        try (ResultSet rs = getObjectParameters(qn, isFunction)) {
+            while (rs.next()) {
+                String paramName = defaultIfNull(rs.getString("COLUMN_NAME"), "");
+                allParams.put(NameNormaliser.normaliseName(paramName),
+                              createObjectDbParamAccessor(paramName, rs, isFunction));
+            }
+        }
+        return allParams;
+    }
+
+    private class DbObjectName {
+        String schemaName, objectName;
+
+        private DbObjectName(String defaultSchemaName, String objName) {
+            String[] qualifiers = objName.split("\\.");
+            schemaName = (qualifiers.length == 2) ? qualifiers[0] : defaultSchemaName;
+            objectName = (qualifiers.length == 2) ? qualifiers[1] : qualifiers[0];
+        }
+
+        private String getSchemaName() {
+            return schemaName;
+        }
+
+        private String getObjectName() {
+            return objectName;
+        }
+    }
+
+    public Map<String, DbParameterAccessor> getAllProcedureParameters(String procName)
+                                                                   throws SQLException {
+        DbObjectName qn = new DbObjectName(getConnection().getSchema(), procName.toUpperCase());
+        boolean isFunction;
+        if (isProcedureOrFunction(qn, false)) {
+            isFunction = false;
+        } else if (isProcedureOrFunction(qn, true)) {
+            isFunction = true;
+        } else {
+            throw new SQLException("Cannot find Procedure or Function with name " + procName);
+        }
+        return getParamMap(qn, isFunction);
     }
 
     public Class<?> getJavaClass(final String dataType) {
--- a/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/FlowMode/CallingFunctions.wiki
+++ b/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/FlowMode/CallingFunctions.wiki
@@ -1,0 +1,16 @@
+---
+Test
+---
+!3 Functions are tested like procedures, with ? used for the result
+
+|Execute procedure|Multiply|
+|n1|n2|?|
+|10|15|150|
+|2|8|16|
+
+Order of parameters in table should not be important (does not have to match db order)
+
+|Execute procedure|Multiply|
+|n2|?|n1|
+|15|150|10|
+|8|16|2|
--- a/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/FlowMode/SetUp/content.txt
+++ b/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/FlowMode/SetUp/content.txt
@@ -6,4 +6,8 @@
 
 |Execute Ddl|${CREATEMAKEUSERPROC}|
 
+|Execute Ddl|${CREATECALCLENGTHPROC}|
+
+|Execute Ddl|${CREATEMULTIPLYPROC}|
+
 |Commit|
--- a/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/FlowMode/SimpleStoredProc.wiki
+++ b/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/FlowMode/SimpleStoredProc.wiki
@@ -1,0 +1,10 @@
+---
+Test
+---
+This stored proc tests output parameters - !-CalcLength-! returns string length as second output param
+
+|Execute Procedure       |!-CalcLength-!|
+|name                    |str length?   |
+|mika                    |4             |
+|paradajz                |8             |
+|saeiluhrweurhfi7fhi7rhgf|24            |
--- a/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/FlowMode/TearDown/content.txt
+++ b/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/FlowMode/TearDown/content.txt
@@ -4,4 +4,8 @@
 
 |Execute Ddl|${DROPMAKEUSERPROC}|
 
+|Execute Ddl|${DROPCALCLENGTHPROC}|
+
+|Execute Ddl|${DROPMULTIPLYPROC}|
+
 |Commit|
--- a/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/StandaloneFixtures/SetUp/content.txt
+++ b/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/StandaloneFixtures/SetUp/content.txt
@@ -8,5 +8,9 @@
 
 |Execute Ddl|${CREATEMAKEUSERPROC}|
 
+|Execute Ddl|${CREATECALCLENGTHPROC}|
+
+|Execute Ddl|${CREATEMULTIPLYPROC}|
+
 |Database Environment|
 |Commit|
--- a/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/StandaloneFixtures/TearDown/content.txt
+++ b/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/StandaloneFixtures/TearDown/content.txt
@@ -5,6 +5,10 @@
 
 |Execute Ddl|${DROPMAKEUSERPROC}|
 
+|Execute Ddl|${DROPCALCLENGTHPROC}|
+
+|Execute Ddl|${DROPMULTIPLYPROC}|
+
 |Database Environment|
 |Commit|
 
--- a/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/content.txt
+++ b/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/content.txt
@@ -2,5 +2,9 @@
 !define DROPUSERTABLE {DROP TABLE USERS}
 !define CREATEMAKEUSERPROC {CREATE PROCEDURE !-MakeUser-!() LANGUAGE JAVA PARAMETER STYLE JAVA EXTERNAL NAME '!-dbfit.environment.DerbyRegressionSupportStoredProcs.MakeUser-!'}
 !define DROPMAKEUSERPROC {DROP PROCEDURE !-MakeUser-!}
+!define CREATECALCLENGTHPROC {CREATE PROCEDURE !-CalcLength-!(IN !-Name-! VARCHAR(200), OUT !-StrLength-! INTEGER) LANGUAGE JAVA PARAMETER STYLE JAVA EXTERNAL NAME '!-dbfit.environment.DerbyRegressionSupportStoredProcs.calcLength-!'}
+!define DROPCALCLENGTHPROC {DROP PROCEDURE !-CalcLength-!}
+!define CREATEMULTIPLYPROC {CREATE FUNCTION !-Multiply-!(n1 DOUBLE, n2 DOUBLE) RETURNS DOUBLE LANGUAGE JAVA PARAMETER STYLE JAVA EXTERNAL NAME '!-dbfit.environment.DerbyRegressionSupportStoredProcs.multiply-!'}
+!define DROPMULTIPLYPROC {DROP FUNCTION !-Multiply-!}
 
 !contents -R


javornikolov:
Ah, it's interesting to see that FitNesse has changed the default format of storing tests (`.wiki` files instead of folder + `content.txt`). I guess some day would be good to migrate the existing tests to the new style.

MMatten:
Yes. May be someone's got a nice migration utility script out there somewhere.

MMatten:
Note that I've only included the tests in the `FlowMode` suite for now I'm planning on revising the stored proc/function tests and I think we can now get more tests into the `CoreTests` suite.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -84,7 +85,80 @@ public Pattern getParameterPattern() {
...
+            schemaName = qualifiers[0];
+            procedureName = qualifiers[1];
+        } else {
+            schemaName = getConnection().getSchema();
+            procedureName = qualifiers[0];
+        }
+        procedureName = procedureName.toUpperCase();
+        DatabaseMetaData dbmd = getConnection().getMetaData();
+        ResultSet rs;
+        rs = dbmd.getProcedures(null, schemaName, procedureName);
```
javornikolov: Is it possible to have the same name both as function and as a procedure? (It's good to know if we have such limitation here. I think we don't properly support such cases in other database types either yet).

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -84,7 +85,80 @@ public Pattern getParameterPattern() {
...
+        } else {
+            rs.close();
+            rs = dbmd.getFunctions(null, schemaName, procedureName);
+            if (rs.next()) {
+                isFunction = true;
+                rs.close();
+                rs = dbmd.getFunctionColumns(null, schemaName, procedureName, "%");
+            }
+        }
+        Map<String, DbParameterAccessor> allParams = new HashMap<String, DbParameterAccessor>();
```
javornikolov: For sake of simplicity - would that work with just `new HashMap<>()` constructor call?

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -84,7 +85,80 @@ public Pattern getParameterPattern() {
...
+                    case DatabaseMetaData.procedureColumnIn:
+                        paramDirection = Direction.INPUT;
+                        break;
+                    case DatabaseMetaData.procedureColumnInOut:
+                        paramDirection = Direction.INPUT_OUTPUT;
+                        break;
+                    case DatabaseMetaData.procedureColumnOut:
+                        paramDirection = Direction.OUTPUT;
+                        break;
+                    case DatabaseMetaData.procedureColumnReturn:
```
javornikolov: Is it possible to have `RETURN_VALUE` for a procedure?

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -84,7 +85,80 @@ public Pattern getParameterPattern() {
 
     public Map<String, DbParameterAccessor> getAllProcedureParameters(
             String procName) throws SQLException {
-        throw new UnsupportedOperationException();
+        String[] qualifiers = procName.split("\\.");
```
MMatten: Ok. That's the existing name and it implements the method from the `DBEnvironment` interface.

dbfit-java/derby/src/integration-test/java/dbfit/environment/DerbyRegressionSupportStoredProcs.java
```diff
@@ -18,4 +18,12 @@ public static void MakeUser() throws SQLException {
             stmt.execute(sql);
         }
     }
+
+    public static void CalcLength(String Name, int[] StrLength) throws SQLException {
```
MMatten: Not really as far as I can tell. I'll standardise them.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: > Is it possible to have `RETURN_VALUE` for a procedure?

It's not possible in `Derby`.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: It appears that this is possible in Derby. So we could pick the wrong one. May be in other DB types too.

We have no way of knowing what type the `Execute Procedure` fixture is expecting with the current code base.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: Yes. I didn't realise this was acceptable but compiler seems to infer the specific types for the generic constructor from the variable definition.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -84,7 +85,80 @@ public Pattern getParameterPattern() {
...
+                isFunction = true;
+                rs.close();
+                rs = dbmd.getFunctionColumns(null, schemaName, procedureName, "%");
+            }
+        }
+        Map<String, DbParameterAccessor> allParams = new HashMap<String, DbParameterAccessor>();
+        int position = 0;
+        while (rs.next()) {
+            String paramName = rs.getString("COLUMN_NAME");
+            if (paramName == null)
```
MMatten: Yes. Thanks. Will apply.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: Really? I could trim it down to 80 chars (or so). It's a bit 1970s green-on-black style though! What do you think the limit should be?

Personally I find split method parms and arguments split over lines to not look good.

MMatten: I guess the trade off of moving some logic out to other methods is that some people will find reading though the method hard because of the need to jump about to different methods instead of having a continuous flow.

I'll see what it looks like with some parts moved out.

MMatten:
I've had a go at refactoring to break up the long method, at the expense of some duplicated code. Let me know what you think or other suggestions are welcomed.

javornikolov:
> I've had a go at refactoring to break up the long method, at the expense of some duplicated code. Let me know what you think or other suggestions are welcomed.

Can we simplify it further - including reduction of duplicated code?

MMatten:
> Can we simplify it further - including reduction of duplicated code?

What do you think of this one? I think it's easier to follow now but there's a lot of code. Can you spot any good simplifications?

MMatten:
I've kicked of the Travis build again (it couldn't download the Oracle JDBC driver).

javornikolov:
> What do you think of this one? I think it's easier to follow now but there's a lot of code. Can you spot any good simplifications?

Here are a few ideas:
* `isFunction`/`isProcedure` share a common pattern - can be extracted into something which checks if given result set contains any records. (Could possibly also take care of closing it).
* Looping through parameters result set and constructing accessors out of it basically the same for functions and procedures:
  - The different part is something which I believe we can extract to a single place. (We can trust the JDBC driver and don't bother so much if `getFunctionColumns` is returning procedure-specific types).

MMatten:
> isFunction/isProcedure share a common pattern - can be extracted into something which checks if given result set contains any records. (Could possibly also take care of closing it).

I've combined these now. I've let the fact that `ResultSet` is `AutoCloseable`able close it.

> Looping through parameters result set and constructing accessors out of it basically the same for functions and procedures: 

This is already one method but I've condensed a few more bits.

What do you think now?

javornikolov:
It looks better now! I would go on a bit further.

E.g. - I guess a single `switch` statement would be enough. The `boolean` flag/parameter/checks I think can be reduced - essentially the only place it's really needed is in opening the parameters result set. There is an `ORDINAL_POSITION` metadata column to calculate the parameter position.

MMatten:
> The boolean flag/parameter/checks I think can be reduced - essentially the only place it's really needed is in opening the parameters result set.

As the `DatabaseMetaData` constants are different for functions and procedures I don't see how I can treat functions and procedures different when checking the parameters result set. What were you thinking we could do?

MMatten:
I guess we can check `functionColumnIn` or `procedureColumnIn` gives us `INPUT`.

But we `functionReturn` and `procedureColumnIn` both *currently* have the same integer value and so would map to either `RETURN_VALUE` or `OUTPUT`.

javornikolov:
> But we functionReturn and procedureColumnIn both currently have the same integer value and so would map to either RETURN_VALUE or OUTPUT.

Ah, that's a good point... I thought all constants have different values. I think `functionReturn` is not applicable for column types, all relevant constants are `functionColumn*` and `procedureColumn*`, and they seem to be aligned for what we currently support. So maybe these can be combined after all though I admit it would be a bit obscure.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -82,9 +84,95 @@ public Pattern getParameterPattern() {
...
+    }
+
+    private Direction getParamDirection(int columnType, boolean isFunction) {
+        Direction dir = null;
+        if (isFunction) {
+            switch (columnType) {
+                case DatabaseMetaData.functionColumnIn:
+                    dir = Direction.INPUT;
+                    break;
+                case DatabaseMetaData.functionReturn:
```
javornikolov: Isn't it possible a function to also have an output parameter?

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -82,9 +84,95 @@ public Pattern getParameterPattern() {
...
+        }
+    }
+
+    private Direction getParamDirection(int columnType, boolean isFunction) {
+        Direction dir = null;
+        if (isFunction) {
+            switch (columnType) {
+                case DatabaseMetaData.functionColumnIn:
+                    dir = Direction.INPUT;
+                    break;
```
javornikolov: These could be shortcut `return` statements to shrink a bit the number of lines and to avoid the need of `dir` variable.

MMatten:
> and they seem to be aligned for what we currently support.

If feels like we'd be making ourselves dependent upon upon these not changing in future JDBC versions. It might be safer to keep them separate.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -82,9 +84,95 @@ public Pattern getParameterPattern() {
...
+            allParams.put(NameNormaliser.normaliseName(paramName), dbp);
+        }
+        return allParams;
+    }
+
     public Map<String, DbParameterAccessor> getAllProcedureParameters(
             String procName) throws SQLException {
-        throw new UnsupportedOperationException();
+        String[] qualifiers = procName.split("\\.");
+        String schemaName, procedureName;
```
javornikolov: That parsing also looks like a candidate for refactoring.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -82,9 +84,95 @@ public Pattern getParameterPattern() {
...
+                                                         throws SQLException {
+        DatabaseMetaData dbmd = getConnection().getMetaData();
+        int position = 0;
+        Direction paramDirection;
+        String dataType, paramName;
+        Map<String, DbParameterAccessor> allParams = new HashMap<>();
+        ResultSet rs = isFunction ? dbmd.getFunctionColumns(null, schemaName, procedureName, "%") :
+                                    dbmd.getProcedureColumns(null, schemaName, procedureName, "%");
+        while (rs.next()) {
+            paramName = defaultIfNull(rs.getString("COLUMN_NAME"), "");
```
javornikolov: The details of unwrapping a of single parameter into accessor is a candidate to be extracted.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -82,9 +84,95 @@ public Pattern getParameterPattern() {
...
+    private Map<String, DbParameterAccessor> getParamMap(String schemaName,
+                                                         String procedureName,
+                                                         boolean isFunction)
+                                                         throws SQLException {
+        DatabaseMetaData dbmd = getConnection().getMetaData();
+        int position = 0;
+        Direction paramDirection;
+        String dataType, paramName;
+        Map<String, DbParameterAccessor> allParams = new HashMap<>();
+        ResultSet rs = isFunction ? dbmd.getFunctionColumns(null, schemaName, procedureName, "%") :
```
javornikolov: Do we close that `ResultSet` somewhere?

MMatten: Good spot.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: If we did this then would this do for us?

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: Ok. I have created a couple of new methods for this.

MMatten:
Ok. Some changes pushed. Let me know what you think again please.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -82,9 +84,92 @@ public Pattern getParameterPattern() {
...
+        try (ResultSet rs = isFunction ? dbmd.getFunctionColumns(null, schemaName, procedureName, "%") :
+                                         dbmd.getProcedureColumns(null, schemaName, procedureName, "%")) {
+            while (rs.next()) {
+                paramName = defaultIfNull(rs.getString("COLUMN_NAME"), "");
+                paramDirection = getParamDirection(rs.getInt("COLUMN_TYPE"), isFunction);
+                dataType = rs.getString("TYPE_NAME");
+                DbParameterAccessor dbp =
+                    createDbParameterAccessor(paramName, paramDirection,
+                        typeMapper.getJDBCSQLTypeForDBType(dataType), getJavaClass(dataType),
+                        (isFunction ? (paramDirection == Direction.RETURN_VALUE ? -1 : position++) :
```
javornikolov: Would it work to get position from `ORDINAL_POSITION` (http://docs.oracle.com/javadb/10.10.1.2/ref/rrefcrsrgpc1.html). I think it should return 0 for return values but it's good to double check.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -82,9 +84,92 @@ public Pattern getParameterPattern() {
...
+    private Map<String, DbParameterAccessor> getParamMap(String schemaName,
+                                                         String procedureName,
+                                                         boolean isFunction)
+                                                         throws SQLException {
+        DatabaseMetaData dbmd = getConnection().getMetaData();
+        int position = 0;
+        Direction paramDirection;
+        String dataType, paramName;
+        Map<String, DbParameterAccessor> allParams = new HashMap<>();
+        try (ResultSet rs = isFunction ? dbmd.getFunctionColumns(null, schemaName, procedureName, "%") :
```
javornikolov: What about extracting that `get...Columns` expression as separate method?

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -82,9 +84,92 @@ public Pattern getParameterPattern() {
...
+                      + (isFunction ? "function" : "procedure") + " parameters");
+    }
+
+    private Map<String, DbParameterAccessor> getParamMap(String schemaName,
+                                                         String procedureName,
+                                                         boolean isFunction)
+                                                         throws SQLException {
+        DatabaseMetaData dbmd = getConnection().getMetaData();
+        int position = 0;
+        Direction paramDirection;
```
javornikolov: Some of these variables are only used in the loop below - can be declared there in place. But assembling accesso out of jdbc record can also go as separate method.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: > I think it should return 0 for return values but it's good to double check.

It does return 0 for return values. Is this good practice though / is it reliable?

javornikolov: Yes, it should be reliable enough. It's also documented in the JDBC API docs.

javornikolov: https://docs.oracle.com/javase/7/docs/api/java/sql/DatabaseMetaData.html#getFunctionColumns(java.lang.String,%20java.lang.String,%20java.lang.String,%20java.lang.String)

MMatten: Awesome. Thanks.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: What does doing that do for us? We'll have more code.

javornikolov: Seems a diverging thought at a bit lower level of detail. But it's just a point which made me wonder, not necessarily a good idea. Another stretch (possibly one going too far) around that could be swapping the if/conditional with polymorphic hierarchy.

In general my wondering was mostly about [SLAP, SRP, DRY principles](http://nithinbekal.com/notes/design-principles/). Which have their pros but cons too. So it's up to experimentation and to your judgement.

MMatten:
Next version pushed for review.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -82,9 +84,92 @@ public Pattern getParameterPattern() {
...
+            throws SQLException {
+        String[] qualifiers = objectName.split("\\.");
+        return (qualifiers.length == 2) ? qualifiers[1] : qualifiers[0];
+    }
+
     public Map<String, DbParameterAccessor> getAllProcedureParameters(
             String procName) throws SQLException {
-        throw new UnsupportedOperationException();
+        String schemaName = getObjectSchemaName(procName.toUpperCase());
+        String procedureName = getObjectObjectName(procName.toUpperCase());
```
MMatten: I've created `DbObjectName`.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -82,9 +84,100 @@ public Pattern getParameterPattern() {
...
+    private DbParameterAccessor crtProcDbParamAccessor(ResultSet rs,
+                                                       boolean isFunction)
+                                                       throws SQLException {
+        String paramName = defaultIfNull(rs.getString("COLUMN_NAME"), "");
+        Direction paramDirection = getParamDirection(rs.getInt("COLUMN_TYPE"), isFunction);
+        String paramTypeName = rs.getString("TYPE_NAME");
+        int position = rs.getInt("ORDINAL_POSITION"); // We rely upon Derby returning 0 for function return value.
+        return createDbParameterAccessor(paramName, paramDirection,
+                   typeMapper.getJDBCSQLTypeForDBType(paramTypeName), getJavaClass(paramTypeName),
+                   (position > 0 ? position : -1));
```
javornikolov: The accessor index is 0-based so I think we should always subtract 1.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -82,9 +84,100 @@ public Pattern getParameterPattern() {
...
+                case DatabaseMetaData.procedureColumnOut:
+                    return Direction.OUTPUT;
+            }
+        }
+        throw new UnsupportedOperationException("Type " + columnType
+                      + " is not a supported type for "
+                      + (isFunction ? "function" : "procedure") + " parameters");
+    }
+
+    private DbParameterAccessor crtProcDbParamAccessor(ResultSet rs,
```
javornikolov: The name looks a bit cryptic.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -82,9 +84,100 @@ public Pattern getParameterPattern() {
...
+                          : dbmd.getProcedureColumns(null, qn.getSchemaName(), qn.getObjectName(), "%");
+    }
+
+    private Map<String, DbParameterAccessor> getParamMap(DbObjectName qn,
+                                                         boolean isFunction)
+                                                         throws SQLException {
+        Map<String, DbParameterAccessor> allParams = new HashMap<>();
+        try (ResultSet rs = getObjectParameters(qn, isFunction)) {
+            while (rs.next()) {
+                String paramName = defaultIfNull(rs.getString("COLUMN_NAME"), "");
```
javornikolov: Can we avoid that code duplication (the same line is in `crtProcDbParamAccessor`)?

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: Ah, yes. I forgot to tidy that one.

MMatten: I'm still not even sure about something like `createObjectDbParamAccessor` as it sounds like it should also create them for DB tables too. Perhaps we could handle the accessors created for table columns too in `readIntoParams` (by making the columns query return the same column names).

javornikolov: BTW - perhaps the table columns metadata can also be extracted from the jdbc DatabaseMetaData (but maybe columns will be a bit different). Or vice versa (a raw SQL query against Derby sys tables to extract procedure/function parameters info).

javornikolov: What about `readProcedureParameter`? Or `readParameter` in case we manage to generalize, or in case we organize it into a class hierarchy.

MMatten:
Another commit added to see what things look like the table column accessor creation reworked too.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -55,32 +55,32 @@ public Pattern getParameterPattern() {
 
     public Map<String, DbParameterAccessor> getAllColumns(
             final String tableOrViewName) throws SQLException {
-        String qry = "SELECT COLUMNNAME, COLUMNDATATYPE "
+        String qry = "SELECT COLUMNNAME AS COLUMN_NAME, COLUMNDATATYPE AS TYPE_NAME, "
+                + " ROW_NUMBER() OVER() AS ORDINAL_POSITION, " +
```
javornikolov: Hmm, it's not something new but now I see an issue in that query... I wouldn't rely on rows order in absence of an `order by` clause. There is an `COLUMNNUMBER` column which perhaps should help for that..

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -116,16 +116,16 @@ private Direction getParamDirection(int columnType, boolean isFunction) {
                       + (isFunction ? "function" : "procedure") + " parameters");
     }
 
-    private DbParameterAccessor crtProcDbParamAccessor(ResultSet rs,
-                                                       boolean isFunction)
-                                                       throws SQLException {
-        String paramName = defaultIfNull(rs.getString("COLUMN_NAME"), "");
+    private DbParameterAccessor createObjectDbParamAccessor(String paramName,
```
javornikolov: Some doubts about the name, do we need the `Object` part (gets me thinking whether we have accessors to non-objects).

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -55,32 +55,32 @@ public Pattern getParameterPattern() {
...
+                + DatabaseMetaData.procedureColumnIn + " AS COLUMN_TYPE "
                 + "FROM SYS.SYSCOLUMNS WHERE REFERENCEID = "
                 + "(SELECT TABLEID FROM SYS.SYSTABLES WHERE TABLENAME = ?)";
-        return readIntoParams(tableOrViewName.toUpperCase(), qry);
+        return fetchIntoParams(tableOrViewName.toUpperCase(), qry);
     }
 
-    private Map<String, DbParameterAccessor> readIntoParams(
-            String tableOrViewName, String query) throws SQLException {
+    private Map<String, DbParameterAccessor> readIntoParams(ResultSet rs)
```
javornikolov: `readIntoParams` and `getObjectParameters` are now pretty similar. How can we resolve that code duplication?

javornikolov:
> Another commit added to see what things look like the table column accessor creation reworked too.

As of now it looks like to be a good idea. Maybe even more can be generalized if we use the same approach for deriving metadata for both table columns and procedure parameters (i.e. either both SQL queries or both JDBC metadata calls).

MMatten:
> if we use the same approach for deriving metadata for both table columns and procedure parameter

It's not possible to use a SQL query to get the procedure parameters. I did spend a while trying to find out but there seems to be no dictionary data for them.

javornikolov:
> It's not possible to use a SQL query to get the procedure parameters. I did spend a while trying to find out but there seems to be no dictionary data for them.

There are queries in the link I sent: https://github.com/apache/derby/blob/10.13/java/engine/org/apache/derby/impl/jdbc/metadata.properties

MMatten:
That's interesting looking SQL as it's laced with Java. I'll see if it actually runs from a client.

MMatten:
> That's interesting looking SQL as it's laced with Java. I'll see if it actually runs from a client.

I've tried executing the query: -
```
SELECT 
	CAST(NULL AS VARCHAR(128)) AS FUNCTION_CAT, 
	S.SCHEMANAME       AS FUNCTION_SCHEM, 
	A.ALIAS            AS FUNCTION_NAME, 
	V."COLUMN_NAME"    AS COLUMN_NAME, 
	V."COLUMN_TYPE"    AS COLUMN_TYPE, 
	V."DATA_TYPE"      AS DATA_TYPE, 
	V."TYPE_NAME"      AS TYPE_NAME, 
	V."PRECISION"      AS PRECISION, 
	V."LENGTH"         AS LENGTH, 
  	CASE WHEN (V."DATA_TYPE" IN (java.sql.Types::DECIMAL, 
 		java.sql.Types::NUMERIC, java.sql.Types::INTEGER, 
 		java.sql.Types::SMALLINT, java.sql.Types::TINYINT, 
 		java.sql.Types::BIGINT, java.sql.Types::DATE, 
 		java.sql.Types::TIME, java.sql.Types::TIMESTAMP)) 
	THEN 
	V."SCALE" 
	ELSE CAST (NULL AS SMALLINT) END 
	AS SCALE, 
 	CASE WHEN (V."DATA_TYPE" IN (java.sql.Types::DECIMAL, 
 		java.sql.Types::NUMERIC, java.sql.Types::INTEGER, 
 		java.sql.Types::SMALLINT, java.sql.Types::TINYINT, 
 		java.sql.Types::BIGINT, java.sql.Types::DOUBLE, 
 		java.sql.Types::FLOAT, java.sql.Types::REAL, 
 		java.sql.Types::DATE, java.sql.Types::TIME, 
 		java.sql.Types::TIMESTAMP)) 
	THEN V."RADIX" 
	ELSE CAST (NULL AS SMALLINT) END AS RADIX, 
   	V."NULLABLE"          AS NULLABLE, 
	V."REMARKS"           AS REMARKS, 
	CASE WHEN (V."DATA_TYPE" IN ( 
		 java.sql.Types::CHAR, 
		 java.sql.Types::VARCHAR, 
		 java.sql.Types::BINARY, 
		 java.sql.Types::VARBINARY)) 
		 THEN V."LENGTH" 
		 ELSE CAST(NULL AS INT) 
	END AS CHAR_OCTET_LENGTH, 
	CASE WHEN (V."COLUMN_TYPE" = 5) 
		 THEN CAST((V."PARAMETER_ID" + 1 - V."METHOD_ID") AS INT) 
		 ELSE CAST((V."PARAMETER_ID" + 1) AS INT) 
	END AS ORDINAL_POSITION, 
	CAST(( 
	CASE 
		 WHEN V."NULLABLE" = java.sql.DatabaseMetaData::procedureNullable 
		 THEN 'YES' ELSE 'NO' 
	END) AS VARCHAR(128)) AS IS_NULLABLE, 
	A.SPECIFICNAME        AS SPECIFIC_NAME, 
	V."METHOD_ID"         AS METHOD_ID, 
	V."PARAMETER_ID"      AS PARAMETER_ID 
FROM 
     SYS.SYSALIASES A, SYS.SYSSCHEMAS S, 
     NEW org.apache.derby.catalog.GetProcedureColumns(A.ALIASINFO, 
                                                      A.ALIASTYPE) V 
WHERE 
      A.ALIASTYPE = 'F' 
      AND S.SCHEMANAME LIKE '%'
      AND A.SCHEMAID = S.SCHEMAID 
      AND A.ALIAS LIKE '%'
      AND (V."COLUMN_NAME" LIKE '%') 
ORDER BY FUNCTION_SCHEM, FUNCTION_NAME, SPECIFIC_NAME, PARAMETER_ID, ORDINAL_POSITION
```

in both SquirrelSQL and via a DbFit `Query` fixture and I get a syntax error via both clients:
```
Error: Syntax error: java.sql.Types::DECIMAL.
SQLState:  42X01
```
so I don't think we can use this and will have to stick with `getProcedureColumns`.

MMatten:
I might have been a little hasty here, as I can execute this cut down version: -

```
SELECT *
FROM 
     SYS.SYSALIASES A, SYS.SYSSCHEMAS S, 
     NEW org.apache.derby.catalog.GetProcedureColumns(A.ALIASINFO, 
                                                      A.ALIASTYPE) V 
WHERE 
      A.ALIASTYPE = 'F' 
      AND S.SCHEMANAME LIKE '%'
      AND A.SCHEMAID = S.SCHEMAID 
      AND A.ALIAS LIKE '%'
      AND (V."COLUMN_NAME" LIKE '%') 
```
I play around some more...

MMatten:
It's just the constants such as `java.sql.DatabaseMetaData::procedureNullable` that it doesn't like.

We could get away with substituting for them literals of course or just stick with `getProcedure/FunctionColumns`.

As for a common approach for tables/functions/procedures; I don't think we have any requirements to use dictionary queries at the moment for Derby so the JDBC methods are probably neater. What do you think?

javornikolov:
The JDBC is using these queries somehow (maybe replacing the constants behind the scenes or in some other way). Or the raw values may be fetched from the query directly and any mappings done on Java side (comparison to the Java constants, etc).

But to be honest - for the current functionality we have I think we don't need to involve any of these constants.

MMatten:
> But to be honest - for the current functionality we have I think we don't need to involve any of these constants.

True, and we could just use the JDBC `getColumns/getFunctionColumns/getProcedureColumns` instead of dictionary queries too. I think the code would be more compact.

javornikolov:
Yes, that's an option too. I don't know which option would be cleaner.

Ideally it's good to avoid/minimize the places of decision branching based on the object type (table, function, procedure).

With plain SQL queries the moment of unification of data structure could be done earlier (the SQL query text). With JDBC metadata API we need the unification layer on top of result sets. Perhaps it needs some experimentation to see which would look better.

MMatten:
I've just drafted an idea out on #524.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: Update applied and force-pushed the revised version.

MMatten:
I guess I'll go ahead on this PR and see what things look like if we pull out procedure and function parameters using SQL queries.

MMatten:
Superseded by #524.

PR #524, opened 2016-12-03
Support stored and function parameters in Derby

MMatten:
Initial draft of one approach to using only JDBC method for querying Derby object metadata.

Potential for reuse for a generic DB adapter(?)

### DIFF
--- a/dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
+++ b/dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
@@ -3,10 +3,11 @@ package dbfit.environment;
 import dbfit.annotations.DatabaseEnvironment;
 import dbfit.api.AbstractDbEnvironment;
 import dbfit.util.DbParameterAccessor;
-import dbfit.util.NameNormaliser;
+import dbfit.util.Direction;
+import static dbfit.util.NameNormaliser.normaliseName;
 
 import java.math.BigDecimal;
-import java.sql.PreparedStatement;
+import java.sql.DatabaseMetaData;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Time;
@@ -15,8 +16,9 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.regex.Pattern;
+import static java.util.Objects.requireNonNull;
 
-import dbfit.util.Direction;
+import static org.apache.commons.lang3.ObjectUtils.defaultIfNull;
 
 /**
  * Encapsulates support for the Derby database (also known as JavaDB). Operates
@@ -51,42 +53,189 @@ public class DerbyEnvironment extends AbstractDbEnvironment {
         return paramRegex;
     }
 
-    public Map<String, DbParameterAccessor> getAllColumns(
-            final String tableOrViewName) throws SQLException {
-        String qry = "SELECT COLUMNNAME, COLUMNDATATYPE "
-                + "FROM SYS.SYSCOLUMNS WHERE REFERENCEID = "
-                + "(SELECT TABLEID FROM SYS.SYSTABLES WHERE TABLENAME = ?)";
-        return readIntoParams(tableOrViewName.toUpperCase(), qry);
+    @Override
+    public Map<String, DbParameterAccessor> getAllColumns(String tableName)
+            throws SQLException {
+        return new DatabaseTable(buildDatabaseObjectName(tableName)).getParams();
     }
 
-    private Map<String, DbParameterAccessor> readIntoParams(
-            String tableOrViewName, String query) throws SQLException {
-        try (PreparedStatement dc = getConnection().prepareStatement(query)) {
-            dc.setString(1, tableOrViewName);
+    private abstract class DatabaseObject {
+        protected DatabaseObjectName name;
 
-            ResultSet rs = dc.executeQuery();
-            Map<String, DbParameterAccessor> allParams = new HashMap<String, DbParameterAccessor>();
-            int position = 0;
-            while (rs.next()) {
-                String columnName = rs.getString(1);
-                String dataType = rs.getString(2);
-                DbParameterAccessor dbp = createDbParameterAccessor(
-                        columnName,
-                        Direction.INPUT,
-                        typeMapper.getJDBCSQLTypeForDBType(dataType),
-                        getJavaClass(dataType), position++);
-                allParams.put(NameNormaliser.normaliseName(columnName), dbp);
+        private DatabaseObject(DatabaseObjectName objName) {
+            name = objName;
+        }
+
+        protected DatabaseMetaData getDbMetaData() throws SQLException {
+            return getConnection().getMetaData();
+        }
+
+        abstract protected ResultSet getObjectMetaData() throws SQLException;
+
+        public boolean exists() throws SQLException {
+            try (ResultSet rs = getObjectMetaData()) {
+                return rs.next();
             }
-            rs.close();
+        }
+
+        abstract protected ResultSet getColumns() throws SQLException;
+
+        public Map<String, DbParameterAccessor> getParams() throws SQLException {
+            Map<String, DbParameterAccessor> allParams = new HashMap<>();
+            try (ResultSet rs = getColumns()) {
+                while (rs.next()) {
+                    DbParameterAccessor accessor = createColumnAccessor(rs);
+                    allParams.put(normaliseName(accessor.getName()), accessor);
+                }
+            }
             return allParams;
+        }
+
+        protected DbParameterAccessor createColumnAccessor(ResultSet rs)
+                throws SQLException {
+            String paramTypeName = rs.getString("TYPE_NAME");
+            return createDbParameterAccessor(getColumnName(rs), columnDirection(rs),
+                           typeMapper.getJDBCSQLTypeForDBType(paramTypeName),
+                           getJavaClass(paramTypeName), rs.getInt("ORDINAL_POSITION") - 1);
+        }
+
+        protected String getColumnName(ResultSet rs) throws SQLException {
+            return defaultIfNull(rs.getString("COLUMN_NAME"), "");
+        }
+
+        protected Direction columnDirection(ResultSet rs) throws SQLException {
+            return requireNonNull(getColumnDirection(rs),
+                            "Invalid type for column/parameter " + getColumnName(rs));
+        }
+
+        abstract protected Direction getColumnDirection(ResultSet rs) throws SQLException;
+    }
+
+    private class DatabaseTable extends DatabaseObject {
+
+        private DatabaseTable(DatabaseObjectName objName) {
+            super(objName);
+        }
+
+        @Override
+        protected ResultSet getObjectMetaData() throws SQLException {
+            return getDbMetaData().getTables(null, name.getSchemaName(), name.getObjectName(), null);
+        }
+
+        @Override
+        public ResultSet getColumns() throws SQLException {
+            return getDbMetaData().getColumns(null, name.getSchemaName(), name.getObjectName(), "%");
+        }
+
+        @Override
+        protected Direction getColumnDirection(ResultSet rs) {
+            return Direction.INPUT;
         }
     }
 
-    public Map<String, DbParameterAccessor> getAllProcedureParameters(
-            String procName) throws SQLException {
-        throw new UnsupportedOperationException();
+    private class DatabaseProcedure extends DatabaseObject {
+
+        private DatabaseProcedure(DatabaseObjectName objName) {
+            super(objName);
+        }
+
+        @Override
+        protected ResultSet getObjectMetaData() throws SQLException {
+            return getDbMetaData().getProcedures(null, name.getSchemaName(), name.getObjectName());
+        }
+
+        @Override
+        public ResultSet getColumns() throws SQLException {
+            return getDbMetaData().getProcedureColumns(null, name.getSchemaName(), name.getObjectName(), "%");
+        }
+
+        @Override
+        protected Direction getColumnDirection(ResultSet rs) throws SQLException {
+            switch (rs.getInt("COLUMN_TYPE")) {
+                case DatabaseMetaData.procedureColumnIn:
+                    return Direction.INPUT;
+                case DatabaseMetaData.procedureColumnInOut:
+                    return Direction.INPUT_OUTPUT;
+                case DatabaseMetaData.procedureColumnOut:
+                    return Direction.OUTPUT;
+                default:
+                    return null;
+            }
+        }
     }
 
+    private class DatabaseFunction extends DatabaseObject {
+
+        private DatabaseFunction(DatabaseObjectName objName) {
+            super(objName);
+        }
+
+        @Override
+        protected ResultSet getObjectMetaData() throws SQLException {
+            return getDbMetaData().getFunctions(null, name.getSchemaName(), name.getObjectName());
+        }
+
+        @Override
+        public ResultSet getColumns() throws SQLException {
+            return getDbMetaData().getFunctionColumns(null, name.getSchemaName(), name.getObjectName(), "%");
+        }
+
+        @Override
+        protected Direction getColumnDirection(ResultSet rs) throws SQLException {
+            switch (rs.getInt("COLUMN_TYPE")) {
+                case DatabaseMetaData.functionColumnIn:
+                    return Direction.INPUT;
+                case DatabaseMetaData.functionReturn:
+                    return Direction.RETURN_VALUE;
+                default:
+                    return null;
+            }
+        }
+    }
+
+    private static class DatabaseObjectName {
+        String schemaName, objectName;
+
+        private DatabaseObjectName(String objSchemaName, String objName) {
+            schemaName = objSchemaName;
+            objectName = objName;
+        }
+
+        private String getSchemaName() {
+            return schemaName;
+        }
+
+        String getObjectName() {
+            return objectName;
+        }
+    }
+
+    DatabaseObjectName buildDatabaseObjectName(String objName) throws SQLException {
+        String[] qualifiers = objName.toUpperCase().split("\\.");
+        String schemaName = (qualifiers.length == 2) ? qualifiers[0] : getConnection().getSchema();
+        String objectName = (qualifiers.length == 2) ? qualifiers[1] : qualifiers[0];
+        return new DatabaseObjectName(schemaName, objectName);
+    }
+
+    DatabaseObject findStoredRoutine(DatabaseObjectName objName) throws SQLException {
+        List<DatabaseObject> prioritisedCandidates = Arrays.asList(
+            new DatabaseProcedure(objName),
+            new DatabaseFunction(objName));
+        for (DatabaseObject object : prioritisedCandidates) {
+            if (object.exists()) {
+                return object;
+            }
+        }
+        throw new SQLException("Procedure/function " + objName.getObjectName() + " does not exist");
+    }
+
+    @Override
+    public Map<String, DbParameterAccessor> getAllProcedureParameters(String callableName)
+            throws SQLException {
+        return findStoredRoutine(buildDatabaseObjectName(callableName)).getParams();
+    }
+
+    @Override
     public Class<?> getJavaClass(final String dataType) {
         return typeMapper.getJavaClassForDBType(dataType);
     }
@@ -127,6 +276,7 @@ public class DerbyEnvironment extends AbstractDbEnvironment {
         private static final List<String> timeTypes = Arrays
                 .asList(new String[] { "TIME" });
 
+        @Override
         public Class<?> getJavaClassForDBType(final String dbDataType) {
             String dataType = normaliseTypeName(dbDataType);
             if (stringTypes.contains(dataType))
@@ -153,6 +303,7 @@ public class DerbyEnvironment extends AbstractDbEnvironment {
                     + "' is not supported for Derby");
         }
 
+        @Override
         public int getJDBCSQLTypeForDBType(final String dbDataType) {
             String dataType = normaliseTypeName(dbDataType);
             if (stringTypes.contains(dataType))
--- a/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/FlowMode/CallingFunctions.wiki
+++ b/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/FlowMode/CallingFunctions.wiki
@@ -1,0 +1,16 @@
+---
+Test
+---
+!3 Functions are tested like procedures, with ? used for the result
+
+|Execute procedure|Multiply|
+|n1|n2|?|
+|10|15|150|
+|2|8|16|
+
+Order of parameters in table should not be important (does not have to match db order)
+
+|Execute procedure|Multiply|
+|n2|?|n1|
+|15|150|10|
+|8|16|2|
--- a/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/FlowMode/SetUp/content.txt
+++ b/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/FlowMode/SetUp/content.txt
@@ -6,4 +6,8 @@
 
 |Execute Ddl|${CREATEMAKEUSERPROC}|
 
+|Execute Ddl|${CREATECALCLENGTHPROC}|
+
+|Execute Ddl|${CREATEMULTIPLYPROC}|
+
 |Commit|
--- a/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/FlowMode/SimpleStoredProc.wiki
+++ b/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/FlowMode/SimpleStoredProc.wiki
@@ -1,0 +1,10 @@
+---
+Test
+---
+This stored proc tests output parameters - !-CalcLength-! returns string length as second output param
+
+|Execute Procedure       |!-CalcLength-!|
+|name                    |str length?   |
+|mika                    |4             |
+|paradajz                |8             |
+|saeiluhrweurhfi7fhi7rhgf|24            |
--- a/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/FlowMode/TearDown/content.txt
+++ b/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/FlowMode/TearDown/content.txt
@@ -4,4 +4,8 @@
 
 |Execute Ddl|${DROPMAKEUSERPROC}|
 
+|Execute Ddl|${DROPCALCLENGTHPROC}|
+
+|Execute Ddl|${DROPMULTIPLYPROC}|
+
 |Commit|
--- a/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/StandaloneFixtures/SetUp/content.txt
+++ b/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/StandaloneFixtures/SetUp/content.txt
@@ -8,5 +8,9 @@
 
 |Execute Ddl|${CREATEMAKEUSERPROC}|
 
+|Execute Ddl|${CREATECALCLENGTHPROC}|
+
+|Execute Ddl|${CREATEMULTIPLYPROC}|
+
 |Database Environment|
 |Commit|
--- a/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/StandaloneFixtures/TearDown/content.txt
+++ b/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/StandaloneFixtures/TearDown/content.txt
@@ -5,6 +5,10 @@
 
 |Execute Ddl|${DROPMAKEUSERPROC}|
 
+|Execute Ddl|${DROPCALCLENGTHPROC}|
+
+|Execute Ddl|${DROPMULTIPLYPROC}|
+
 |Database Environment|
 |Commit|
 
--- a/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/content.txt
+++ b/FitNesseRoot/DbFit/AcceptanceTests/JavaTests/DerbyTests/content.txt
@@ -2,5 +2,9 @@
 !define DROPUSERTABLE {DROP TABLE USERS}
 !define CREATEMAKEUSERPROC {CREATE PROCEDURE !-MakeUser-!() LANGUAGE JAVA PARAMETER STYLE JAVA EXTERNAL NAME '!-dbfit.environment.DerbyRegressionSupportStoredProcs.MakeUser-!'}
 !define DROPMAKEUSERPROC {DROP PROCEDURE !-MakeUser-!}
+!define CREATECALCLENGTHPROC {CREATE PROCEDURE !-CalcLength-!(IN !-Name-! VARCHAR(200), OUT !-StrLength-! INTEGER) LANGUAGE JAVA PARAMETER STYLE JAVA EXTERNAL NAME '!-dbfit.environment.DerbyRegressionSupportStoredProcs.calcLength-!'}
+!define DROPCALCLENGTHPROC {DROP PROCEDURE !-CalcLength-!}
+!define CREATEMULTIPLYPROC {CREATE FUNCTION !-Multiply-!(n1 DOUBLE, n2 DOUBLE) RETURNS DOUBLE LANGUAGE JAVA PARAMETER STYLE JAVA EXTERNAL NAME '!-dbfit.environment.DerbyRegressionSupportStoredProcs.multiply-!'}
+!define DROPMULTIPLYPROC {DROP FUNCTION !-Multiply-!}
 
 !contents -R
--- a/dbfit-java/derby/src/integration-test/java/dbfit/environment/DerbyRegressionSupportStoredProcs.java
+++ b/dbfit-java/derby/src/integration-test/java/dbfit/environment/DerbyRegressionSupportStoredProcs.java
@@ -18,4 +18,12 @@ public class DerbyRegressionSupportStoredProcs {
             stmt.execute(sql);
         }
     }
+
+    public static void calcLength(String name, int[] strLength) throws SQLException {
+        strLength[0] = name.length();
+    }
+
+    public static double multiply(double n1, double n2) throws SQLException {
+        return n1 * n2;
+    }
 }


dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +53,176 @@ public Pattern getParameterPattern() {
...
 
-    public Map<String, DbParameterAccessor> getAllColumns(
-            final String tableOrViewName) throws SQLException {
-        String qry = "SELECT COLUMNNAME, COLUMNDATATYPE "
-                + "FROM SYS.SYSCOLUMNS WHERE REFERENCEID = "
-                + "(SELECT TABLEID FROM SYS.SYSTABLES WHERE TABLENAME = ?)";
-        return readIntoParams(tableOrViewName.toUpperCase(), qry);
+    public Map<String, DbParameterAccessor> getAllColumns(String tableName)
+                                                        throws SQLException {
+        DatabaseTable t = new DatabaseTable(getConnection(), getConnection().getSchema(), tableName.toUpperCase());
```
javornikolov: That temp variable is only used at next line so possibly we can skip it and call `.getParams()` here directly.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +53,176 @@ public Pattern getParameterPattern() {
...
+
+    private class DatabaseFunction extends DatabaseObject {
+
+        private DatabaseFunction(Connection conn, String defaultSchemaName, String objName) {
+            super(conn, defaultSchemaName, objName);
+        }
+
+        @Override
+        public boolean exists() throws SQLException {
+            ResultSet rs = con.getMetaData().getFunctions(null, name.getSchemaName(), name.getObjectName());
```
javornikolov: The cursor is not being closed here. Perhaps the helper method from the other PR can be reused here.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +53,176 @@ public Pattern getParameterPattern() {
...
+        }
+
+        @Override
+        public ResultSet getColumns() throws SQLException {
+            return con.getMetaData().getFunctionColumns(null, name.getSchemaName(), name.getObjectName(), "%");
+        }
+
+        @Override
+        protected Direction getParamDirection(ResultSet rs)
+                                        throws SQLException {
```
javornikolov: I think the method name and throws clause is short enough to fit in single line here.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +53,176 @@ public Pattern getParameterPattern() {
...
+
+    private class DatabaseProcedure extends DatabaseObject {
+
+        private DatabaseProcedure(Connection conn, String defaultSchemaName, String objName) {
+            super(conn, defaultSchemaName, objName);
+        }
+
+        @Override
+        public boolean exists() throws SQLException {
+            ResultSet rs = con.getMetaData().getProcedures(null, name.getSchemaName(), name.getObjectName());
```
javornikolov: ResultSet remains unclosed.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +53,176 @@ public Pattern getParameterPattern() {
...
+                                                                       throws SQLException {
+        DatabaseObject o;
+        String name = callableName.toUpperCase();
+        DatabaseProcedure p = new DatabaseProcedure(getConnection(), getConnection().getSchema(), name);
+        if (p.exists()) {
+            o = p;
+        } else {
+            DatabaseFunction f = new DatabaseFunction(getConnection(), getConnection().getSchema(), name);
+            if (f.exists()) {
+                o = f;
```
javornikolov: Here we go again in multiple levels of nesting and some single-letter variable names. The `if/else if/else` in the other PR looked a bit easier to read.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +53,176 @@ public Pattern getParameterPattern() {
...
+        }
+
+        @Override
+        protected Direction getParamDirection(ResultSet rs)
+                                        throws SQLException {
+            Direction dir = null;
+            switch (rs.getInt("COLUMN_TYPE")) {
+                case DatabaseMetaData.procedureColumnIn:
+                    dir = Direction.INPUT;
+                    break;
```
javornikolov: What about shortcut `return` to avoid the multiple `break` statements and the temp variable?

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +53,176 @@ public Pattern getParameterPattern() {
...
+        DatabaseTable t = new DatabaseTable(getConnection(), getConnection().getSchema(), tableName.toUpperCase());
+        return t.getParams();
     }
 
-    private Map<String, DbParameterAccessor> readIntoParams(
-            String tableOrViewName, String query) throws SQLException {
-        try (PreparedStatement dc = getConnection().prepareStatement(query)) {
-            dc.setString(1, tableOrViewName);
+    private abstract class DatabaseObject {
+        protected Connection con;
```
javornikolov: Hmm, we have many names and abbreviations for the connection (con, conn, ...).

Technically here we have a direct access here to the parent `getConnection()` of DerbyEnvironment (since it's a non-static inner class of it).

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +53,176 @@ public Pattern getParameterPattern() {
...
+        protected DbObjectName name;
+        Map<String, DbParameterAccessor> allParams = new HashMap<String, DbParameterAccessor>();
+
+        private DatabaseObject(Connection conn, String defaultSchemaName, String objName) {
+            con = conn;
+            name = new DbObjectName(defaultSchemaName, objName);
```
javornikolov: That `DbObjectName` can come directly from the constructor.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +53,176 @@ public Pattern getParameterPattern() {
...
 
-            ResultSet rs = dc.executeQuery();
-            Map<String, DbParameterAccessor> allParams = new HashMap<String, DbParameterAccessor>();
-            int position = 0;
+        abstract protected boolean exists() throws SQLException;
+
+        abstract protected ResultSet getColumns() throws SQLException;
+
+        public Map<String, DbParameterAccessor> getParams() throws SQLException {
+            ResultSet rs = getColumns();
```
javornikolov: This rs is never closed.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: I can get at that method with `DerbyEnvironment.this.getConnection()` but then end up with long expressions.
E.g.
`DerbyEnvironment.this.getConnection().getMetaData().getFunctionColumns(null, name.getSchemaName(), name.getObjectName(), "%");`

What preferred practice here? Introduce another variable to reference the result of ``DerbyEnvironment.this.getConnection()` or just break up the expression over multiple lines. E.g.

```
ResultSet rs = DerbyEnvironment.this.getConnection().getMetaData()
                   .getFunctionColumns(null, name.getSchemaName(),
                   name.getObjectName(), "%");
```

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: Yes. Good point.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: I will need a `default` option with a `return` in the `switch` to ensure compilation. I.e. make the `switch` statement be: -

```
            switch (rs.getInt("COLUMN_TYPE")) {
                case DatabaseMetaData.procedureColumnIn:
                    return Direction.INPUT;
                case DatabaseMetaData.procedureColumnInOut:
                    return Direction.INPUT_OUTPUT;
                default:
                    return Direction.OUTPUT;
            }
```

javornikolov: It can be either return in the default option or it can be return statement after the switch (both compile fine).

But good that you turned attention into that: right now it returns `null` in case of unknown type. Shouldn't it be throwing an exception?

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: It's not exactly the same case as I have more than one statement at some of the branching points.

I'll see how I can squash it a bit.

MMatten:
I've made an attempt at addressing all of these points and have force pushed the revisions.

MMatten:
Ah. I didn't see you latest comments until I'd pushed revisions. I'll go through those now.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: > Shouldn't it be throwing an exception?

Yes. Theoretically it can't happen but I'll raise a `SQLException` for the `default` option.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +52,193 @@ public Pattern getParameterPattern() {
...
+                case DatabaseMetaData.functionReturn:
+                    return Direction.RETURN_VALUE;
+                default:
+                    throw new SQLException("Unpexpected function parameter type for parameter " +
+                                  rs.getString("COLUMN_NAME"));
+            }
+        }
+    }
+
+    private class DbObjectName {
```
MMatten: May be this should be `DatabaseObjectName` to align with the other new inner classes...?

MMatten:
I've applied your ideas. Let me know what you think.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
javornikolov: Yes, sounds like good idea.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +52,193 @@ public Pattern getParameterPattern() {
...
+        }
+
+        @Override
+        protected ResultSet getObjects() throws SQLException {
+            return getMetaData().getProcedures(null, name.getSchemaName(), name.getObjectName());
+        }
+
+        @Override
+        public ResultSet getColumns() throws SQLException {
+            ResultSet rs = getMetaData().getProcedureColumns(null, name.getSchemaName(), name.getObjectName(), "%");
```
javornikolov: Do we need the rs variable?

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +52,193 @@ public Pattern getParameterPattern() {
...
+
+    private class DatabaseFunction extends DatabaseObject {
+
+        private DatabaseFunction(DbObjectName objName) {
+            super(objName);
+        }
+
+        @Override
+        public boolean exists() throws SQLException {
+            try (ResultSet rs = getObjects()) {
```
javornikolov: Duplicated code (it's the same in all child classes).

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: Update applied.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +52,193 @@ public Pattern getParameterPattern() {
...
+
+        private DatabaseObject(DatabaseObjectName objName) {
+            name = objName;
+        }
+
+        protected DatabaseMetaData getMetaData() throws SQLException {
+            return getConnection().getMetaData();
+        }
+
+        abstract protected ResultSet getObjects() throws SQLException;
```
javornikolov: The name is a bit confusing: since in general we intend to represent a single database object.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +52,193 @@ public Pattern getParameterPattern() {
...
+            switch (rs.getInt("COLUMN_TYPE")) {
+                case DatabaseMetaData.procedureColumnIn:
+                    return Direction.INPUT;
+                case DatabaseMetaData.procedureColumnInOut:
+                    return Direction.INPUT_OUTPUT;
+                case DatabaseMetaData.procedureColumnOut:
+                    return Direction.OUTPUT;
+                default:
+                    throw new SQLException("Unpexpected procedure parameter type for parameter " +
+                                  rs.getString("COLUMN_NAME"));
```
javornikolov: What indentation conventions are being used here?

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +52,193 @@ public Pattern getParameterPattern() {
...
+            return schemaName;
+        }
+
+        String getObjectName() {
+            return objectName;
+        }
+    }
+
+    public Map<String, DbParameterAccessor> getAllProcedureParameters(String callableName)
+                                                                       throws SQLException {
```
javornikolov: What is the rule of indentation here?

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +52,193 @@ public Pattern getParameterPattern() {
...
+        String name = callableName.toUpperCase();
+        DatabaseObjectName objName = new DatabaseObjectName(getConnection().getSchema(), name);
+        DatabaseProcedure dbproc = new DatabaseProcedure(objName);
+        if (dbproc.exists()) {
+            return dbproc.getParams();
+        } else {
+            DatabaseFunction dbfunc = new DatabaseFunction(objName);
```
javornikolov: Hmm, that still looks a bit complicated if compared to the other. And indeed with this design of `exists` it's not quite the same as it used to be in previous PR. What about having one more helper method to get the object. E.g. something like:
``` java
DatabaseObject getStoredCall(objName) {
    List<DatabaseObject> prioritisedCandidates = Arrays.asList(
        new DatabaseProcedure(objName),
        new DatabaseFunction(objName));

    for (DatabaseObject object : prioritisedCandidaes)
        if (object.exists()) {
            return object;
        }
    }

    throw new SQLException("Procedure/function " + objName + " does not exist");
}
```

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: Agreed. I didn't really like it that much. I'm now thinking of having `getObjectMetaData()` and `getDBMetaData()`.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: Oops. Thanks.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: It's a bit random. The approach was merely to align the opening brace `{` with where it would be if the method had no `throws` clause.

What approach do you take?

javornikolov: My approach for throws formatting is also not very streamlined. Usually I try to keep all indentations at position divisible by 4 spaces (or whatever the unit is). Things I might do for that case:

``` java
public Map<String, DbParameterAccessor> getAllProcedureParameters(String callableName)
        throws SQLException {
    // 2 indentation (just a single 1 would align with the method code and won't stand out well)
```

or 
``` java
        String callableName) throws SQLException {

    // with multiple params
            String callableName,
            String param2,
            String param3) throws SQLException, SQLException2 {
        // code
    }
```

At some point I used to do something like following but I think now I would rather take the approach in the previous sample.
``` java
public Map<String, DbParameterAccessor> getAllProcedureParameters(String callableName) throws SQLException {
```

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: In fact `getDBMetaData()`, from the connection, and `getMetaData()` for the object.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: That's nice. Thanks. I'll copy it in.

MMatten: Would you usually include the blank lines in this method?

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +52,177 @@ public Pattern getParameterPattern() {
...
-    public Map<String, DbParameterAccessor> getAllColumns(
-            final String tableOrViewName) throws SQLException {
-        String qry = "SELECT COLUMNNAME, COLUMNDATATYPE "
-                + "FROM SYS.SYSCOLUMNS WHERE REFERENCEID = "
-                + "(SELECT TABLEID FROM SYS.SYSTABLES WHERE TABLENAME = ?)";
-        return readIntoParams(tableOrViewName.toUpperCase(), qry);
+    public Map<String, DbParameterAccessor> getAllColumns(String tableName)
+            throws SQLException {
+        return new DatabaseTable(new DatabaseObjectName(getConnection().getSchema(),
+                                 tableName.toUpperCase())).getParams();
```
javornikolov: Here the the two parameters of `DatabaseObjectName` are not quite aligned. I would approach that maybe like that:
``` java
        return new DatabaseTable(new DatabaseObjectName(
            getConnection().getSchema(), tableName.toUpperCase())).getParams();
// or
        return new DatabaseTable(new DatabaseObjectName(
            getConnection().getSchema(),
            tableName.toUpperCase())).getParams();
```

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: Again, nothing very scientific here. I can change it to align with the opening parenthesis. I.e.:

```
                default:
                    throw new SQLException("Unpexpected procedure parameter type for parameter " +
                                           rs.getString("COLUMN_NAME"));
```

MMatten:
More polishing applied.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +52,177 @@ public Pattern getParameterPattern() {
...
-                allParams.put(NameNormaliser.normaliseName(columnName), dbp);
+    private abstract class DatabaseObject {
+        protected DatabaseObjectName name;
+        Map<String, DbParameterAccessor> allParams = new HashMap<String, DbParameterAccessor>();
+
+        private DatabaseObject(DatabaseObjectName objName) {
+            name = objName;
+        }
+
+        protected DatabaseMetaData getDBMetaData() throws SQLException {
```
javornikolov: Should it be `DB` or `Db`? I think we have of both around, maybe `Db` is more common (like `DbParameterAccessor`).

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +52,177 @@ public Pattern getParameterPattern() {
...
+        }
+
+        @Override
+        public ResultSet getColumns() throws SQLException {
+            return getDBMetaData().getFunctionColumns(null, name.getSchemaName(), name.getObjectName(), "%");
+        }
+
+        @Override
+        protected Direction getParamDirection(ResultSet rs) throws SQLException {
+            int paramType = rs.getInt("COLUMN_TYPE");
```
javornikolov: In `DatabaseProcedure` we don't have that local variable.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +52,177 @@ public Pattern getParameterPattern() {
...
+            switch (paramType) {
+                case DatabaseMetaData.functionColumnIn:
+                    return Direction.INPUT;
+                case DatabaseMetaData.functionReturn:
+                    return Direction.RETURN_VALUE;
+                default:
+                    throw new SQLException("Unpexpected function parameter type for parameter " +
+                                  rs.getString("COLUMN_NAME"));
```
javornikolov: That's aligned a bit differently in `DatabaseProcedure`.

I'm wondering about how to avoid the small bits of duplication here - the throw exception is the same, the get.. column type is the same. Also there is a similar pattern (switch statement).

I think that there are some options to tackle that too. What do you think?

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +52,177 @@ public Pattern getParameterPattern() {
...
+        throw new SQLException("Procedure/function " + objName.getObjectName() + " does not exist");
     }
 
-    public Map<String, DbParameterAccessor> getAllProcedureParameters(
-            String procName) throws SQLException {
-        throw new UnsupportedOperationException();
+    public Map<String, DbParameterAccessor> getAllProcedureParameters(String callableName)
+            throws SQLException {
+        String name = callableName.toUpperCase();
+        return findStoredRoutine(new DatabaseObjectName(getConnection().getSchema(), name)).getParams();
```
javornikolov: That `new DatabaseObjectName(getConnection().getSchema(), somename.toUpperCase())` is a bit long and we have in one more place. So could we fine tune that too :-)

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
javornikolov: Yes, I would usually include the blank lines. Maybe the 1st one could be omitted since the above statement is multi-line with some indentation with 2nd, 3rd lines. But if the `List...` was all on single line - the `for` loop block wouldn't stand out very well without an empty line in between.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: Ok. It's subjective and tricky stuff. Do you know of anything in `checkstyle` that we can use to pick a standard?

I'll choose one of your options for now.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: Good spot.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: As parameter direction is not really appropriate for tables I was originally thinking of having another level of abstraction, possible `DatabaseExecutable` or `DatabaseCallable` and pushing as much as I could into a method in that class. I still can't quite put my finger on how to have a single place to throw the exception if the param type from the `get...Columns` is invalid.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
javornikolov: I'm not sure if `checkstyle` has something good enough.

There are some related notes in the style guide: http://www.oracle.com/technetwork/java/javase/documentation/codeconventions-136091.html#262

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
javornikolov: Could be some map (int -> direction) like we have with type mapping in db environments.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: Yes. I can't think of a nice short name. How's this one: -
```
    DatabaseObjectName createDbObjectNameWithCurrentDefaultSchema(String objName) throws SQLException {
        return new DatabaseObjectName(getConnection().getSchema(), objName);
    }
```

javornikolov: Maybe it could be just `parseDatabaseObjectName` or even `parseObjectName` for now. Can be renamed some day if some day we have more options and risk of mis-interpretation. Possibly objeName.toUpperCase() can be injected here (to be honest I'm not sure about case sensitivity in Derby but if uppercase of name is needed, I guess it would be fine for schema name too).

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: I'm thinking of using something like guava's `ImmutableMap.of()` or `ImmutableMap.builder` to initialise a `static Map` from a list of literals. Otherwise with `HashMap` it has to be done in a method (perhaps the `DatabaseProcedure/Function` constructor).

Then we have a method in the new `DatabaseExecutable` (or `DatabaseRoutine`?) that gets the `Direction` value from the JDBC parameter type `int`. If it's not in the map then throw the exception.

It still sounds like I have something like 
```
    if (getParamDirection(jdbcIntValue) == null) {
        throw new SQLException("Invalid parameter type");
    }
    return (getParamDirection(jdbcIntValue);
````

Comment (javornikolov) [inline dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java]: > to initialise a static Map from a list of literals

Why should it be a `static` one? I'm not sure the performance overhead would be so terrible even with non-static one.

> Otherwise with HashMap it has to be done in a method

That can even be in the `getXXX` method.

> It still sounds like I have something like
> 
>     if (getParamDirection(jdbcIntValue) == null) {
>         throw new SQLException("Invalid parameter type");
>     }
>     return (getParamDirection(jdbcIntValue);

I think yes. Maybe with a variable to avoid calling the getXXx twice.

---
Maybe member variable (static or not) would makes some sense. Just I don't see that yet, good to start with a more obvious implementation.

Comment (javornikolov) [inline dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java]: BTW, initialization logic of static members can be done in static initialiser like:
``` java
static {
    map.put(key, value);
    ...
}
```

Or to do an assignment by call to static method.

But in general static members are not supported in non-static inner classes. If desired in our case - this could rather be done in the parent class.

javornikolov: Initializer blocks may be actually used in non-static case too:

``` java
    abstract class Base {
        protected final Map<String, String> map = new HashMap<>();
    }

    class Child1 extends Base {
        {
            map.put("0", "value1");
            map.put("1", "value2");
            map.put("2", "value3");
        }
    }
```

MMatten:
I draft another approach to get rid of the dummy method returning `Direction.INPUT` for `DatabaseTable` types. Any thoughts on this approach?

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +52,199 @@ public Pattern getParameterPattern() {
...
+
+    abstract private class DatabaseExecutable extends DatabaseObject {
+
+        private DatabaseExecutable(DatabaseObjectName objName) {
+            super(objName);
+        }
+
+        abstract protected Direction columnTypeToDir(int columnType);
+
+        protected Direction getParamDirection(ResultSet rs) throws SQLException {
```
javornikolov: An alternative idea about that mapping: `Direction getParamDirection(ResultSet rs)` can be abstracted in separate interface and in child classes we can just `plug` different implementations of it (e.g. via initialization of common member via the constructor or in some other way).

Maybe that doesn't make sense in our particular case here. I'm just sharing a thought which came to me while wondering about alternatives without a deep 3-level hierarchy (substituting inheritance with composition in this case).

javornikolov:
The compilation error should be possible to handle via `? extends T` (a bit subtle caveat of Java generics).

javornikolov:
> The compilation error should be possible to handle via ? extends T (a bit subtle caveat of Java generics).

Or change the type in the list.
Interesting enough: looks like we're not using `exists` for `DatabaseTable`... Maybe it's fine that way, I'm not sure what happens if we reach to getting table columns of non-existent table?

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +52,199 @@ public Pattern getParameterPattern() {
...
+        }
+    }
+
+    abstract private class DatabaseExecutable extends DatabaseObject {
+
+        private DatabaseExecutable(DatabaseObjectName objName) {
+            super(objName);
+        }
+
+        abstract protected Direction columnTypeToDir(int columnType);
```
MMatten: You're right. I'll expand that.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +52,199 @@ public Pattern getParameterPattern() {
...
+            Direction paramDir = columnTypeToDir(rs.getInt("COLUMN_TYPE"));
+            if (paramDir == null) {
+                throw new SQLException("Invalid parameter type");
+            }
+            return paramDir;
+        }
+
+        protected DbParameterAccessor createColumnAccessor(String paramName, ResultSet rs)
+                throws SQLException {
+            return createDbParamAccessor(paramName, rs, getParamDirection(rs));
```
MMatten: We did have that earlier. It still didn't feel right having this method that didn't use any of it's inputs/params. What do you think about dummy methods like this?

javornikolov: I think it's OK in that case.

MMatten: The notion of `Direction` seems contrived and a bit ugly for tables. I can only think DbFit was originally developed with stored procs in mind and tables/views/queries were added later.

javornikolov: > Do you mean you think it's OK to have a dummy one? Or it OK to remove it (as it is right now)?

In general it's OK to have unused parameter in some child classes: since we try to abstract things in a more general method.


> The notion of Direction seems contrived and a bit ugly for tables. I can only think DbFit was originally developed with stored procs in mind and tables/views/queries were added later.

That's a a good point. Basically the direction makes sense for SQL queries (input and output parameters). It's possible e.g. for insert statements to have output parameters for auto-generated values (by triggers, auto-increments, etc.).
The thing is that we're using `DbParameterAccessor` both for describing database meta-data (parameters or columns), and also for binding to SQL parameters. It might make sense to split.

Anyway - it's broader topic. Maybe a separate issue can be created to think about such re-design.

MMatten:
> > The compilation error should be possible to handle via ? extends T (a bit subtle caveat of Java generics).

> Or change the type in the list.

It only fails to compile with JDK 7, but with 8 it's fine.

I've update the `List` to be based upon `DatabaseExecutable`.

MMatten:
> Interesting enough: looks like we're not using exists for DatabaseTable... Maybe it's fine that way, I'm not sure what happens if we reach to getting table columns of non-existent table?

I've not yet changed the behaviour of getting table column accessors. Currently it returns an empty `Map` and the behaviour is consistent with how the other adaptors work for tables and stored code.

We can instead make it call the `exists()` for `DatabaseTable`s and throw an exception instead. What do you think?

javornikolov:
> We can instead make it call the exists() for DatabaseTables and throw an exception instead. What do you think?

Let's keep that behavior as is. If we change it we should rather change all other adapters to keep them consistent.

Since we're not using `DatabaseTable.exists()` - I guess it makes sense to drop that code and make it relevant for stored routines only (for now).

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: Perhaps we can rework `DbParameterAccessor` into something like `DbObjectColumnAccessor`, `DbTableColumnAccessor` (no direction) and `DbExecutableColumnAccessor` (has direction) at some future point.

MMatten:
> Since we're not using DatabaseTable.exists() - I guess it makes sense to drop that code and make it relevant for stored routines only (for now).

And revert behaviour for procs/funcs too?

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
javornikolov: In my understanding **all** accessors have a Direction (all JDBC statements are like that) - these are accessors to parameters of SQL statements. Unless we split them into separate Input and Output accessors which is also possible but a separate topic.

javornikolov:
>> Since we're not using DatabaseTable.exists() - I guess it makes sense to drop that code and make it relevant for stored routines only (for now).

> And revert behaviour for procs/funcs too?

Could be just moving the abstract exists to `*Executable`.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: Well, `getColumns` does not return a `COLUMN_TYPE` column so table columns don't have direction according to JDBC.

MMatten:
Tidied up (apart from commits still to squash).

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +52,199 @@ public Pattern getParameterPattern() {
...
+            return schemaName;
+        }
+
+        String getObjectName() {
+            return objectName;
+        }
+    }
+
+    DatabaseObjectName createDbObjectNameWithCurrentDefaultSchema(String objName) throws SQLException {
+        return new DatabaseObjectName(getConnection().getSchema(), objName);
```
MMatten: Ah, didn't address these yet...

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
javornikolov: I meant direction in terms of [PreparedStatement](https://docs.oracle.com/javase/7/docs/api/java/sql/PreparedStatement.html) parameters. I agree that columns and stored proc parameters have different properties, I'm trying to make the point that none of them are database *statements* at all.

MMatten: Right, yes. I see what you mean.

javornikolov:
> Could be just moving the abstract exists to *Executable.

Or it could be some default implementation in the base class throwing 'not implemented yet'. And maybe it's possible to get rid of DatabaseExecutable (if nothing specific remains for it). After all - having exists() in DatabaseTable is not so bad (though leading to some confusion where it's used).

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: Object names need to be upper case when using the JDBC `get...` methods. I guess Derby translate names to upper case when objects are created.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: > I meant direction in terms of PreparedStatement parameters.

We're using `DbParameterAccessor` for non `PreparedStatement`/`CallableStatement` too. E.g. for query results and we use a dummy `INPUT` `Direction` for these may be we could clean this up some time.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: > What about moving the parsing logic from DatabaseObjectName c-tor to here?

What would that do for us?

What about adding a `parse` method to the `DatabaseObjectName` class although it can't be `static`?

MMatten:
> And maybe it's possible to get rid of DatabaseExecutable (if nothing specific remains for it). After all - having exists() in DatabaseTable is not so bad (though leading to some confusion where it's used).

I think I'll still have some common functionality at this level. Let me push the latest version to get a view of what it all looks like.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
javornikolov: >> What about moving the parsing logic from DatabaseObjectName c-tor to here?
>
> What would that do for us?

* Would allow more expressive name for the `DatabaseObjectName` construction method.
* There is a general rant about factory methods vs constructors, and avoiding complex logic in constructors. As we already have a factory method (create/build ... with current default schema): it looks logical to keep non-core construction details outside the c-tor.
* It's interesting to have a look at [Kotlin Data Classes](https://kotlinlang.org/docs/reference/data-classes.html). Unfortunately not available in Java language itself (maybe some day).

> What about adding a parse method to the `DatabaseObjectName` class although it can't be static?

The issue is we can't use non-static method before construction... But it's possible to make `DatabaseObjectName` class static. Or the parsing method can be outside the class (as we now have buildDatabaseObjectName).

MMatten: > Would allow more expressive name for the `DatabaseObjectName` construction method.

That makes sense.

> There is a general rant about factory methods vs constructors, and avoiding complex logic in constructors.

This is fairly subjective. All we're currently doing is splitting a string and substituting a default. Do you mean that *data*-like classes should just really have a constructor that assigns values to internal fields and *getters* to retrieve them?

> It's interesting to have a look at Kotlin Data Classes.

As the the comment above.

MMatten: > The issue is we can't use non-static method before construction... But it's possible to make DatabaseObjectName class static.

We have no reliance upon a parent class object instance at the moment so that's also possible.

javornikolov: > Do you mean that data-like classes should just really have a constructor that assigns values to internal fields and getters to retrieve them?

Yes, if this is simpler, trivial and obvious representation which does the job.

Other points might be: 

* In sense of simplicity - avoid mixing multiple responsibilities.
* Consistency with other common API (e.g. [QName](https://docs.oracle.com/javase/8/docs/api/javax/xml/namespace/QName.html), [Integer](https://docs.oracle.com/javase/8/docs/api/java/lang/Integer.html), [LocalDateTime](https://docs.oracle.com/javase/8/docs/api/java/time/LocalDateTime.html) - even though these have other additional stronger reasons to that approach)


Indeed, it's fairly subjective matter and the parsing is simple. It's hard to make a strong case (e.g. like in http://misko.hevery.com/code-reviewers-guide/flaw-constructor-does-real-work/).

What matters now is the current context so if things look better as they are - no problem to be that way.

MMatten: It makes sense. 

My only doubt is that having this common parsing outside of the class seems to inhibit potential reuse of the `DatabaseObjectName` class. If other classes were to instantiate it they'd need the same parsing method (duplicated) or I guess moved into a factory class or some other initialiser method. What would you tend to do in this case?

javornikolov: Yes, some kind of factory/builder/parser class or method comes to my mind too. Parsing methods or c-tors inside the same class itself can't solve such problem. Or it could be a static method with a different name.

It's something to look into when we extend the usage to another place.

MMatten: Thanks. I've pushed some updates for the above discussion.

MMatten:
Ooops. Something doesn't look right here. Let me push again.

MMatten:
Apologies. I've managed to make a mess of this last push.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +52,203 @@ public Pattern getParameterPattern() {
...
-                        getJavaClass(dataType), position++);
-                allParams.put(NameNormaliser.normaliseName(columnName), dbp);
+    public Map<String, DbParameterAccessor> getAllColumns(String tableName)
+            throws SQLException {
+        return new DatabaseTable(buildDatabaseObjectName(tableName)).getParams();
+    }
+
+    private abstract class DatabaseObject {
+        protected DatabaseObjectName name;
+        Map<String, DbParameterAccessor> allParams = new HashMap<String, DbParameterAccessor>();
```
MMatten: Now addressed. I've pushed again.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +52,206 @@ public Pattern getParameterPattern() {
...
+        }
+    }
+
+    abstract private class DatabaseExecutable extends DatabaseObject {
+
+        private DatabaseExecutable(DatabaseObjectName objName) {
+            super(objName);
+        }
+
+        protected DbParameterAccessor createColumnAccessor(String paramName, ResultSet rs)
```
javornikolov: The implementation looks the same in the parent.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +52,206 @@ public Pattern getParameterPattern() {
...
+        protected ResultSet getObjectMetaData() throws SQLException {
+            return getDbMetaData().getTables(null, name.getSchemaName(), name.getObjectName(), null);
+        }
+
+        @Override
+        public ResultSet getColumns() throws SQLException {
+            return getDbMetaData().getColumns(null, name.getSchemaName(), name.getObjectName(), "%");
+        }
+
+        protected Direction columnTypeToDirection(ResultSet rs) {
```
javornikolov: Missing `@Override`

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,40 +52,206 @@ public Pattern getParameterPattern() {
...
+        protected DbParameterAccessor createColumnAccessor(String paramName, ResultSet rs)
+                throws SQLException {
+            return createDbParamAccessor(paramName, rs, getParamDirection(rs));
+        }
+
+        protected Direction columnTypeToDirection(ResultSet rs) throws SQLException {
+            return jdbcTypeToDirection(rs.getInt("COLUMN_TYPE"));
+        }
+
+        abstract protected Direction jdbcTypeToDirection(int jdbcType);
```
javornikolov: Would it work if we use the same `columnTypeToDirection` name here? Also I'm not quite sure about the name of `jdbcType`.

javornikolov: > Would it work if we use the same columnTypeToDirection name here

Or even something like `getColumnDirection` for both.

MMatten: Would it work if we use the same columnTypeToDirection name here?

Yes. Does overloading internal method names make it harder to follow though?

I'm about to push a new version.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,42 +52,208 @@ public Pattern getParameterPattern() {
...
+        private DatabaseExecutable(DatabaseObjectName objName) {
+            super(objName);
+        }
+
+        @Override
+        protected Direction getColumnDirection(ResultSet rs) throws SQLException {
+            return columnTypeToDirection(rs.getInt("COLUMN_TYPE"));
+        }
+
+        abstract protected Direction columnTypeToDirection(int columnType);
```
javornikolov: Looking at the additional complications to avoid the duplication of `rs.getInt("COLUMN_TYPE")` I'm wondering didn't we go too far for that simple case.

Maybe at the cost of little duplication we could take a step back and move that down to the children: `switch(rs.getInt("COLUMN_TYPE"))`. Or alternatively declare a `getColumnTypeId(ResultSet)` inside `DatabaseObject` (despite that we don't use it for tables). Maybe it's a good tradeoff since this way we can keep the class hierarchy 2-levels only and perhaps drop the `int->Direction` mapping function.

In any case the current version looks quite well.
What do you think?

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,42 +52,208 @@ public Pattern getParameterPattern() {
...
+
+        protected DbParameterAccessor createColumnAccessor(String paramName, ResultSet rs)
+                throws SQLException {
+            return createDbParamAccessor(paramName, rs, columnDirection(rs));
+        }
+
+        protected Direction columnDirection(ResultSet rs) throws SQLException {
+            Direction columnDir = getColumnDirection(rs);
+            if (columnDir == null) {
+                throw new SQLException("Invalid column/parameter type");
```
MMatten: I can pass in the name to this method too.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: >>> Would it work if we use the same columnTypeToDirection name here?

>> Yes. Does overloading internal method names make it harder to follow though?

> If they're doing the same (just on different format of parameters): it seems fine if name is the same.

OK. I'll align them.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
javornikolov: > I can pass in the name to this method too.

If you consider passing the name: introducing `getColumnName(rs)` is also an option (to avoid carrying around too many parameters).

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: > Maybe at the cost of little duplication we could take a step back and move that down to the children: `switch(rs.getInt("COLUMN_TYPE"))`. Or alternatively declare a `getColumnTypeId(ResultSet)` inside `DatabaseObject` (despite that we don't use it for tables). Maybe it's a good tradeoff since this way we can keep the class hierarchy 2-levels only and perhaps drop the `int->Direction` mapping function.

I'm not sure about it. I think my own preferred version was actually where we only had any methods relating to `Direction` in `DatabaseExecutable` and below.

MMatten: This trade off consideration is of course another subjective area. 2 levels is more obscure (irrelevant concepts for some cases) whereas 3 levels is more complex but is more expressive. I think this is something where you are probably more experienced than me (certainly in the Java space) so I don't mind which way we go.

javornikolov: No strong opinion on that, either way looks OK to me.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: > If you consider passing the name: introducing getColumnName(rs) is also an option (to avoid carrying around too many parameters).

An interesting point with this is that our use of `defaultIfNull` is only really relevant for function return values. It wouldn't be obvious if we just define it in `DatabseObject`. May be we should have a default in `DatabaseObject` and override it in `DatabaseFunction`. What do you think about obviousness vs conciseness?

javornikolov: Good point indeed. Perhaps I would keep `defaultIfNull` in `DatabaseObject` only. We replace nulls in order to avoid complications when working with the map keys. There are other attributes which clearly distinguish the return values.

javornikolov: I mean, the `null` name specific doesn't seem to be something so important in order to need to be emphasized. It's just kind of type conversion to a `not null value`.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: Maybe the new class names do not help to clarify what their purpose is. Maybe they should be more aligned with the types of statements that we're actually executing: -
* DML targeting tables (CUD, not R).
* Executing stored code.

Would that make names like: -
* `StatementTargetObject`
  * `StoredCodeExecutionStatementTargetObject`
  * `DataModificationStatementTargetObject`

more valuable even though these are ridiculously long?

javornikolov: The new classes are details supporting implementation of:
`getAllProcedureParameters()`,
`getAllColumns()`.

The result is metadata about object (table/view) columns or stored procedure parameters. We don't want to couple too much to assumptions about specific uses of these metadata. (It may be DML statement, but it may be read-only metadata inspection, or supporting Query execution).

MMatten: Yes. It's just the `Direction` that's been confusing me. For queries and inspections I don't think we use `DbParameterAccessor`. So for tables it's always DML (not SELECT). 

So `Direction` for tables is for a very specific purpose.

javornikolov: Maybe we should change the types returned as database metadata in future, and convert/adapt them to accessors as separate step when needed.

Yes, for tables *currently* it' seems DML only. Hence the direction is by default `INPUT`. Sometimes we have additional manipulations on top of the returned lists (splitting in/out parameters as in+out, cloning for the purpose of auto-generated keys, etc.).

MMatten:
>> Looking at the additional complications to avoid the duplication of rs.getInt("COLUMN_TYPE") I'm wondering didn't we go too far for that simple case.

> Maybe at the cost of little duplication we could take a step back and move that down to the children: switch(rs.getInt("COLUMN_TYPE")).

I'll apply this and push again.

MMatten:
I think we can reasonably merge `createDbParamAccessor` into `createColumnAccessor`. What do you think?

javornikolov:
> I think we can reasonably merge createDbParamAccessor into createColumnAccessor. What do you think?

Maybe not a bad idea. Possibly also param name can be extracted from within.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,42 +52,203 @@ public Pattern getParameterPattern() {
...
-                String dataType = rs.getString(2);
-                DbParameterAccessor dbp = createDbParameterAccessor(
-                        columnName,
-                        Direction.INPUT,
-                        typeMapper.getJDBCSQLTypeForDBType(dataType),
-                        getJavaClass(dataType), position++);
-                allParams.put(NameNormaliser.normaliseName(columnName), dbp);
+            try (ResultSet rs = getColumns()) {
+                while (rs.next()) {
+                    String paramName = getColumnName(rs);
```
javornikolov: Things look good so far. What do you think about moving the `getColumnName(rs)` inside `createColumnAccessor`? That way it will be closer to other similar activities. And here we can take the name from the accessor after creating it.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,42 +52,203 @@ public Pattern getParameterPattern() {
...
+                    allParams.put(NameNormaliser.normaliseName(paramName),
```
javornikolov: Possibly we can shrink length of `normaliseName` calls via static import.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: So like this?

```
        public Map<String, DbParameterAccessor> getParams() throws SQLException {
            Map<String, DbParameterAccessor> allParams = new HashMap<String, DbParameterAccessor>();
            try (ResultSet rs = getColumns()) {
                while (rs.next()) {
                    DbParameterAccessor accessor = createColumnAccessor(rs);
                    allParams.put(normaliseName(accessor.getName()),
                            accessor);
                }
            }
            return allParams;
        }

        protected DbParameterAccessor createColumnAccessor(ResultSet rs)
                throws SQLException {
            String paramTypeName = rs.getString("TYPE_NAME");
            return createDbParameterAccessor(getColumnName(rs), columnDirection(rs),
                           typeMapper.getJDBCSQLTypeForDBType(paramTypeName),
                           getJavaClass(paramTypeName), rs.getInt("ORDINAL_POSITION") - 1);
        }
```

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
MMatten: Good idea.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
javornikolov: Yes. The `allParams.put(...)`  is maybe short enough to fit on single line.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,42 +52,203 @@ public Pattern getParameterPattern() {
...
+    }
+
+    abstract private class DatabaseExecutable extends DatabaseObject {
+
+        private DatabaseExecutable(DatabaseObjectName objName) {
+            super(objName);
+        }
+
+        @Override
+        abstract protected Direction getColumnDirection(ResultSet rs) throws SQLException;
```
javornikolov: Abstract stuff doesn't need to be overridden (it's inherited). But here `DatabaseExecutable` is already completely hollow so it can be dropped.

MMatten: Right, yes. I'd still like to see if I can have method names that more closely match the intent for the columns (`PreparedStatement` / `CallableStatement` paramaters) to give some more content to the idea of `Direction`.

I guess maybe `Direction` itself should be `StatementParameterDirection` or the `DBEnvironment` API method names should be more representative. That's a different PR.

Maybe there's some stuff local to this PR. I'll have a play around...

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,42 +52,192 @@ public Pattern getParameterPattern() {
...
+                           getJavaClass(paramTypeName), rs.getInt("ORDINAL_POSITION") - 1);
+        }
+
+        protected String getColumnName(ResultSet rs) throws SQLException {
+            return defaultIfNull(rs.getString("COLUMN_NAME"), "");
+        }
+
+        protected Direction columnDirection(ResultSet rs) throws SQLException {
+            Direction columnDir = getColumnDirection(rs);
+            if (columnDir == null) {
```
javornikolov: I just found out a `java.util.Objects` utility function which could be handy here:
``` java
    return requireNonNull(getColumnDirection(rs), "Our error message....");
```

MMatten: That's neat. I guess we let let the `NullPointerException` just propogate.

javornikolov: Yes, NPE instead of SQLException looks just fine to me.

MMatten: This method isn't defined as throwing an NPE.

https://docs.oracle.com/javase/7/docs/api/java/util/Objects.html#requireNonNull(T,%20java.lang.String)

So I'm not required to `throws NullPointerException`.

We still have `SQLExceptions` from `getColumnDirection` because the the `rs` field access.

javornikolov: Yes, it's not needed to declare `throws` for children of `RuntimeException`. It's fine that we have `SQLException` for the rest.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,42 +52,192 @@ public Pattern getParameterPattern() {
...
+            try (ResultSet rs = getObjectMetaData()) {
+                return rs.next();
+            }
+        }
+
+        abstract protected ResultSet getColumns() throws SQLException;
 
-            ResultSet rs = dc.executeQuery();
+        public Map<String, DbParameterAccessor> getParams() throws SQLException {
             Map<String, DbParameterAccessor> allParams = new HashMap<String, DbParameterAccessor>();
```
javornikolov: Should be fine to call `new HashMap<>()` here, no need to list types on the right side (I think we already did that somewhere).

MMatten:
I'll leave the names as they are at this stage. Any other thoughts on this PR?

javornikolov:
> I'll leave the names as they are at this stage. Any other thoughts on this PR?

It looks good to me :+1: No more comments from my side.

I think we can go on and merge it. What do you think? (If for some reason we conclude that alternative implementation based on plain SQL is better - it won't be a problem to change it).

MMatten:
Regarding the naming I'm now leaning more towards calling stuff `parameter` rather than `column`.  I.e.: table parameters and procedure/function parameters. 

In the API too perhaps methods like `getTableParameters` and `getStoredRoutineParameters` would make things clearer. They're available parameters for statements affecting those target objects. 

What do you think?

MMatten:
I think this PR makes a good foundation for a generic adapter too. May be we should use the SQL query one for Derby and the JDBC one for Generic (as per the jdbcSlim project). Any thoughts?

javornikolov:
> Regarding the naming I'm now leaning more towards calling stuff parameter rather than column. I.e.: table parameters and procedure/function parameters.
> 
> In the API too perhaps methods like getTableParameters and getStoredRoutineParameters would make things clearer. They're available parameters for statements affecting those target objects.
> 
> What do you think?

Indeed, in some places we already have `parameter` as name linked to both sp parameters and table columns. As discussed, maybe makes sense to separate the `metadata` type from the one used to build `statement parameters`. Even in that case, there is a lot of commonality between table columns and sp parameters: and a common name for both would be needed I guess...

In #463 I can see I've introduced another type `DbParameterDescriptor` (much leaner than DbParameterAccessor)... That's unfinished concept though, and I don't remember details of what I had in mind by that time.

javornikolov:
> I think this PR makes a good foundation for a generic adapter too. May be we should use the SQL query one for Derby and the JDBC one for Generic (as per the jdbcSlim project). Any thoughts?

Yes, it's would be good foundation for a generic adapter (though in general it's not clear yet what else would be able compatible). Unless we end up with extra features with the SQL-query alternative, likely it would make sense to reuse it for Derby (otherwise we end up with extra code to maintain for the same purpose).

javornikolov:
I'm merging that... Thank you @MMatten!

PR #527, opened 2016-12-12
Support stored and function parameters in Derby

MMatten:
Here is a (very - plenty of trimming, polishing and cleaning to do) rough look at what a SQL query only approach might look like.

### DIFF
--- a/dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
+++ b/dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
@@ -4,8 +4,10 @@ import dbfit.annotations.DatabaseEnvironment;
 import dbfit.api.AbstractDbEnvironment;
 import dbfit.util.DbParameterAccessor;
 import dbfit.util.NameNormaliser;
+import dbfit.util.Direction;
 
 import java.math.BigDecimal;
+import java.sql.DatabaseMetaData;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
@@ -16,7 +18,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.regex.Pattern;
 
-import dbfit.util.Direction;
+import static org.apache.commons.lang3.ObjectUtils.defaultIfNull;
 
 /**
  * Encapsulates support for the Derby database (also known as JavaDB). Operates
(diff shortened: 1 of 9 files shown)


dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -53,38 +55,148 @@ public Pattern getParameterPattern() {
...
+        "       V.\"LENGTH\"         AS LENGTH," +
+        "       CASE WHEN (V.\"DATA_TYPE\" IN (" + java.sql.Types.DECIMAL + "," +
+        "        " + java.sql.Types.NUMERIC + ", " + java.sql.Types.INTEGER + "," +
+        "        " + java.sql.Types.SMALLINT + ", " + java.sql.Types.TINYINT + ", " +
+        "        " + java.sql.Types.BIGINT + ", " + java.sql.Types.DATE + ", " +
+        "        " + java.sql.Types.TIME + ", " + java.sql.Types.TIMESTAMP + "))" +
+        "            THEN" +
+        "            V.\"SCALE\"" +
+        "            ELSE CAST (NULL AS SMALLINT) END" +
+        "            AS SCALE," +
```
javornikolov: Many of the columns here are not used. If the query after all is still too hairy, might make some sense to stored in a resource file.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -53,38 +55,148 @@ public Pattern getParameterPattern() {
...
             }
-            rs.close();
-            return allParams;
         }
     }
 
-    public Map<String, DbParameterAccessor> getAllProcedureParameters(
-            String procName) throws SQLException {
-        throw new UnsupportedOperationException();
+    private Direction getParamDirection(int columnType, boolean isFunction) {
```
javornikolov: It should be possible the result to be unified on SQL query level without a need to handle results differently for different types of objects (tables, procedures functions).

MMatten:
TravisCI can't seem to download from Maven Central. I'll try it again later.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,42 +53,120 @@ public Pattern getParameterPattern() {
...
+    private Map<String, DbParameterAccessor> createParamMap(ResultSet rs)
+            throws SQLException {
+        Map<String, DbParameterAccessor> allParams = new HashMap<>();
+        while (rs.next()) {
+            String paramName = defaultIfNull(rs.getString("COLUMN_NAME"), "");
+            String paramTypeName = rs.getString("TYPE_NAME");
+            allParams.put(NameNormaliser.normaliseName(paramName),
+                          createDbParameterAccessor(paramName,
+                                                    getParamDirection(rs.getString("COLUMN_DIRECTION")),
+                                                    typeMapper.getJDBCSQLTypeForDBType(paramTypeName),
```
MMatten: Indentation style?

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,42 +53,120 @@ public Pattern getParameterPattern() {
...
+        }
+    }
+
+    @Override
+    public Map<String, DbParameterAccessor> getAllProcedureParameters(String procName)
+                                                                   throws SQLException {
+        // https://github.com/apache/derby/blob/10.13/java/engine/org/apache/derby/impl/jdbc/metadata.properties
+        String sql =
+        "SELECT CAST(NULL AS VARCHAR(128)) AS <<ProcFuncTypeName>>_CAT," +
+        "       S.SCHEMANAME               AS <<ProcFuncTypeName>>_SCHEM," +
```
javornikolov: It makes sense the column aliases in the result to be the same for all the queries. (I'm not sure we need these specific columns for at all for what we're trying to do).

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
javornikolov:
``` java
            allParams.put(normaliseName(paramName),
                createDbParameterAccessor(
                    paramName,
                    Direction(valueOf(rs.getString("COLUMN_DIRECTION"))),
                    typeMapper.getJDBCSQLTypeForDBType(paramTypeName));
```

But it's also possible to break it down in a way which makes lines shorter.
``` java
    while (rs.next()) {
        addSingleParam(allParams, rs);
    }

    addSingleParam(allParams, rs) {
        DbParameterAccessor accessor = createAccessor(rs);
        allParams.put(normaliseName(accessor.getName()), accessor);
    }

    createAccessor(rs) {
        return createDbParameerAccessor(getName(rs), getDirection(rs), getType(rs));
    }
```

MMatten:
> TravisCI can't seem to download from Maven Central. I'll try it again later.

@javornikolov any idea what's going on with TravisCI here? May be I should just force a new build.

javornikolov:
> TravisCI can't seem to download from Maven Central. I'll try it again later.
> @javornikolov any idea what's going on with TravisCI here? May be I should just force a new build.

 I tried yesterday (restarted the Travis build) but it failed. I just triggered the build again, let's see what will happen.

javornikolov:
Oops. I pressed the wrong button... I'm reopening the PR now.

MMatten:
Some tidying on the SQL queries done. Still some tidying up to do in `createParamMap` as per your ideas.

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,42 +53,153 @@ public Pattern getParameterPattern() {
...
+    @Override
+    public Map<String, DbParameterAccessor> getAllProcedureParameters(String procName)
+            throws SQLException {
+        DatabaseObjectName name = buildDatabaseObjectName(procName);
+        // https://github.com/apache/derby/blob/10.13/java/engine/org/apache/derby/impl/jdbc/metadata.properties
+        String sql =
+        "SELECT V.COLUMN_NAME" +
+        "           AS COLUMN_NAME" +
+        "     , CASE WHEN A.ALIASTYPE = 'P'" +
+        "            THEN CASE WHEN V.COLUMN_TYPE = " + DatabaseMetaData.procedureColumnIn +
```
javornikolov: We don't need to decode to some JDBC Type ID here. We can directly have it as string: `INPUT`, `OUTPUT`, ....

dbfit-java/derby/src/main/java/dbfit/environment/DerbyEnvironment.java
```diff
@@ -51,42 +53,153 @@ public Pattern getParameterPattern() {
...
+        "   AND A.SCHEMAID = S.SCHEMAID" +
+        "   AND A.ALIAS = ?" +
+        "   AND S.SCHEMANAME = ?" +
+        " ORDER" +
+        "    BY PARAMETER_ID" +
+        "     , ORDINAL_POSITION";
+        String query = sql.replaceAll("<<ProcFuncTypeFlag>>", "P");
+        Map<String, DbParameterAccessor> params = fetchIntoParams(name, query);
+        if (!params.isEmpty()) {
+            return params;
```
javornikolov: That `trick` perhaps is better to be extracted as small separate function, otherwise it's a bit invisible within the rest of the code.

javornikolov: Let's imagine we have a `PROCEDURE` which doesn't have any parameters, and doesn't have a return value. What will happen in that case?

MMatten:
Superseded by #524.

MMatten:
I guess we should leave the branch as a useful future reference but how do we remember to not delete it!?

PR #535, opened 2016-12-23
Support stored unit calls in HSQLDB

MMatten:
Support stored procedure and function execution in HSQLDB.


dbfit-java/hsqldb/src/main/java/dbfit/environment/HSQLDBEnvironment.java
```diff
@@ -65,47 +68,99 @@ protected Pattern getParameterPattern() {
...
             Map<String, DbParameterAccessor> allParams = new HashMap<String, DbParameterAccessor>();
             int position = 0;
             while (rs.next()) {
-                String columnName = rs.getString(1);
-                String dataType = rs.getString(2);
+                String columnName = rs.getString("PARAMETER_NAME");
+                String dataType = rs.getString("TYPE_NAME");
+                Direction direction = null;
+                switch (rs.getString("PARAMETER_MODE")) {
+                    case "IN":
```
javornikolov: If we have control over query column names, can we just make them match to Direction names (e.g. `IN` -> `INPUT`, `INOUT` -> `INPUT_OUTPUT`? Than it's possible to parse them directly via Direction.

dbfit-java/hsqldb/src/main/java/dbfit/environment/HSQLDBEnvironment.java
```diff
@@ -65,47 +68,99 @@ protected Pattern getParameterPattern() {
...
+                        direction = Direction.RETURN_VALUE;
+                        break;
+                }
                 DbParameterAccessor dbp = createDbParameterAccessor(
                         columnName,
-                        INPUT,
+                        direction,
                         typeMapper.getJDBCSQLTypeForDBType(dataType),
-                        getJavaClass(dataType), position++);
+                        getJavaClass(dataType), position++ - 1);
```
javornikolov: Can we simplify that a bit (the `-1` looks a bit scary :-)).

dbfit-java/hsqldb/src/main/java/dbfit/environment/HSQLDBEnvironment.java
```diff
@@ -65,47 +68,99 @@ protected Pattern getParameterPattern() {
...
                 allParams.put(NameNormaliser.normaliseName(columnName), dbp);
             }
             rs.close();
             return allParams;
         }
     }
 
-    public Map<String, DbParameterAccessor> getAllProcedureParameters(String s)
+    public Map<String, DbParameterAccessor> getAllProcedureParameters(String routineName)
```
javornikolov: The parameter name in `DBEnvironment` is `procName`.

dbfit-java/hsqldb/src/main/java/dbfit/environment/HSQLDBEnvironment.java
```diff
@@ -65,47 +68,99 @@ protected Pattern getParameterPattern() {
...
+                   + "            , para.data_type"
+                   + "            , para.parameter_mode"
+                   + "            , para.ordinal_position"
+                   + "         FROM information_schema.routines rout"
+                   + "        INNER"
+                   + "         JOIN information_schema.parameters para"
+                   + "           ON rout.specific_catalog = para.specific_catalog"
+                   + "          AND rout.specific_schema = para.specific_schema"
+                   + "          AND rout.specific_name = para.specific_name"
+                   + "        UNION"
```
javornikolov: Should it be `UNION` or `UNION ALL`?

dbfit-java/hsqldb/src/main/java/dbfit/util/sql/HSQLDBPreparedStatements.java
```diff
@@ -0,0 +1,8 @@
+package dbfit.util.sql;
+
+public class HSQLDBPreparedStatements extends PreparedStatements {
+
+    public static String buildFunctionCall(String procName, int numberOfParameters) {
+        return "{ call " + procName + "(" + buildParamList(numberOfParameters - 1) + ")}";
```
javornikolov: This looks pretty much like `buildStoredProcedureCall` with reduced number of parameters. What about the option to rather use `buildStoredProcedureCall` directly instead of introducing this child HSQLDBPreparedStatements?

PreparedStatements used to be a simple bunch of helpers. Possibly their names were not the best. Would it be better to have `buildCallWithReturnValue` and `builtCallWithoutReturnValue`?

dbfit-java/hsqldb/src/main/java/dbfit/environment/HSQLDBEnvironment.java
```diff
@@ -65,47 +68,99 @@ protected Pattern getParameterPattern() {
...
+            dc.setString(2, schemaName);
             ResultSet rs = dc.executeQuery();
             Map<String, DbParameterAccessor> allParams = new HashMap<String, DbParameterAccessor>();
             int position = 0;
             while (rs.next()) {
-                String columnName = rs.getString(1);
-                String dataType = rs.getString(2);
+                String columnName = rs.getString("PARAMETER_NAME");
+                String dataType = rs.getString("TYPE_NAME");
+                Direction direction = null;
```
javornikolov: Maybe it's OK but some wondering emerges about that `null` initialization.

dbfit-java/core/src/main/java/dbfit/api/DBEnvironment.java
```diff
@@ -189,4 +189,6 @@ void connectUsingFile(String filePath) throws SQLException, IOException,
             FileNotFoundException;
 
     DbStoredProcedureCall newStoredProcedureCall(String name, DbParameterAccessor[] accessors);
+
+    String buildFunctionCall(String procOrFuncName, int numberOfParameters);
```
MMatten: I guess that new method is part of a environment vendor API rather than a client API but we haven't separated them.

javornikolov: Yes, it's true we don't have these cleanly separated yet. Still I'm wondering what simple enough alternative options could we figure out for that specific case now.

dbfit-java/core/src/main/java/dbfit/fixture/FunctionResultSetStatementExecution.java
```diff
@@ -1,14 +1,12 @@
-package dbfit.environment;
+package dbfit.fixture;
```
MMatten: That's the same as the base class. In that other big API related PR I think we moved it (StatementExecution`) out to somewhere else.

dbfit-java/core/src/main/java/dbfit/api/DBEnvironment.java
MMatten: What about starting to separate it now with a different interface for the vendor API under `api.vendor` that `AbstractDbEnvironment` also implements?

dbfit-java/core/src/main/java/dbfit/fixture/FunctionResultSetStatementExecution.java
javornikolov: Ah, indeed. OK then.

dbfit-java/core/src/main/java/dbfit/api/DBEnvironment.java
javornikolov: It's possible but there is some risk to end up with too deep refactoring again. It's actually possible to have smaller more focused interface for the part of interest now.

But what about just extending `DbStoredProcedureCall`? I think this way we can avoid the need for tweaking the environment API. So far this has been the way to customize behavior e.g. for Oracle.

MMatten: > But what about just extending DbStoredProcedureCall? I think this way we can avoid the need for tweaking the environment API. So far this has been the way to customize behavior e.g. for Oracle.

It seems by doing that that we are hiding, or at least not making explicit, the requirements/contract that an environment vendor must fulfil. We have our implementation and we can get away with changes under the covers. 

What would you think about establishing a minimal start at a vendor interface?

javornikolov: For stored procedures we already have `DbStoredProcedureCall newStoredProcedureCall(String name, DbParameterAccessor[] accessors)` which should be fulfilled by the environment.

Let's think about what we're unhappy about it? (Some points are maybe already touched in #463 - e.g. like returning type is *too much*, but it doesn't hurt to take a fresh look at it).

MMatten:
@javornikolov it might be a few days before I look at this again so I'd just like to thank you for your generosity and patience and to wish you a very Merry Christmas and a very Happy 2017. Talk soon. Mark

javornikolov:
Thank you @MMatten! Same to you, it's a pleasure working with you. Thank you for the thoughtful discussions, questions, and great contributions! Wishing you Merry Christmas, happy holidays and lots of health and success in 2017!

dbfit-java/hsqldb/src/main/java/dbfit/environment/HSQLDBEnvironment.java
```diff
@@ -16,7 +19,7 @@
 import java.util.Map;
 import java.util.regex.Pattern;
 
-import static dbfit.util.Direction.*;
+import dbfit.util.Direction;
```
MMatten: Just to allow me to declare a variable of type `Direction`.

dbfit-java/hsqldb/src/main/java/dbfit/environment/HSQLDBEnvironment.java
MMatten: The values are currently coming from the HSQLDB dictionary views but I can translate them during the query (then parse with `Direction.valueOf()`).

dbfit-java/hsqldb/src/main/java/dbfit/environment/HSQLDBEnvironment.java
MMatten: To be it's not a clear cut case which is more natural. If `UNION ALL` were defined as 'do not remove duplicates' then it might suggest that we have them and we want to allow them.

If we were to use `UNION` then it might suggest that we have, or could have, duplicates and that we want them removed.

What's your view?

dbfit-java/hsqldb/src/main/java/dbfit/environment/HSQLDBEnvironment.java
MMatten: I'll align them.

dbfit-java/hsqldb/src/main/java/dbfit/environment/HSQLDBEnvironment.java
MMatten: I'm now parsing with `Direction.valueOf()`. Would you still use 'static'?

dbfit-java/core/src/main/java/dbfit/api/DBEnvironment.java
MMatten: > Let's think about what we're unhappy about it? 

For starters `DbStoredProcedureCall` should be an interface, perhaps, for the vendor API (as well as the other return types in `DBEnvironment`) and the default implmentation could be renamed `DefaultDbStoredProcedureCall` (or similar) but I do want to keep this PR to minimal refactoring of course, and split off anything else to another one.

I'll explore what an approach of changing `DbStoredProcedureCall newStoredProcedureCall(String name, DbParameterAccessor[] accessors)` might look like.

MMatten: The existing/proposed changes to `DBEnvironment` were in line with what we'd done previously, especially for Informix, where we allowed environment-specific function call construction/execution.

MMatten: OK. I've removed the `DBEnvironment` changes and created PR #541.

dbfit-java/hsqldb/src/main/java/dbfit/util/sql/HSQLDBPreparedStatements.java
MMatten: I've had a go at generalising the helper class in #541

MMatten:
Superseded by #554.

PR #541, opened 2017-02-04
Support stored unit calls in HSQLDB with no DBEnvironment changes

MMatten:
Support stored procedure and function execution in HSQLDB.

TODO:
- [ ] Decide about current `DbStoredProcedureCall` interface and classes implementing it (`DefaultDbStoredProcedureCall`, `OracleStoredProcedureCall`). 
- [ ] Package name for any new/modified API objects.
- [x] Check if `isFunction` is a good name (not so critical though since it's just a private one)
- [ ] Other stuff (e.g. ???)
- [ ] Decide upon naming of `FunctionResultSetStatementExecution` (or `FunctionStatementExecution` or `ResultSetStatementExecution`, ...).


MMatten:
Now contradicting the current PR title, I've extracted a `DbStoredProcedureCall` interface from the existing implementation. This does still include a reference to the `StatementExecution` implementation so maybe it would be good to create the others all in separate PR.

But, what do you think of this direction?

dbfit-java/core/src/main/java/dbfit/api/vendor/DbStoredProcedureCall.java
```diff
@@ -0,0 +1,16 @@
+package dbfit.api.vendor;
```
javornikolov: We're introducing a new package here... Is it a good place to have it under the current `api` package?

javornikolov: Also some wondering about the `vendor` - is it a good name?

dbfit-java/core/src/main/java/dbfit/fixture/FunctionResultSetStatementExecution.java
```diff
@@ -1,14 +1,12 @@
-package dbfit.environment;
+package dbfit.fixture;
 
 import java.sql.*;
 
-import dbfit.fixture.StatementExecution;
-
-public class InformixFunctionStatementExecution extends StatementExecution {
+public class FunctionResultSetStatementExecution extends StatementExecution {
```
MMatten: The function return value is retried via a `ResultSet` rather than via a `RETURN_VALUE` parameter. There's probably a better name yet to be imagined.

dbfit-java/core/src/main/java/dbfit/api/vendor/DbStoredProcedureCall.java
MMatten: I'm not sure. I was thinking of something like `api.vendor` and `api.client` packages.

dbfit-java/core/src/main/java/dbfit/api/vendor/DefaultDbStoredProcedureCall.java
```diff
@@ -1,25 +1,29 @@
...
 
-    public DbStoredProcedureCall(DBEnvironment environment, String name, DbParameterAccessor[] accessors) {
+    public DefaultDbStoredProcedureCall(DBEnvironment environment, String name, DbParameterAccessor[] accessors) {
         this.environment = environment;
         this.name = name;
         this.accessors = new DbParameterAccessors(accessors);
     }
 
+    /* (non-Javadoc)
+     * @see dbfit.api.IDbStoredProcedureCall#getName()
```
MMatten: That's an _oops_ created by Eclipse refactoring functionality. Apologies for the sloppyness.

dbfit-java/core/src/main/java/dbfit/util/sql/PreparedStatements.java
```diff
@@ -8,11 +8,9 @@ private static String buildParamList(int numberOfParameters) {
         return join(repeat("?", numberOfParameters), ", ");
     }
 
-    public static String buildStoredProcedureCall(String procName, int numberOfPararameters) {
-        return "{ call " + procName + "(" + buildParamList(numberOfPararameters) + ")}";
-    }
-
-    public static String buildFunctionCall(String procName, int numberOfParameters) {
-        return "{ ? = call " + procName + "(" + buildParamList(numberOfParameters - 1) + ")}";
+    public static String buildStoredRoutineCallCmdText(
```
MMatten: Either `buildStoredRoutineCallCommand` or `buildStoredRoutineCallText` might do. I think `buildStoredRoutineCallStatementText` is the most expressive.

dbfit-java/core/src/main/java/dbfit/fixture/FunctionResultSetStatementExecution.java
javornikolov: > Now contradicting the current PR title, I've extracted a DbStoredProcedureCall interface from the existing implementation. This does still include a reference to the StatementExecution implementation so maybe it would be good to create the others all in separate PR.
> 
> But, what do you think of this direction?

Looks like a potentially good approach to examine.

dbfit-java/core/src/main/java/dbfit/api/vendor/DbStoredProcedureCall.java
```diff
@@ -0,0 +1,16 @@
...
+
+import java.sql.SQLException;
+
+import dbfit.fixture.StatementExecution;
+
+public interface DbStoredProcedureCall {
+
+    String getName();
+
+    boolean isFunction();
```
MMatten: Not really. It's really for internal use in the implementations.

dbfit-java/core/src/main/java/dbfit/api/vendor/DbStoredProcedureCall.java
```diff
@@ -0,0 +1,16 @@
...
+
+    String toSqlString();
```
MMatten: Not really. It's really for internal use in the implementations.

MMatten: Stripping out both `toSqlString()` and `isFunction` we're left with just: -

```
package dbfit.api.vendor;

import java.sql.SQLException;

import dbfit.fixture.StatementExecution;

public interface DbStoredProcedureCall {

    StatementExecution toStatementExecution() throws SQLException;
}
```

On `DbObject` we already have `StatementExecution buildPreparedStatement(DbParameterAccessor accessors[])` and for stored routines this creates the appropriate procedure/function executor but for tables creates an `INSERT` (not other types) statement executor. This is a bit obscure.

Perhaps we don't need a separate `DbStoredProcedureCall` interface and implementations. Perhaps the classes implementing `DbObject` should provide these executors directly. What do you think?

dbfit-java/core/src/main/java/dbfit/util/sql/PreparedStatements.java
```diff
@@ -8,11 +8,9 @@ private static String buildParamList(int numberOfParameters) {
...
+            String routineName, int numberOfPararameters, boolean paramsIncludeReturnValue) {
+        return "{ " + (paramsIncludeReturnValue ? "? = " : "") + "call " + routineName + "(" +
```
MMatten: I've pushed a revision to this. Hopefully a little better.

dbfit-java/core/src/main/java/dbfit/api/vendor/DbStoredProcedureCall.java
MMatten: @javornikolov any thoughts on, perhaps, removing the `DbStoredProcedureCall` interface and letting `buildPreparedStatement` of `DbObject` create the appropriate statement executor?

dbfit-java/core/src/main/java/dbfit/util/sql/PreparedStatements.java
javornikolov: Yep, it's better. Still a little bit convoluted. I would think of something like following if I have to describe the syntax:
```
routine_invocation: <routine_name>(<params_list>)
routine_with_return_value:    { ? = call <routine_invocation> }
routine_without_return_value: { call <routine_invocation> }
```

That's subjective though... (there could be simpler way to express it). My point is just to make that form more obvious from the code.

dbfit-java/core/src/main/java/dbfit/api/vendor/DbStoredProcedureCall.java
javornikolov: > @javornikolov any thoughts on, perhaps, removing the DbStoredProcedureCall interface and letting buildPreparedStatement of DbObject create the appropriate statement executor?

@MMatten, I took also a look at #463 which is touching related parts. Looks like what's enough for `DbObject` is quite simple (just a method to execute it and ability to release resources).

Maybe indeed a separate `DbStoredProcedureCall` interface would be redundant (something simpler like `DbCommand` should be enough).

dbfit-java/core/src/main/java/dbfit/api/vendor/DbStoredProcedureCall.java
javornikolov: https://theamiableapi.com/java-cl-section-1-2/#cl.item.1.2.7

MMatten: Ok. So we should drop the `api` part completely, may be in a new PR.

dbfit-java/core/src/main/java/dbfit/util/sql/PreparedStatements.java
```diff
@@ -8,11 +8,16 @@ private static String buildParamList(int numberOfParameters) {
...
 
-    public static String buildFunctionCall(String procName, int numberOfParameters) {
-        return "{ ? = call " + procName + "(" + buildParamList(numberOfParameters - 1) + ")}";
+    public static String buildStoredRoutineCallStatementText(
+            String routineName, int numberOfPararameters, boolean hasReturnValueParameter) {
+        String callText = "{ " + buildReturnValue(hasReturnValueParameter);
+        callText += "call " + routineName + "(";
+        callText += buildParamList((hasReturnValueParameter ? numberOfPararameters - 1 : numberOfPararameters));
+        callText += ") }";
+        return callText;
```
javornikolov: What if we skip the `callText` temporary variable here?

And yet again, it's a bit hard to grasp the overall call pattern. As soon as it makes things clearer some temp variables would be OK.

dbfit-java/core/src/main/java/dbfit/util/sql/PreparedStatements.java
```diff
@@ -8,11 +8,16 @@ private static String buildParamList(int numberOfParameters) {
...
+        return returnsValue ? "? = " : "";
     }
 
-    public static String buildFunctionCall(String procName, int numberOfParameters) {
-        return "{ ? = call " + procName + "(" + buildParamList(numberOfParameters - 1) + ")}";
+    public static String buildStoredRoutineCallStatementText(
+            String routineName, int numberOfPararameters, boolean hasReturnValueParameter) {
+        String callText = "{ " + buildReturnValue(hasReturnValueParameter);
+        callText += "call " + routineName + "(";
+        callText += buildParamList((hasReturnValueParameter ? numberOfPararameters - 1 : numberOfPararameters));
```
javornikolov: That could also be `numberOfParameters - (hasReturnValueParameter ? 1 : 0)`.

MMatten: I've pushed another draft for this.

dbfit-java/core/src/main/java/dbfit/util/sql/PreparedStatements.java
```diff
@@ -8,11 +8,18 @@ private static String buildParamList(int numberOfParameters) {
...
+        return returnsValue ? "? = " : "";
     }
 
-    public static String buildFunctionCall(String procName, int numberOfParameters) {
-        return "{ ? = call " + procName + "(" + buildParamList(numberOfParameters - 1) + ")}";
+    private static String buildRoutineInvocation(String routineName, int numberOfParameters) {
+        return "call " + routineName + "(" + buildParamList(numberOfParameters) + ")";
+    }
+
+    public static String buildStoredRoutineCallStatementText(
```
javornikolov: What about naming that `buildStoredRoutineCallText`?

dbfit-java/core/src/main/java/dbfit/util/sql/PreparedStatements.java
```diff
@@ -8,11 +8,18 @@ private static String buildParamList(int numberOfParameters) {
...
 
-    public static String buildStoredProcedureCall(String procName, int numberOfPararameters) {
-        return "{ call " + procName + "(" + buildParamList(numberOfPararameters) + ")}";
+    private static String buildReturnValue(boolean returnsValue) {
+        return returnsValue ? "? = " : "";
     }
 
-    public static String buildFunctionCall(String procName, int numberOfParameters) {
-        return "{ ? = call " + procName + "(" + buildParamList(numberOfParameters - 1) + ")}";
+    private static String buildRoutineInvocation(String routineName, int numberOfParameters) {
```
javornikolov: I got thinking about `buildRoutineInvocation` vs simply `routineInvocation` (at least for the private functions here). What would be easier to follow when reading the code?

dbfit-java/core/src/main/java/dbfit/api/vendor/DbStoredProcedureCall.java
MMatten: Might we still have `client` and `vendor` though?

javornikolov: Maybe, I don't know yet. There also a vendor of the testing framework (that's FitNesse/Fit). It's a bigger topic I think other classes also need to find a better home. So maybe it's better to defer package reorganizations for another PR unless we really benefit from it here.

dbfit-java/core/src/main/java/dbfit/api/vendor/DbStoredProcedureCall.java
javornikolov: What we decide here? Maybe we can drop the redundant functions from the interface and see if we'd like to drop the whole interface or to keep it as subsequent step?

dbfit-java/core/src/main/java/dbfit/util/sql/PreparedStatements.java
MMatten: Yes. I like that better.

dbfit-java/core/src/main/java/dbfit/api/vendor/DbStoredProcedureCall.java
MMatten: Well, I was thinking we could drop the `DbStoeedProcedureCall`interface and merge the remnants into `DbObject`.

javornikolov: That might be fine too - as soon as it's simple enough.

dbfit-java/core/src/main/java/dbfit/util/sql/PreparedStatements.java
MMatten: Yes, I think I like that better. Or perhaps `executeRoutine`. I think any are fine. I'll push an update.

dbfit-java/core/src/main/java/dbfit/util/sql/PreparedStatements.java
```diff
@@ -8,11 +8,23 @@ private static String buildParamList(int numberOfParameters) {
...
+    public static String storedRoutineStatement(
+            String routineName, int numberOfParams, boolean hasReturnValueParam) {
+        int executionParams = numberOfParams - (hasReturnValueParam ? 1 : 0);
+        return returnValueParam(hasReturnValueParam) +
+            storedRoutineExecution(routineName, executionParams);
+    }
+
+    public static String storedRoutineCall(
+            String routineName, int numberOfParams, boolean hasReturnValueParam) {
+        return "{ " + storedRoutineStatement(routineName, numberOfParams, hasReturnValueParam) + " }";
```
javornikolov: We ended up with several names: `execution`, `statement`, `call`... I may add `invocation`. I'm wondering whether we can reduce the confusion, maybe there is no ideal answer... General ideas:
* The public function seems most eligible to tag with `statement` as part of its name (the others are some fractions of the final statement)
* I'm not sure whether all slicing into separate functions helps (at the moment). An example with less procs might look like:

``` java 
    public static String buildStoredRoutineCallText(
            String name, int numberOfParams, boolean hasReturnValueParam) {
        int numArgs = numberOfParams - (hasReturnValueParam ? 1 : 0);
        String resultAssignment = hasReturnValueParam ? "? = " : "";
        String voidInvocation = name + "(" + buildParamList(numArgs) + ")";
        return "{ " + resultAssignment + "call " + voidInvocation + " }";
    }
```

MMatten: Great. I'm happy with that. I'll apply it directly.

dbfit-java/core/src/main/java/dbfit/util/sql/PreparedStatements.java
```diff
@@ -8,11 +8,23 @@ private static String buildParamList(int numberOfParameters) {
...
+
+    public static String storedRoutineStatement(
+            String routineName, int numberOfParams, boolean hasReturnValueParam) {
+        int executionParams = numberOfParams - (hasReturnValueParam ? 1 : 0);
+        return returnValueParam(hasReturnValueParam) +
+            storedRoutineExecution(routineName, executionParams);
+    }
+
+    public static String storedRoutineCall(
+            String routineName, int numberOfParams, boolean hasReturnValueParam) {
```
MMatten: Only `returnsParam` (not so sure about this), `returnsVal` and `hasReturn` come to mind. Any other offers?

dbfit-java/core/src/main/java/dbfit/util/sql/PreparedStatements.java
```diff
@@ -8,11 +8,23 @@ private static String buildParamList(int numberOfParameters) {
...
 
-    public static String buildFunctionCall(String procName, int numberOfParameters) {
-        return "{ ? = call " + procName + "(" + buildParamList(numberOfParameters - 1) + ")}";
+    private static String storedRoutineExecution(String routineName, int numberOfParameters) {
+        return "call " + routineName + "(" + buildParamList(numberOfParameters) + ")";
+    }
+
+    public static String storedRoutineStatement(
+            String routineName, int numberOfParams, boolean hasReturnValueParam) {
+        int executionParams = numberOfParams - (hasReturnValueParam ? 1 : 0);
```
MMatten: Yes. That one's not so good. But we're going with `numArgs` now. May be we should be consistent (`numArgs` and `numParams`).

javornikolov: In one of the cases we're counting the return value as a parameter, in the other - not. It's good to somehow differentiate... That's not quite clear based on `args` vs `params` either - but it's kind of a hint that they're different.

MMatten:
@javornikolov I've tried rounding up the issues into the TODO list in the PR text. Anything to add?

javornikolov:
> @javornikolov I've tried rounding up the issues into the TODO list in the PR text. Anything to add?

Thanks, I'll take a look again.

dbfit-java/core/src/main/java/dbfit/api/vendor/DefaultDbStoredProcedureCall.java
```diff
@@ -28,26 +28,23 @@ protected DbParameterAccessors getAccessors() {
         return accessors;
     }
 
-    public boolean isFunction() {
+    private boolean isFunction() {
```
javornikolov: With those special cases I'm wondering whether `isFunction` is a good name or it should rather be something like `hasReturnValue`?

dbfit-java/oracle/src/main/java/dbfit/environment/OracleStoredProcedureCall.java
```diff
@@ -15,13 +15,13 @@
 
 import dbfit.util.Direction;
 
-public class OracleStoredProcedureCall extends DbStoredProcedureCall {
+public class OracleStoredProcedureCall extends DefaultDbStoredProcedureCall {
```
javornikolov: We're _extending_ the **default** in order to get a **non-default** implementation. It used to be that way before, just now the naming of the base class made it more obvious.
Perhaps that shouldn't bother us much right now - unless it helps to resolve some of the other questions.

javornikolov:
> @javornikolov I've tried rounding up the issues into the TODO list in the PR text. Anything to add?

I added a small question about the `private isFunction` name. Regarding the package names - if we still keep `DbStoredProcedureCall`, maybe it's fine to remain in its original package for now.

dbfit-java/core/src/main/java/dbfit/api/vendor/DefaultDbStoredProcedureCall.java
MMatten: > With those special cases I'm wondering whether `isFunction` is a good name or it should rather be something like `hasReturnValue`?

I think `hasReturnValue` is better. We might come across some future DBMS where we have stored procedures that also have a return value as well as functions.

dbfit-java/oracle/src/main/java/dbfit/environment/OracleStoredProcedureCall.java
MMatten: Any ideas for renaming these at this point? May be drop the `Db` prefix off of the interface or the classes...?

javornikolov: I think we can keep the names as they are now for this PR. Maybe the class hierarchy and choice of inheritance needs revising but that's a separate story.

dbfit-java/core/src/main/java/dbfit/api/vendor/DefaultDbStoredProcedureCall.java
MMatten: Applied + pushed.

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
```diff
@@ -241,7 +243,7 @@ public String buildInsertCommand(String tableName,
     }
 
     public DbStoredProcedureCall newStoredProcedureCall(String name, DbParameterAccessor[] accessors) {
```
MMatten: Now I'm thinking...couldn't this method just return a `StatementExecution` directly instead of a new `DbStoredProcedureCall` to `DbObject.buildPreparedStatement`?

I'll push an experimental commit to review.

MMatten: Done (553ed4e5b78e091c8a40aba02fa4b2241020f9b6).

javornikolov: At first glimpse that makes sense.

MMatten: Now I'm not so sure. I've now made the new interface `DbStoredProcedureCall` somewhat redundant :( I'm in danger of slowly moving into the area of #463 in this PR. I can see the evolution to `DbCommand` / `DbStatement` again now.

I can roll this back. Are we happy with the current `DbStoredProcedureCall` and the related classes for this PR?

MMatten:
I've just pushed another commit (I'll need to tidy up later on) to introduce `DbCommand` from PR #463.

javornikolov:
> I've just pushed another commit (I'll need to tidy up later on) to introduce DbCommand from PR #463.

OK, we can keep that.

dbfit-java/core/src/main/java/dbfit/api/DbCommand.java
```diff
@@ -0,0 +1,25 @@
+package dbfit.api;
+
+import java.sql.SQLException;
+
+/**
+ * An object representing an executable database operation
+ * with no result returned.
```
MMatten: This condition is actually violated by `FunctionResultSetStatementExecution` as this is used for calling functions.

MMatten: Should we loosen our definition here and re-address it when we implement for of PR #463?

javornikolov: Even in #463 I think we're using `DbCommand` for functions too. And also it's always possible to have output parameters of non-function stored procedures.

Maybe with that comment my intent was to make some distinction with `DbQuery` and to highlight that results are not retrieved through the interface itself.

MMatten:
Regarding 
> - [ ] Decide about current DbStoredProcedureCall interface and classes implementing it (DefaultDbStoredProcedureCall, OracleStoredProcedureCall).
I'm no longer sure of the `Default` prefix here so I'm neutral about keeping it or reverting it. What's your preference?

MMatten:
Regarding
> - [ ] Package name for any new/modified API objects.

I'm not wedded to the `vendor` sub-package name in `dbfit.api.vendor.DefaultDbStoredProcedureCall` and it might be better to save this for implementing more of PR #463.

What's your preference?

javornikolov:
> I'm no longer sure of the Default prefix here so I'm neutral about keeping it or reverting it. What's your preference?

I would suggest reverting it because it's diverging a bit from the focus of the current PR. We might refactor it separately if it makes sense. (In general maybe `Default` is a more meaningful name; and possibly the Oracle implementation needs some refactoring. But I'm not sure yet how exactly, maybe it's good enough now.)

MMatten:
Regarding
> Decide upon naming of FunctionResultSetStatementExecution (or FunctionStatementExecution or ResultSetStatementExecution, ...).

I'm stumped for a better name at the moment. My last one for now is `StatementReturningResultSetExecution`. Have you got anything better, otherwise are we OK to stick with these for now?

javornikolov:
> I'm not wedded to the vendor sub-package name in dbfit.api.vendor.DefaultDbStoredProcedureCall and it might be better to save this for implementing more of PR #463.
> 
> What's your preference?

Similar as the other question. Let's keep this PR focused on its primary goal.

javornikolov:
> Regarding
> 
> Decide upon naming of FunctionResultSetStatementExecution (or FunctionStatementExecution or ResultSetStatementExecution, ...).
> I'm stumped for a better name at the moment. My last one for now is StatementReturningResultSetExecution. Have you got anything better, otherwise are we OK to stick with these for now?

What about `StatementExecutionCapturingResultSetValue `?

MMatten:
> What about StatementExecutionCapturingResultSetValue?

Sounds as good as any so far. I'll apply this all tomorrow and hopefully be ready to merge.

MMatten:
>> I've just pushed another commit (I'll need to tidy up later on) to introduce DbCommand from PR #463.

> OK, we can keep that.

Would it be better to chop this part out onto another PR or OK on this one as a separate commit?

javornikolov:
> Would it be better to chop this part out onto another PR or OK on this one as a separate commit?

I separate PR would be nice.

dbfit-java/core/src/main/java/dbfit/api/DbObject.java
```diff
@@ -1,12 +1,11 @@
...
 
-import dbfit.fixture.StatementExecution;
 import dbfit.util.DbParameterAccessor;
 import dbfit.util.Direction;
 
 import java.sql.SQLException;
 
 public interface DbObject {
-    public StatementExecution buildPreparedStatement(DbParameterAccessor accessors[]) throws SQLException ;
+    public DbCommand buildPreparedStatement(DbParameterAccessor accessors[]) throws SQLException ;
```
javornikolov: There is extra space before `;` at the end.

MMatten:
Superseded by #554.

Issue #542, opened 2017-02-08
Dev VM provisioning failure (lvm pvcreate /dev/sdb)


MMatten:
I'm guessing this is because the CentOS image has been updated yet again and breaks the provisioning.

```
==> default:     ================================================================================
==> default:     Error executing action `create` on resource 'lvm_physical_volume[/dev/sdb]'
==> default:     ================================================================================
==> default:     
==> default:     LVM::External::ExternalFailure
==> default:     ------------------------------
==> default:     Fatal error, `/sbin/lvm pvcreate /dev/sdb ` returned 5 with 'Device /dev/sdb not found (or ignored by filtering).'
==> default:     
==> default:     Cookbook Trace:
==> default:     ---------------
==> default:     /var/chef/cache/cookbooks/lvm/libraries/provider_lvm_physical_volume.rb:51:in `action_create'
==> default:     
==> default:     Resource Declaration:
==> default:     ---------------------
==> default:     # In /var/chef/cache/cookbooks/dbfit_test/recipes/diskspace.rb
==> default:     
==> default:       1: lvm_physical_volume '/dev/sdb' do
==> default:       2:   action :create
==> default:       3: end
==> default:       4: 
==> default:     
==> default:     Compiled Resource:
==> default:     ------------------
==> default:     # Declared in /var/chef/cache/cookbooks/dbfit_test/recipes/diskspace.rb:1:in `from_file'
==> default:     
==> default:     lvm_physical_volume("/dev/sdb") do
==> default:       provider Chef::Provider::LvmPhysicalVolume
==> default:       action [:create]
==> default:       retries 0
==> default:       retry_delay 2
==> default:       default_guard_interpreter :default
==> default:       declared_type :lvm_physical_volume
==> default:       cookbook_name "dbfit_test"
==> default:       recipe_name "diskspace"
==> default:     end
==> default:     
==> default:     Platform:
==> default:     ---------
==> default:     x86_64-linux
==> default:     
==> default: [2017-02-08T10:27:41+00:00] INFO: Running queued delayed notifications before re-raising exception
==> default: 
==> default: Running handlers:
==> default: [2017-02-08T10:27:41+00:00] ERROR: Running exception handlers
==> default: Running handlers complete
==> default: [2017-02-08T10:27:41+00:00] ERROR: Exception handlers complete
==> default: Chef Client failed. 9 resources updated in 01 minutes 02 seconds
[...]

PR #543, opened 2017-02-11
Add DB2 to TravisCI build

MMatten:
Add DB2 integration tests to CI builds.

This seems to add roughly 2 minutes to the build time.

It might be possible run the container setup and DB scheme create as a background job, as per the Oracle/Docker setup, but I can't see how the method for Oracle ensure that the schema creation has completed successfully (@javornikolov any pointers welcomed).


javornikolov:
> It might be possible run the container setup and DB scheme create as a background job, as per the Oracle/Docker setup, but I can't see how the method for Oracle ensure that the schema creation has completed successfully (@javornikolov any pointers welcomed).

Yep, it was a bit tricky. The Oracle Docker image starts `ssh` server as last step  https://github.com/wnameless/docker-oracle-xe-11g/blob/master/Dockerfile (i.e. if sshd is up - it means Oracle is already up). So I'm waiting for SSH daemon (remapped to port 49161) - via `nc` polling for an appropriate SSH response (its not enough to check that the port is open since the port remapping makes it seem like open too early).

.travis.yml
```diff
@@ -5,6 +5,8 @@ services:
 
 language: java
 before_install:
+  - docker run --name dbfitdb2 -d -p 50000:50000 -v "`pwd`/dbfit-java/db2/src/integration-test/resources:/var/dbfit/dbfit-java/db2/src/integration-test/resources" -e DB2INST1_PASSWORD=db2inst1-pwd -e LICENSE=accept ibmcom/db2express-c:latest db2start
+  - docker exec dbfitdb2 /var/dbfit/dbfit-java/db2/src/integration-test/resources/create_db_schema.sh
```
javornikolov: Also I moved the oracle scripts below from java resources folder to `test_vm/scripts` - since it's not exactly a Java resource. Maybe we can do the same for db2?

.travis.yml
```diff
@@ -5,6 +5,8 @@ services:
 
 language: java
 before_install:
+  - docker run --name dbfitdb2 -d -p 50000:50000 -v "`pwd`/dbfit-java/db2/src/integration-test/resources:/var/dbfit/dbfit-java/db2/src/integration-test/resources" -e DB2INST1_PASSWORD=db2inst1-pwd -e LICENSE=accept ibmcom/db2express-c:latest db2start
```
javornikolov: That's a pretty long line. Can we do anything about it?

MMatten:
>> It might be possible run the container setup and DB scheme create as a background job, as per the Oracle/Docker setup, but I can't see how the method for Oracle ensure that the schema creation has completed successfully (@javornikolov any pointers welcomed).

> Yep, it was a bit tricky. The Oracle Docker image starts ssh server as last step https://github.com/wnameless/docker-oracle-xe-11g/blob/master/Dockerfile (i.e. if sshd is up - it means Oracle is already up). So I'm waiting for SSH daemon (remapped to port 49161) - via nc polling for an appropriate SSH response (its not enough to check that the port is open since the port remapping makes it seem like open too early).

For Oracle, I not sure we can  tell if the Oracle schema setup script has run through to completion.

For DB2, I think the details are those in the `docker-entrypoint.sh` (https://github.com/IMC3ofC/db2express-c.docker/blob/master/docker-entrypoint.sh) but even then I think I don't know enough about DB2 services to be able to check for something specific that would mean everything's up).

In any case, the DB2 schema setup is run with `docker exec`, after starting the DB2 services. 

I think I could create something to poll for the existence of some of the DbFit test objects. I guess that would be the best option). But, is it worth bothering right now with this as we're only adding a couple of minutes to the process? @javornikolov What do you think?

javornikolov:
> I think I could create something to poll for the existence of some of the DbFit test objects. I guess that would be the best option). But, is it worth bothering right now with this as we're only adding a couple of minutes to the process? @javornikolov What do you think?

To be honest - I think you run it in background mode anyway (`-d` means detached mode). So I'm not sure you can guarantee DB2 schema is up on time with the current setup.

.travis.yml
MMatten: I think putting these long command lines into shell scripts is probably the way to go. I could call a shell script instead in that `before_install` section.

But I'm not sure where the files from the Git repo are placed on the build server filesystem. Do you know?

.travis.yml
MMatten: Indeed. I meant to ask you about that. I've got shell and SQL scripts for this one. Do you think we should bother having `shell` and `sql` directories?

.travis.yml
javornikolov: You shouldn't rely on hard-coded paths, you only rely on having current dir in the project root.
Anyway - separate script is perhaps not worth for that. Lines can be wrapped like I did in the Oracle polling script. Also these paths seem excessively long to me (do we really need so long paths?)

.travis.yml
javornikolov: The reason I placed the `.sql` in separate folder alone is because the Oracle docker startup script is picking everything from the given path and running it.

MMatten:
> To be honest - I think you run it in background mode anyway (-d means detached mode). So I'm not sure you can guarantee DB2 schema is up on time with the current setup.

It's proven to be reliable so far. The DB/schema/objects creation step has never failed so the container startup seems to be a very fast method of getting the services running. May be the actual `docker run` command is actually synchronous then in that respect.

Do you think we need to do any more with this at the moment?

javornikolov:
I can see at the end of docker-entrypoint.sh it's waiting for some entries in a log file - maybe that guarantees that the db2 is up.

> Do you think we need to do any more with this at the moment?

Perhaps not.

build.gradle
```diff
@@ -246,6 +246,7 @@ task fastbuild(dependsOn: [
 
 task travisbuild(dependsOn: [
     fastbuild,
+    ':dbfit-java:db2:integrationTest',
```
javornikolov: What about moving that to the end of the list? I think mysql is the most light-weight so good to have it on first place I think.

MMatten:
> I can see at the end of docker-entrypoint.sh it's waiting for some entries in a log file - maybe that guarantees that the db2 is up.

This seems to come into effect when `-d` is passed, as the first arg, to `docker-entrypoint.sh` but I'm not sure what would invoke this or how. 

Having had a look at the log file though (it gets pretty big) I can see 
```
MESSAGE : ADM7513W  Database manager has started.
```
which is probably something we can use. So we could get the Oracle and DB2 being pulled, extracted and started up in parallel. I'll try calling the DB/schema creation then from the `before_script` section.

MMatten:
>> Having had a look at the log file though (it gets pretty big) I can see
>
> `MESSAGE : ADM7513W  Database manager has started.`
>
>which is probably something we can use. So we could get the Oracle and DB2 being pulled, extracted and >started up in parallel. I'll try calling the DB/schema creation then from the before_script section.

The log exists in the container even before I start it and the log files has a previous entry with this string. I can check for a second record in the log file (from our `docker run`) but I'm not sure the additional complexity is actually required. The (pretty) Travis log file view becomes a bit of a mess too as the Oracle and DB2 pull/extract messages are interleaved.

I think I could leave these are they were for the time being (i.e. pull DB2, then create the DB and schema objects) then do the Oracle pull/extract stuff. What do you think?

build.gradle
MMatten: Light weight because of the provisioning or because of the tests that we're running?

javornikolov: It's the total time until the first test is done. It's a speculative guess so far - I don't really know whether db2 is slower or not. Maybe the order is just fine as it is.

javornikolov:
Why the build has failed now?

MMatten:
? Why the build has failed now?

I need to correct the character escaping in the command line.

MMatten:
Either I can't see the wood for the trees or there's something in the travis build script that's causing my docker command line to fail. Better options are i) a new shell script ii) go back to the original commit/PR.

build.gradle
MMatten: The DB2 tests are ready to run as soon at the `before_script` section has run. The complete build is taking 7 minutes to complete. I don't think moving the order will change that.

.travis.yml
```diff
@@ -5,13 +5,16 @@ services:
 
 language: java
 before_install:
+  - docker run --name dbfitdb2 -d -p 50000:50000 -v "`pwd`/test_vm/scripts/db2:/var/dbfit/test_vm/scripts/db2" -e DB2INST1_PASSWORD=db2inst1-pwd -e LICENSE=accept ibmcom/db2express-c:latest db2start &
```
javornikolov: I'm still wondering if it makes sense to have both `-d` and run it in background in the shell (via `&` suffix).

javornikolov: Even /var/dbfit/test_vm/scripts/db2 is quite long path. What about mounting it e.g. under `/tmp`?

build.gradle
javornikolov: It's more about how long it takes for a failing test to fail. But anyway - I guess the difference is negligible compared to the total time with the setup before that. OK to be as is.

MMatten:
@javornikolov I've done some tidying up and tried a few new moves (in order to get DB2 pulling in parallel with the Oracle pull). Would you mind doing another review please?

javornikolov:
> @javornikolov I've done some tidying up and tried a few new moves (in order to get DB2 pulling in parallel with the Oracle pull). Would you mind doing another review please?

I already took a look. It looks pretty good to me :+1: Just a minor remark about the paths length.

.travis.yml
MMatten: I was trying to avoid too many changes to the shell scripts to maintain compatibility when they're being used to install DB2 via vagrant. I'll have another look at the scripts.

javornikolov:
Hmm, one of the test runs failed - Oracle. I suspect we might have loaded the machine with many things and the timeout loop is not long enough - we may need to increase the number of polling steps.

MMatten:
> Hmm, one of the test runs failed - Oracle. I suspect we might have loaded the machine with many things and the timeout loop is not long enough - we may need to increase the number of polling steps.

That's the second time I've seen that today. Even by the time the tests ran it still wasn't available. ☹️ How long do we wait?

javornikolov:
> How long do we wait?

30 * 6 seconds - that is 3 minutes. Let's double it.

MMatten:
Ok. I've tidied up the path names now. I can't fully test that the vagrant VM build is still fully working because of #542

MMatten:
> I'm still wondering if it makes sense to have both -d and run it in background in the shell (via & suffix).

You mean perhaps we should just run everything in the foreground?

Sent from my iPhone

On 12 Feb 2017, at 23:15, Yavor Nikolov <notifications@github.com<mailto:notifications@github.com>> wrote:


@javornikolov commented on this pull request.

________________________________

In .travis.yml<https://github.com/dbfit/dbfit/pull/543#pullrequestreview-21420837>:

> @@ -5,13 +5,16 @@ services:

 language: java
 before_install:
+  - docker run --name dbfitdb2 -d -p 50000:50000 -v "`pwd`/test_vm/scripts/db2:/var/dbfit/test_vm/scripts/db2" -e DB2INST1_PASSWORD=db2inst1-pwd -e LICENSE=accept ibmcom/db2express-c:latest db2start &


I'm still wondering if it makes sense to have both -d and run it in background in the shell (via & suffix).

________________________________

In .travis.yml<https://github.com/dbfit/dbfit/pull/543#pullrequestreview-21420837>:

> @@ -5,13 +5,16 @@ services:

 language: java
 before_install:
+  - docker run --name dbfitdb2 -d -p 50000:50000 -v "`pwd`/test_vm/scripts/db2:/var/dbfit/test_vm/scripts/db2" -e DB2INST1_PASSWORD=db2inst1-pwd -e LICENSE=accept ibmcom/db2express-c:latest db2start &


Even /var/dbfit/test_vm/scripts/db2 is quite long path. What about mounting it e.g. under /tmp?

________________________________

In build.gradle<https://github.com/dbfit/dbfit/pull/543#pullrequestreview-21420837>:

> @@ -246,6 +246,7 @@ task fastbuild(dependsOn: [

 task travisbuild(dependsOn: [
     fastbuild,
+    ':dbfit-java:db2:integrationTest',


It's more about how long it takes for a failing test to fail. But anyway - I guess the difference is negligible compared to the total time with the setup before that. OK to be as is.

MMatten:
> Even /var/dbfit/test_vm/scripts/db2 is quite long path. What about mounting it e.g. under /tmp?

Now mounted at `/scripts`.

Sent from my iPhone

On 12 Feb 2017, at 23:15, Yavor Nikolov <notifications@github.com<mailto:notifications@github.com>> wrote:

Even /var/dbfit/test_vm/scripts/db2 is quite long path. What about mounting it e.g. under /tmp?

MMatten:
> Ok. I've tidied up the path names now. I can't fully test that the vagrant VM build is still fully working because of #542

I seem to have broken the vagrant build with the shell script changes. Another push required...

MMatten:
>> Ok. I've tidied up the path names now. I can't fully test that the vagrant VM build is still fully working because of #542

> I seem to have broken the vagrant build with the shell script changes. Another push required...

Ok. The vagrant VM build is now working again with this latest push.

MMatten:
> @javornikolov  javornikolov
I'm still wondering if it makes sense to have both -d and run it in background in the shell (via & suffix).

I'm not sure what you meant here. Just run it all sequentially and remove the checks for the services being up?

javornikolov:
>> @javornikolov javornikolov
>> I'm still wondering if it makes sense to have both -d and run it in background in the shell (via & suffix).
>
> I'm not sure what you meant here. Just run it all sequentially and remove the checks for the services being up?

@MMatten, sorry I didn't mean to to suggesting anything specific (maybe it's all fine). Just I'm noting that `-d` means to run the container in detached (background) mode and that `&` also means to run the command in background mode. So I just got puzzled about the _double background mode_. But it's possible that it's necessary (I'm not really aware in detail how it works).

MMatten:
> Just I'm noting that -d means to run the container in detached (background) mode and that & also means to run the command in background mode. 

Right. I see what you mean now. I had noticed that a while ago but forgot to try it without `&`. May be we should use `-d` for the Oracle XE container too. I'll try DB2 first.

MMatten:
> >  Just I'm noting that -d means to run the container in detached (background) mode and that & also means to run the command in background mode.

> Right. I see what you mean now. I had noticed that a while ago but forgot to try it without &. May be we should use -d for the Oracle XE container too. I'll try DB2 first.

It seems the difference in the behaviour, for the DB2 entry, with `-d` present and `&` removed, is that the `pull` and extract phase runs in the foreground, then the background service starts. It seems marginally slower (perhaps 30 seconds, perhaps less).

I'll try applying it to the Oracle container too now.

javornikolov:
I think when using `-d` the output is suppressed and it's not quite visible what happened. That's why I ended up using `&` for Oracle.

MMatten:
> I'll try applying it to the Oracle container too now.

Another few seconds is all it seems to add, perhaps we're only running one pull/extract at a time. I think the build log is tidier too.

What do you think of the options?

javornikolov:
> What do you think of the options?

What if we try both using `&` only without `-d`?

MMatten:
Ok. I'll give that combination a go. I think the pull / extract messages might make a mess of the raw log (but ok when looking at the pretty version with the collapsed sections).

MMatten:
> Ok. I'll give that combination a go. I think the pull / extract messages might make a mess of the raw log (but ok when looking at the pretty version with the collapsed sections).

That shaves about 30 seconds off of the time and the log (even the raw log) looks fine actually. Looks good to me. Any other points you've noticed?

javornikolov:
@MMatten, all looks good to me :+1: I'm merging it. Thanks a lot!

PR #547, opened 2017-02-25
Bump up gradle to 3.4

javornikolov:
One interesting new feature which might be interesting for us is [The Java Library Plugin](https://docs.gradle.org/3.4/userguide/java_library_plugin.html)


MMatten:
> One interesting new feature which might be interesting for us is The Java Library Plugin

So would we want to update the `build.gradle` config for the `core` as well right now?

javornikolov:
> So would we want to update the build.gradle config for the core as well right now?

We might do it if it simply works. But in separate PR (no need to complicate the simple version update here which seems to work OK and can be merged immediately).

Issue #483, opened 2016-01-13
DbFit API redesign


javornikolov:
This issue is to keep track of ideas which initially started emerging around the work of #463

 - [ ] `DBEnvironment` statement creation redesign #463
 - [ ] Improve packages design (e.g. currently there are objects with different purpose mixed in `util`). Consider pushing environment implementations in their own sub-packages (right now it's a common package and in theory a conflict may emerge, e.g. for some of the helper classes)
 - [ ] `EnvironmentFactory` - register with environment's class when scanning the annotations (instead of later relying on hard-coded class-name generation). That would permit moving environment classes to arbitrary package.
 - [ ] Extract interface `DbParameterAccessors` (to avoid coupling with the concrete implementation too much)
 - [ ] Rename `DBEnvironment` to `DbEnvironment` - as it's presently not consistent with the general conventions
 - [ ] Consider adding general plain-text statement creation method: in order to avoid abusing `createDdl*` for purposes other than DDL.

MMatten:
Related to moving the `buildUpdate` command - should we move the `DELETE` statement construction out of the `Clean` fixture?

javornikolov:
> Related to moving the `buildUpdate` command - should we move the `DELETE` statement construction out of the `Clean` fixture?

I believe yes. Added to the list.

javornikolov:
> move the DELETE statement construction out of the Clean fixture

Pushed to #463.

MMatten:
Regarding `DbParameterAccessor` naming:

> I'm not sure we can cay that a parameter is kind of `Attribute`.

I was trying to think of a generalised name for Stored Procedure Parameter or Table/View Column, but perhaps `Attribute` conflicts with well used Java naming. I guess if these handle `PreparedStatement` / `CallableStatement` parameters then perhaps the name of this class should be aligned with our wrapper classes for these JDBC types.

MMatten:
Is there anything that's in the `api` package that doesn't need to be there right now?

E.g. the `Diff` interface. 

Does this need to be here at the moment? There are no references to it within the other objects in the `api` package.

javornikolov:
In `CompareStoredQueries' we have
``` java
DataTableDiff diff = new DataTableDiff(loadRowStructure(lastRow));
```
which perhaps can be replaced with
``` java
Diff diff = new DataTableDiff(loadRowStructure(lastRow));
```
Nevertheless, perhaps it's better to move `Diff` to `diff` package.

[...]

PR #548, opened 2017-03-04
Refactor stored procedure execution

MMatten:
The main theme of this PR is to evaluate the effect of eliminating `DbStoredProcedureCall`.

Does this help or hinder the ease of understanding and maintaining the code base?


dbfit-java/core/src/main/java/dbfit/api/DBEnvironment.java
```diff
@@ -188,5 +188,6 @@ void connect(String dataSource, String username, String password,
     void connectUsingFile(String filePath) throws SQLException, IOException,
             FileNotFoundException;
 
-    DbStoredProcedureCall newStoredProcedureCall(String name, DbParameterAccessor[] accessors);
+    StatementExecution newStoredProcedureCall(String name, DbParameterAccessor[] accessors)
```
javornikolov: Here we're exposing `StatementExecution` which is quite bulky in terms of public methods. Just on this API level I'm not sure the change is good.  I would rather favor a leaner interface: could be as what emerged in #541 and in longer term - maybe something like #483)

dbfit-java/core/src/main/java/dbfit/api/DbStoredProcedure.java
```diff
@@ -21,7 +24,16 @@ public DbStoredProcedure(DBEnvironment environment, String name) {
...
-        return environment.newStoredProcedureCall(name, accessors).toStatementExecution();
+        DbParameterAccessors paramAccessors = new DbParameterAccessors(accessors);
+        PreparedStatement ps = environment.getConnection().prepareCall(toSqlString(paramAccessors));
+        StatementExecution cs;
+        if (paramAccessors.containsReturnValue()) {
+            cs = environment.createFunctionStatementExecution(ps);
+        } else {
+            cs = environment.createStatementExecution(ps);
+        }
+        bindParametersTo(paramAccessors, cs);
```
javornikolov: Perhaps we can do `accessors.bindParameters(cs)` directly here.

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
```diff
@@ -240,8 +240,9 @@ public String buildInsertCommand(String tableName,
         return sb.toString();
     }
 
-    public DbStoredProcedureCall newStoredProcedureCall(String name, DbParameterAccessor[] accessors) {
-        return new DbStoredProcedureCall(this, name, accessors);
+    public StatementExecution newStoredProcedureCall(String name, DbParameterAccessor[] accessors)
+            throws SQLException {
+        return new DbStoredProcedure(this, name).buildPreparedStatement(accessors);
```
javornikolov: `DbStoredProcedure` depends on the environment and the environment implementation here depends on `DbStoredProcedure`...

A more interesting point is that now there's nothing to call `newStoredProcedureCall`. If all works correctly in the current state (I'm not quite sure about Oracle since it's overriding this) - that method could rather be dropped.

MMatten:
> Here we're mixing this with stuff from the other PR and there are many changes to look at.

Yes. The intent here was to try to slice a piece off of #541 (the idea of dropping `DbStoredProcedureCall`) if if a good idea then to merge this first.

javornikolov:
> Yes. The intent here was to try to slice a piece off of #541 (the idea of dropping DbStoredProcedureCall) and if a good idea then to merge this first.

Yes, I understand. Maybe just a separate commit on top of #541 would be good enough.

As of now - I can see that we're breaking the Oracle implementation here.

MMatten:
> As of now - I can see that we're breaking the Oracle implementation here.

Yes. This might be a dead end. I'll take another look.

MMatten:
> DbStoredProcedure depends on the environment and the environment implementation here depends on DbStoredProcedure...

Well, this latest push should fix the broken Oracle build but we still have this interdependency.

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
```diff
@@ -240,8 +240,8 @@ public String buildInsertCommand(String tableName,
         return sb.toString();
     }
 
-    public DbStoredProcedureCall newStoredProcedureCall(String name, DbParameterAccessor[] accessors) {
-        return new DbStoredProcedureCall(this, name, accessors);
+    public DbStoredProcedure newDbStoredProcedure(String name) {
```
javornikolov: So we're passing the accessors at later stage... I'm not sure we want that. We're not supporting that yet but it's possible to have multiple routines with same name and differing in their parameters or return value only.
Besides, in `buildInsertPreparedStatement` we're using the approach with passing the accessors.

javornikolov:
@MMatten, by moving some of the internals about statement forming to `DbStoredProcedure` we're loading it with extra responsibility which I'm not sure it's good to belong there. `DbStoredProcedure` does the plumbing between `FitNesse` and the `database environment` but it's good to depend as less as possibles on internal implementation details of both.

Imagine we need to change something in procedure calls (like we do around hsqldb now) - e.g. add support to pass parameters by name instead of by position or something else. Then we're likely to need changes in *both* the db environment and in DbStoredProcedure.

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
MMatten: Why do you think we would we pass in accessors to the `DbStoredProcedure` constructor? It loads it's own accessors from database metadata.

MMatten:
> @MMatten, by moving some of the internals about statement forming to `DbStoredProcedure` we're loading it with extra responsibility which I'm not sure it's good to belong there. `DbStoredProcedure` does the plumbing between FitNesse and the database environment but it's good to depend as less as possibles on internal implementation details of both.

I agree about the extra responsibility. It's not always obvious to me exactly the function that each class serves. We have no formal definitions, only the code itself.

Perhaps we should remove the responsibility of forming `StatementExecution` objects from `DbTable` and `DbStoredProcedure`. Would that give us clearer class purposes?

> Imagine we need to change something in procedure calls (like we do around hsqldb now) - e.g. add support to pass parameters by name instead of by position or something else. Then we're likely to need changes in both the db environment and in `DbStoredProcedure`.

Agreed. It seemed to me that there was some overlap between `DbStoredProcedure` and `DbStoredProcedureCall` because of `buildPreparedStatement` and this was really being delegated to `DbStoredProcedureCall`.

javornikolov:
> Perhaps we should remove the responsibility of forming StatementExecution objects from DbTable and DbStoredProcedure. Would that give us clearer class purposes?

In #463 I tried something - instead of `StatementExecution` a `DbCommand` is formed which is much simpler (it's just an interface with `execute()` and `close()` methods).

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
javornikolov: > Why do you think we would we pass in accessors to the DbStoredProcedure constructor? It loads it's own accessors from database metadata.

If there are multiple procedures with same name - how would it know which one to pick from the database metadata? We don't support that yet - it's just a hint that in future we might need something more than the name. (To be honest - perhaps it's better to create a domain class instead of passing plain string and an array... So that part is anyway a candidate for change in future)

MMatten:
@javornikolov I'm going to close down this PR and delete the related branch unless you think there's anything useful in here (for future reference).

javornikolov:
> @javornikolov I'm going to close down this PR and delete the related branch unless you think there's anything useful in here (for future reference).

OK, you may close it. I guess we can reproduce it from scratch in future if needed. We may keep a local copy of the branch just in case :-)

PR #552, opened 2017-03-10
Introduce DbCommand interface to environment vendor API

MMatten:
Enhance environment vendor API by decoupling `DBEnvironment` interface from `StatementExecution` class.


MMatten:
To precede #541.

dbfit-java/core/src/main/java/dbfit/api/DbObject.java
```diff
@@ -1,12 +1,11 @@
...
 
-import dbfit.fixture.StatementExecution;
 import dbfit.util.DbParameterAccessor;
 import dbfit.util.Direction;
 
 import java.sql.SQLException;
 
 public interface DbObject {
-    public StatementExecution buildPreparedStatement(DbParameterAccessor accessors[]) throws SQLException ;
+    public DbCommand buildPreparedStatement(DbParameterAccessor accessors[]) throws SQLException;
```
javornikolov: Hmm, now I see that the method name is maybe not ideal. On the other hand we don't want go too deep here (as it may end up like #463).
It's good to look again and compare the HSQLDB changes - and see whether would be better to introduce `DbCommand` before or after it.

MMatten: How about renaming this to
```
public DbCommand buildCommand(DbParameterAccessor accessors[]) throws SQLException;
```
?

javornikolov: Maybe yes, it could be (as far as I see in #463 it was pretty similar). And still it's good to compare with the clean version of #541 without `DbCommand` - to see what we don't like in the case with less API changes.

MMatten: I'll update it to `buildDbCommand`, exactly as in #463.

I'll create another PR for #541, omitting the API changes too.

MMatten: > I'll update it to buildDbCommand, exactly as in #463.

Done.

MMatten:
I've just created #553. 

I still need to revert a few small changes and clean up the commits but it gives us a fair view of a PR without `DbCommand` or other earlier similar changes.

javornikolov:
> I've just created #553.
> 
> I still need to revert a few small changes and clean up the commits but it gives us a fair view of a PR without DbCommand or other earlier similar changes.

Great! To me something like #553 seems pretty reasonable and I guess we can go with it. The `DbCommand` and other improvement ideas should be possible to apply in future.

MMatten:
Great. I'm cleaning up the commits and reverting some of the other minor changes. I'll request a review once I've pushed those.

MMatten:
@javornikolov Do you think we should look now to incrementally apply #463 in small PRs like this (i.e. look to merge this one) one or look to complete, the large, #463 in totality?

javornikolov:
> @javornikolov Do you think we should look now to incrementally apply #463 in small PRs like this (i.e. look to merge this one) one or look to complete, the large, #463 in totality?

In general doing it incrementally sounds better (as soon as the chunks alone are meaningful). It may be better to release a new version first and to leave the API refactoring for the next one.

MMatten:
> It may be better to release a new version first and to leave the API refactoring for the next one.

Yes. I'm happy with that. A new version has been a long time coming!

PR #558, opened 2017-03-18
Add postgres to the Travis CI build

javornikolov:
Add postgresql to the Travis CI environment and include postgres integration tests to the Travis build.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -248,7 +248,8 @@ task travisbuild(dependsOn: [
     fastbuild,
     ':dbfit-java:db2:integrationTest',
     ':dbfit-java:mysql:integrationTest',
-    ':dbfit-java:oracle:integrationTest'
+    ':dbfit-java:oracle:integrationTest',
+    ':dbfit-java:postgres:integrationTest'
     ])
 
 if (!project.hasProperty('excludeIntegrationTests')) {
--- a/test_vm/scripts/postgresql/sql/create-db-schema-postgresql.sql
+++ b/test_vm/scripts/postgresql/sql/create-db-schema-postgresql.sql
@@ -1,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS changelog (
+    change_number INTEGER CONSTRAINT Pkchangelog PRIMARY KEY,
+    complete_dt TIMESTAMP NOT NULL,
+    applied_by VARCHAR(100) NOT NULL,
+    description VARCHAR(500) NOT NULL
+);
+
+ALTER TABLE changelog OWNER to dbfit;
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ sudo: required
 
 services:
   - docker
+  - postgresql
 
 language: java
 before_install:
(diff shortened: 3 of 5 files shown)


build.gradle
```diff
@@ -248,6 +248,7 @@ task travisbuild(dependsOn: [
     fastbuild,
     ':dbfit-java:db2:integrationTest',
     ':dbfit-java:mysql:integrationTest',
+    ':dbfit-java:postgres:integrationTest',
```
MMatten: Any particular reason for running before Oracle?

MMatten:
Looks good. Can we merge this before getting the tests passing though?

build.gradle
javornikolov: Not really a great reason. I'll change the order.

javornikolov:
> Looks good. Can we merge this before getting the tests passing though?

Good that you asked since I already got wondering about how to fix the tests. However I already have enough confidence that this PR is working fine (it fires up the postgresql tests) - perhaps no reason to wait further. What do you think?

test_vm/scripts/postgresql/sql/create-db-schema-postgresql.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS changelog (
```
javornikolov: Perhaps that script can be reused from the postgres Chef recipe to avoid duplication.

MMatten:
I think merging it allows us to address the broken tests more easily. So +1.

Sent from my iPhone

On 18 Mar 2017, at 15:19, Yavor Nikolov <notifications@github.com<mailto:notifications@github.com>> wrote:


Looks good. Can we merge this before getting the tests passing though?

Good that you asked since I already got wondering about how to fix the tests. However I already have enough confidence that this PR is working fine (it fires up the postgresql tests) - perhaps no reason to wait further. What do you think?

javornikolov:
I just did some refactoring and force pushed-it - to keep the schema objects setup in a single place ( 0f26a095b4abe714a7de6e7cda6f8aa1dbd77415)

MMatten:
As we've recently done for DB2 and Oracle, should we move other files from `dbfit/dbfit-java/postgres/src/integration-test/resources/` to under `test_vm`? One issue I've noticed with this is when running `gradlew eclipse` the projects include entries in the build paths that don't exist. It's another issue though. Perhaps we can leave an empty directory instead.

javornikolov:
`resources` are embedded in the jar files and is something which can be used at runtime. In general it seems cleaner to me to keep the scripts outside jars, yet it might be maybe relevant for the databases where we're using dbdeploy.
So I'm not quite sure about a universal guideline. We may keep reorganizing these on demand.

javornikolov:
Here b76b7cc1d1b61d2dded126c1b0af0f8f5a0c38f6 I committed an experiment with gradle `--continue` option. We may move it to another PR if we'd like to keep it. The intent is to see that other integration tests are green even in case of postgres failure here.

MMatten:
I think it's a good thing to have for the future. It could certainly speed up fixes to build failures.

Sent from my iPhone

On 18 Mar 2017, at 16:42, Yavor Nikolov <notifications@github.com<mailto:notifications@github.com>> wrote:


Here b76b7cc<https://github.com/dbfit/dbfit/commit/b76b7cc1d1b61d2dded126c1b0af0f8f5a0c38f6> I committed an experiment with gradle --continue option. We may move it to another PR if we'd like to keep it. The intent is to see that other integration tests are green even in case of postgres failure here.

Issue #571, opened 2017-04-27
Vagrant dev VM build hangs when installing Informix


MMatten:
Informix installation hangs at: -

```
==> default:               install_informix.sh: INFO: running ids_install utility...
```

javornikolov:
Resolved by #573

PR #573, opened 2017-05-01
Disable IPV6 Informix connections as problematic on vagrant CentOS image

MMatten:
Prevent `dbaccess` connections using IPv6 as problematic on vagrant CentOS image.


MMatten:
Resolves #571

PR #579, opened 2017-05-27
Fix failing Netezza symbols core tests

javornikolov:
Resolves #578 

* Parameterize CalcLength parameters in Symbols acceptance tests  
* Create CalcLength stored procedure for Netezza
* Implement support for Neteza functions return values

* Fix a small old typo in 001_add_objects_needed_for_acceptance_test.sql

PR #580, opened 2017-05-28
Deploy Netezza acceptance test objects via dbdeploy

MMatten:
To do: -
* [x] create the `NETEZZATEST` database (perhaps via a different dbdeploy invocation as it can't be in the same connection as for the acceptance test objects).
* [x] externalise the Netezza server hostname / IP address (perhaps via an environment variable so it can be shared by FitNesse (i.e. access it via a FitNesse variable) and gradle.
* [x] update documentation
* [x] complete and rebase onto #579 

Comments welcomed!


MMatten:
I think we need some way, other than dbdeploy, to create the `NETEZZATEST` database or the `NETEZZATEST.ADMIN.CHANGELOG`.

We can't get the Netezza client onto the dev VM from any public package repositories. @javornikolov Can we host it privately on S3? 

Alternatively we would need some other generic SQL/JDBC tool.

javornikolov:
> We can't get the Netezza client (so we could invoke the nzsql command interpreter) onto the dev VM from any public package repositories.

Why are we so concerned about the client? I mean - if the **server** is somehow already in place, doesn't it have a client included?

MMatten:
>> We can't get the Netezza client (so we could invoke the nzsql command interpreter) onto the dev VM from any public package repositories.

> Why are we so concerned about the client? I mean - if the server is somehow already in place, doesn't it have a client included?

The Netezza server is running on a VMWare machine. I need a way of creating the `NETEZZATEST` database and creating the `CHANGELOG` table inside that database. I can't use `dbdeploy` for this (because the `CHANGELOG` table won't exist).

I was thinking of using the official Netezza client. We we could use JDBC but we'd have to create a utility.

javornikolov:
Isn't it possible to do this setup within the Netezza VMware machine itself?

javornikolov:
In general Liquibase is something which may have netezza support http://forum.liquibase.org/topic/making-liquibase-work-with-netezza. And I guess some generic JDBC may already exist if needed to run a random sql against the database.

MMatten:
> Isn't it possible to do this setup within the Netezza VMware machine itself?

Yes. It's a manual task though. I had to transfer the SQL scripts to the Netezza host. It just makes this regression testing harder.

MMatten:
The Ant SQL task looks pretty useful: -

https://ant.apache.org/manual/Tasks/sql.html

I'll try this out.

MMatten:
I've added an Ant task to the Netezza project, only.

I can set up the basic DB now using `./gradlew :dbfit-java:netezza:createTestDatabase

What do you think of this?

javornikolov:
It looks pretty good to me :+1:
Is there anything more to do as part of this PR or we rather merge it?

MMatten:
I still want to allow for the variable Netezza VMWare guest having a variable IP address. Maybe `gradle.properties` would be a good place to configure this. Any thoughts? I'd want to make it available to FitNesse too though for the `SetUp` pages in each suite.

Should we extend the current gradle task approach to all sub-projects, not just Netezza, even if we don't use them at the moment?

I still have docs to update once we've decided the above.

javornikolov:
Currently for some of the environments we can configure custom connection details for acceptance tests via `*.properties.custom`. Maybe we can do something similar for Netezza and somehow reuse that file also for the initial setup.

MMatten:
> Currently for some of the environments we can configure custom connection details for acceptance tests via *.properties.custom. Maybe we can do something similar for Netezza and somehow reuse that file also for the initial setup.

Can you point me at an existing example if you can?

javornikolov:
In `dbfit-java/build.gradle`:
``` gradle
    integrationTest.doFirst {
        copy {
            from 'TestDbConnectionDbFitOracle.properties.default'
            into '.'
            rename '(.*).default', '$1'
        }

        copy {
            from 'TestDbConnectionDbFitOracle.properties.custom'
            into '.'
            rename '(.*).custom', '$1'
        }
    }
```

The .custom file is not under version control so if you have local preferences - you create `TestDbConnectionDbFitOracle.properties.custom` and it will overwrite the defaults.

dbfit-java/build.gradle
```diff
@@ -290,10 +290,71 @@ project('teradata') {
...
+            file(propertiesFile).withInputStream { props.load(it) }
+            userName =  props.getProperty('username')
+            passWord =  props.getProperty('password')
+            def service = props.getProperty('service')
+            def serviceParts = service.tokenize(':')
+            def serviceHost = serviceParts[0]
+            def servicePort = serviceParts[1]
+            def propDatabase = props.getProperty('database')
+
+            database = propDatabase.length() > 0 ? ('/' + propDatabase) : ''
```
javornikolov: I guess we can check `.isEmpty()` instead of length > 0

dbfit-java/build.gradle
```diff
@@ -290,10 +290,71 @@ project('teradata') {
...
+                userid: userName,
+                password: passWord,
+                delimiterType: delimiterType,
+                autocommit: true,
+                src: "${rootProject.projectDir}/test_vm/scripts/netezza/sql/create-databases-netezza.sql"
+            )
+        }
+    }
+
+    createTestDatabase.dependsOn setConnectionProperties
```
javornikolov: Would we have the `.properties` file generated in case we run just `createTestDatabase` task?

dbfit-java/build.gradle
```diff
@@ -290,10 +290,71 @@ project('teradata') {
 project('netezza') {
     description = 'DBFit netezza engine'
 
+    def propertiesFile = 'TestDbConnectionDbFitNetezza.properties'
+    def jdbcDriverClass = 'org.netezza.Driver'
+    def jdbcUrlPrefix = 'jdbc:netezza://'
+    def jdbcUrl
```
javornikolov: I'm not quite sure about these kind of global variables (I'm not quite sure how many times and when are these initialized).

dbfit-java/build.gradle
MMatten: Yes. Applied and force pushed.

dbfit-java/build.gradle
MMatten: I'm not sure what you mean. I've just added the `TestDbConnectionDbFitNetezza.properties` file.

We will need this set with the connection details for us to run the `createTestDatabase` task (now renamed to `setupTestDatabase`).

dbfit-java/build.gradle
MMatten: Ok. Do any other approaches come to mind?

MMatten:
Whitespace corrections applied and pushed.

MMatten:
Documentation updates added.

MMatten:
Acceptance Test connections now updated to use properties file.

dbfit-java/build.gradle
javornikolov: I don't   have a precise answer. One thought is to have a function which returns all details.

dbfit-java/build.gradle
javornikolov: Ah, I see now that we have `TestDbConnectionDbFitNetezza.properties`. The approach for other databases has been to have a `*.properties.default` file instead of `*.properties` since the latter could be replaced by `*.custom` version during the build and I don't quite like touching files which are under version control.

javornikolov: When we run the `setupTestDatabase` and if we're using custom database address (i.e. we have `*.custom` properties file) - would we want to connect via the default or via the custom settings?

dbfit-java/build.gradle
javornikolov: Here is one more idea - set these properties on task level:

``` gradle
task databaseConnectionDetails {
    doLast {
        databaseConnectionDetails.ext.url = "my-url"
    }
}

task mytask(dependsOn: databaseConnectionDetails) {
    doLast {
        println "DB url is: ${databaseConnectionDetails.url}"
    }
}
```

MMatten:
I've just pushed another version that adds the properties to the extra properties of the `setConnectionProperties` task.

There's some more tidying up to do but here's a view of what it might look like.

NETEZZA.md
```diff
@@ -28,12 +28,17 @@ It consists of 2 VMware images which theoretically could be run on any VMware en
    * Test if the Netezza Emulator is visible from the development vm (example: $ping 192.168.85.128)
    * Optionally install the linux client tools on the VM.
 
-3. Create the database NETEZZATEST using either the Administrator GUI or nzsql 
+3. Configure the Netezza connection for the DbFit acceptance tests
+   Set up the contents of the 'service=' key in the file 'dbfit-java\TestDbConnectionDbFitNetezza.properties'
```
javornikolov: Maybe we need refer to the `.custom` properties file here?

javornikolov:
> I've just pushed another version that adds the properties to the extra properties of the setConnectionProperties task.
> 
> There's some more tidying up to do but here's a view of what it might look like.

Yes, it could be something like that. When reading the properties I think the `.ext.` part is not needed. Also I'm not sure how'd be best to name that task but having `set*` in the name sounds a bit unnatural when reading the properties.

dbfit-java/build.gradle
```diff
@@ -294,6 +308,74 @@ project('netezza') {
...
+
+    task tearDownTestDatabase(dependsOn: setConnectionProperties) {
+        doLast {
+            sql(setConnectionProperties.ext.jdbcDriverClass,
+                setConnectionProperties.ext.jdbcUrl + '/system',
+                setConnectionProperties.ext.userName,
+                setConnectionProperties.ext.passWord,
+                "${rootProject.projectDir}/test_vm/scripts/netezza/sql/drop-databases-netezza.sql",
+                setConnectionProperties.ext.delimiterType, true
+            )
```
javornikolov: We don't have some prescribed style for groovy/gradle but maybe `)` is most often at the end of the previous line instead of alone on its own.

dbfit-java/build.gradle
```diff
@@ -294,6 +308,74 @@ project('netezza') {
...
+        copy {
+            from setConnectionProperties.ext.propertiesFile + '.custom'
+            into '.'
+            rename '(.*).custom', '$1'
+        }
+
+        dbdeploy(setConnectionProperties.ext.jdbcDriverClass,
+            setConnectionProperties.ext.jdbcUrl + setConnectionProperties.ext.database,
+            setConnectionProperties.ext.userName,
+	    setConnectionProperties.ext.passWord,
```
javornikolov: The `.ext` part here might need to be removed. Indentation needs fixes too.

MMatten: It seems to work OJ with it in. But as it's not needed I'll remove it and fix the whitespace.

dbfit-java/build.gradle
MMatten: I think I got that from the existing `ant.sql` function declaration but, yes, I'll change the new code.

MMatten:
> Also I'm not sure how'd be best to name that task but having set* in the name sounds a bit unnatural when reading the properties.

Yes. It's strange having a task (verb) with the name of a noun though too.

MMatten:
I've just force pushed some changes to rationalise some of the code changes.

@javornikolov can you let me know what you think of these?

NETEZZA.md
MMatten: Agreed. Done.

NETEZZA.md
```diff
@@ -28,12 +28,19 @@ It consists of 2 VMware images which theoretically could be run on any VMware en
    * Test if the Netezza Emulator is visible from the development vm (example: $ping 192.168.85.128)
    * Optionally install the linux client tools on the VM.
 
-3. Create the database NETEZZATEST using either the Administrator GUI or nzsql 
+3. Configure the Netezza connection for the DbFit acceptance tests in the file 'dbfit-java\TestDbConnectionDbFitNetezza.properties.custom':
```
javornikolov: What do you think about using ` instead of ' quotes here for sake of formatting enhancement?

NETEZZA.md
```diff
@@ -28,12 +28,19 @@ It consists of 2 VMware images which theoretically could be run on any VMware en
...
+   * Set the user name in the 'username=' key.
+   * Set the password in the 'password=' key.
+   * Set the hostname and the port number (format is <hostname>:<port>) in the 'service=' key.
 
-4. Run 001_add_objects_needed_for_acceptance_test.sql to create the objects needed for the coretests
+4. Create the database NETEZZATEST and acceptance test database objects either by: -
+   a. Running the command './gradlew :dbfit-java-netezza:setUpTestDatabase'
```
javornikolov: I'm not sure these `a.`, `b.` points would be rendered as intended. What's that `-` sign on the line before before `a.`?

dbfit-java/build.gradle
```diff
@@ -294,6 +308,76 @@ project('netezza') {
...
+    integrationTest.dependsOn connectionProperties
+
+    integrationTest.doFirst {
+        copy {
+            from connectionProperties.propertiesFile + '.default'
+            into '.'
+            rename '(.*).default', '$1'
+        }
+
+        copy {
```
javornikolov: We already have similar `copy` in `copyConnections` I think for the same property file. I'm wondering can we avoid that duplication. Not necessarily but an option might be not creating a copy of the file but just choosing where to read the properties from.

dbfit-java/build.gradle
```diff
@@ -294,6 +308,76 @@ project('netezza') {
...
+            file(propertiesFile).withInputStream { props.load(it) }
+            ext.userName = props.getProperty('username')
+            ext.passWord = props.getProperty('password')
+            def propService = props.getProperty('service')
+            def propServiceParts = propService.tokenize(':')
+            def propServiceHost = propServiceParts[0]
+            def propServicePort = propServiceParts[1]
+            def propDatabase = props.getProperty('database')
+
+            ext.database = propDatabase.isEmpty() ? ('/' + propDatabase) : ''
```
javornikolov: Grouping multiple properties in an `ext { }` block seems nice. What about applying it for these here too?

javornikolov:
@MMatten - it looks better now. I just added a few more remarks.

javornikolov:
Thanks @MMatten. It looks better now. I just added a few more comments.

NETEZZA.md
MMatten: > What's that - sign on the line before before a.?

That's just the way I write!

Now updated.

NETEZZA.md
MMatten: Good idea. Done

dbfit-java/build.gradle
MMatten: Yes. I like this. I'll extend it to the others. One more push to come.

dbfit-java/build.gradle
MMatten: I have an issue with the `integrationTest` task as it's trying to set up the connection properties (as it `dependsOn` the `connectionProperties` task`) but the `integrationTest.doFirst` hasn't copied the file by this point.

NETEZZA.md
javornikolov: Ah, OK :-)

dbfit-java/build.gradle
MMatten: I guess I should move the file copies to a separate task and make the `connectionProperties` one dependant on it.

javornikolov: I just realized we're already doing something with the other databases. Maybe it's a more general thing to consider.

javornikolov: What about let `connectionProperties` not rely on having `.properties` copied at all.

MMatten: I guess we'd prefer a `.custom` over the `.default` then.

I've just pushed a version with the `dependsOn`. Let me know how it looks to you.

dbfit-java/build.gradle
```diff
@@ -294,6 +308,83 @@ project('netezza') {
...
+
+            def props = new Properties()
+            file(connectionPropertyFilename.propertiesFile).withInputStream { props.load(it) }
+            def propService = props.getProperty('service')
+            def propServiceParts = propService.tokenize(':')
+            def propServiceHost = propServiceParts[0]
+            def propServicePort = propServiceParts[1]
+            def propDatabase = props.getProperty('database')
+
+            ext {
```
javornikolov: Do we need to have the two `ext` blocks separate?

MMatten: I guess not as the `def` statement don't result in an extra property being created. I was trying to be very explicit about which ones were creating properties.

I'll try merging them.

MMatten: Done. Better. Thanks.

dbfit-java/build.gradle
```diff
@@ -294,6 +308,80 @@ project('netezza') {
         compile project(':dbfit-java:core')
         runtime 'org.netezza:nzjdbc3:7.2@jar'
     }
+
+    task connectionPropertyFilename {
+        ext.propertiesFile = 'TestDbConnectionDbFitNetezza.properties'
```
javornikolov: That looks like constant. What do you think about just declaring it as `ext.connectionPropertyFilename = ...`?

MMatten: You mean as an extra property of the project?

javornikolov: Yes.

MMatten: Looks good. I'll implement that.

MMatten:
Now rebased onto `master`. @javornikolov would you please review again?

javornikolov:
It looks good :+1: I'm merging it...

PR #586, opened 2017-06-10
Generalize integration test connection properties gradle tasks & gitignore

javornikolov:
* Generalize integration tests properties files Gradle copy tasks and .gitignore
* Effectively the new .gitignire setup takes into account Netezza too

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,9 +46,5 @@ build
 gradle.properties
 target
 dbfit-complete-*.zip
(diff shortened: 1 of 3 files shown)


javornikolov:
@MMatten, I did some refactoring around the connection properties. May we test the integration tests and the netezza db setup tasks.

MMatten:
> May we test the integration tests and the netezza db setup tasks.

All tests passing 👍

PR #591, opened 2017-06-24
Use dbdeploy for Teradata acceptance test object creation

MMatten:
- Use dbdeploy for Teradata acceptance test object creation.
- Align with approach used for Netezza.


MMatten:
@javornikolov We seem to have a problem with the Java 7 build. Have you any idea what's happening?

javornikolov:
> @javornikolov We seem to have a problem with the Java 7 build. Have you any idea what's happening?

```
$ jdk_switcher use oraclejdk7
Switching to Oracle JDK7 (java-7-oracle), JAVA_HOME will be set to /usr/lib/jvm/java-7-oracle
update-java-alternatives: directory does not exist: /usr/lib/jvm/java-7-oracle
```

Same issue seems to be reported here: https://github.com/travis-ci/travis-ci/issues/7964

MMatten:
@javornikolov Thanks for that. I guess some options are:-

* Use Open JDK 7 (still available from yum repos).
* Host our own Oracle JDK 7 installer.
* Abandon Java 7 support (at least in the CI).

I think the OpenJDK option sounds best. What do you think?

javornikolov:
openjdk has other problems: https://travis-ci.org/dbfit/dbfit/jobs/246714503

MMatten:
Yes, I've just witnessed the same.

MMatten:
Well, OpenJDK8 seems fine.

For JDK7 it looks like we're limited to hosting our own (the S3 site) or dropping Java 7 in the CI.

MMatten:
I'll log an issue with Travis-CI anyway.

MMatten:
After doing some testing an another Java project I'm not even sure what's supported under which configuration now. OpenJDK8 doesn't seem available under certain circumstances (e.g. https://travis-ci.org/MMatten/dbdeploy).

javornikolov:
For Java 8 I think we can safely use OracleJDK. The main problem is with Java 7.

For some reason running the build on precise environment (instead of trusty) seems to help: https://travis-ci.org/dbfit/dbfit/builds/246760268

MMatten:
> For some reason running the build on precise environment (instead of trusty) seems to help: https://travis-ci.org/dbfit/dbfit/builds/246760268

That build doesn't look well 😨

javornikolov:
Well, switching to precise makes java work but then docker is broken...

MMatten:
Rebased onto #593 now.

FitNesseRoot/DbFit/AcceptanceTests/JavaTests/TeradataTests/FlowMode/SetUp/content.txt
```diff
@@ -1,7 +1,9 @@
 !|dbfit.TeradataTest|
 
+|Connect Using File|!-TestDbConnectionDbFitTeradata.properties-!|
```
javornikolov: Do we need to wrap the filename in `!- -!` here?

dbfit-java/teradata/src/integration-test/resources/001_acceptance_test_objects-teradata.sql
```diff
@@ -0,0 +1,95 @@
...
+    SET concatenated = firstString || ' ' || secondString;
+END;
+;
+
+ET
+;
+
+REPLACE PROCEDURE nulls_back(OUT out1 VARCHAR, OUT out2 VARCHAR, OUT out3 VARCHAR)
+BEGIN
+   SET out1 = '1';
```
javornikolov: Identation is not consistent across the different blocks.

dbfit-java/core/src/main/java/dbfit/util/DbConnectionProperties.java
```diff
@@ -26,18 +26,28 @@ private static DbConnectionProperties createFrom(Map<String, String> map) {
...
+                case "service":
+                    props.Service = val;
+                    break;
+                case "database":
+                    props.DbName = val;
+                    break;
+                case "connection-string":
+                    props.FullConnectionString = val;
+                    break;
+                case "admin-username":
```
javornikolov: Hmm, we're changing the production code here and I'm wondering - would that be of any use for anything else than the integration test bootstrap here?

Another question is: if we want to use a different set of username/password - should we keep it in the same properties file?

dbfit-java/build.gradle
```diff
@@ -83,7 +83,10 @@ subprojects {
         include('**/*Support*')
     }
 
-    task integrationTest(type: Test) {
+    task dbConnectionProperties {
```
javornikolov: Now that we have both `dbConnectionProperties` and `connectionProperties` - I'm wondering is it clear enough from the name what's the difference between them?

dbfit-java/build.gradle
```diff
@@ -83,7 +83,10 @@ subprojects {
...
+    }
+
+    task integrationTest(type: Test, dependsOn: dbConnectionProperties) {
```
javornikolov: IntegrationTest depends on dbConnectionProperties which includes some admin user/pass... I'm not sure I'm following the whole chain correctly but do we really need admin account to run the integration tests?

FitNesseRoot/DbFit/AcceptanceTests/JavaTests/TeradataTests/FlowMode/SetUp/content.txt
MMatten: I like to keep the *graceful* fixture names (spaces). I've not tried `!|Connect Using File|`. I suspect it won't work. I'll try it.

javornikolov: I was about `|!-TestDbConnectionDbFitTeradata.properties-!|` vs `|TestDbConnectionDbFitTeradata.properties|`.

dbfit-java/core/src/main/java/dbfit/util/DbConnectionProperties.java
MMatten: Good questions. This is the area I doubted most in this PR.

I guess another properties file for DBA credentials could be used with the only downside that we'd need to likely duplicate the database service and database names.

dbfit-java/build.gradle
MMatten: Not really. `connectionProperties` are now really just the values from the properties file. May be renaming `connectionProperties` to `connectionFileProperties` would be enough to distinguish them.

dbfit-java/build.gradle
MMatten: Good point. The sets of properties can be split into different parents / containers.

FitNesseRoot/DbFit/AcceptanceTests/JavaTests/TeradataTests/FlowMode/SetUp/content.txt
MMatten: I'll try that again. I think I was getting an error in FitNesse because of the CamelCase (FitNesse thinks it's a reference to another page).

dbfit-java/core/src/main/java/dbfit/util/DbConnectionProperties.java
javornikolov: Possibly we may read the database names from the other file when dealing for the DBA operations.

FitNesseRoot/DbFit/AcceptanceTests/JavaTests/TeradataTests/FlowMode/SetUp/content.txt
javornikolov: Oracle is using the style without `!--!`, in other places we're using the other style... I'm not sure which is better. The initial tought is that there are some caveats when we're using `!--!` - e.g. when parsing `<`, `>` signs - I guess because that goes to the old Fit classes embedded into FitLibrary instead of the newer FitNesse ones.

MMatten: FitNesse presents the word like this: -

image

The URL has the suffix `?edit&nonExistant=true`.

MMatten: FitNesse also passes the modified word through to Fit:-

image

MMatten:
@javornikolov I've just force-pushed another version for review.

dbfit-java/build.gradle
```diff
@@ -83,7 +83,80 @@ subprojects {
...
+                delimeterType = 'normal'
+                delimeter = ';'
+                setUpTestDbScript = "${rootProject.projectDir}/test_vm/scripts/${project.name}/sql/create-databases-${project.name}.sql"
+                tearDownTestDbScript = "${rootProject.projectDir}/test_vm/scripts/${project.name}/sql/drop-databases-${project.name}.sql"
+            }
+        }
+    }
+
+    // to be overriden when needed
+    ext.connectionPropertiesFile = '/dev/null'
```
javornikolov: Is it needed and safe to have default file here? What would happen if it's not overriden? (I'm not sure whether `/dev/null` is available on Windows).

dbfit-java/build.gradle
```diff
@@ -83,7 +83,80 @@ subprojects {
...
+    ext.connectionPropertiesFiles = [connectionPropertiesFile]
+
+    task connectionFileProperties {
+        doLast {
+            copyConnectionProperties project.connectionPropertiesFiles, '.'
+            def props = new Properties()
+            file(connectionPropertiesFile).withInputStream { props.load(it) }
+
+            ext {
+                userName = props.getProperty('username')
```
javornikolov: Looks like props can be also accessed via alternative syntaxes, e.g.: `props['username']` or `props.username`.

Another wondering is how would it look like if we wrap these in a single ext property of type *map* instead of separate independent string values. (That may make some param lists and assignments shorter. But I'm not sure about all side effects).

dbfit-java/build.gradle
MMatten: @javornikolov I'm currently reworking some of this stuff as I've realised this too since I've pushed this. Some of this stuff isn't very elegant and I'll spin off another branch to review an alternative.

MMatten:
@javornikolov I've just created #598 which should be a little nicer. I've still now yet tried seeing what usage of `Map` would look like but any review comments as it stands now would be appreciated.

MMatten:
Superseded by #598.

MMatten:
Superseded.

PR #593, opened 2017-06-28
Gradle wrapper - use http instead of https

### LINKED ISSUE #592: Java 7 build fails on Travis-CI (missing OracleJDK7 on Ubuntu image)
Transfer of issues from #591

javornikolov:
There is a problem with java 7 when using https so here we workaround it by tweaking the gradle download url.

resolves #592

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ before_script:
   - docker exec dbfitdb2 /scripts/create_db_schema.sh
 script: "./gradlew --continue clean travisbuild"
 jdk:
-- oraclejdk7
+- openjdk7
 - oraclejdk8
 after_success:
 - "./gradlew bundle"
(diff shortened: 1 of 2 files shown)


MMatten:
Do we no longer need to switch to OpenJDK for this PR?

javornikolov:
> Do we no longer need to switch to OpenJDK for this PR?

Actually we do... I just force-pushed to add that too.

MMatten:
How odd:

```
Broadcast message from root@testing-gce-f32bbfb0-1723-42e5-9de6-d687a10839e2
	(unknown) at 20:46 ...

The system is going down for power off NOW!
```

javornikolov:
Strange... I spotted that but I wasn't sure is it a real shutdown and would it somehow resume after it. But it seems stuck to me so I'm going to cancel and restart the build.

PR #598, opened 2017-07-31
Use dbdeploy for Teradata and Netezza

MMatten:
Use dbdeploy for test database setup and teardown and for creating acceptance test objects.


javornikolov:
Thank you, @MMatten! I'll take a look.

dbfit-java/build.gradle
```diff
@@ -303,65 +430,39 @@ project('netezza') {
...
-            connectionProperties.passWord,
-            scriptFile,
-            connectionProperties.delimiterType, ';', true)
-    }
-
-    task setUpTestDatabase(dependsOn: connectionProperties) {
-        doLast {
-            execSqlScriptTestDatabase(
-                "${rootProject.projectDir}/test_vm/scripts/netezza/sql/create-databases-netezza.sql")
+// can we add these defs to all subprojects (and also remove from TD and NZ)?
```
javornikolov: What would be the intent behind that? Is it about avoiding that copy of `ext.jdbcUrl = ...` in several places?

dbfit-java/build.gradle
```diff
@@ -83,7 +83,70 @@ subprojects {
...
+    task dbConnectionProperties(dependsOn: [deploymentProperties]) {
+    }
+
+    task adminConnectionFileProperties {
+        doLast {
+            def adminConnPropsFile
+            if (project.hasProperty('adminConnectionPropertiesFile')) {
+                adminConnPropsFile = adminConnectionPropertiesFile
+            } else {
+                adminConnPropsFile = connectionPropertiesFiles[0]
```
javornikolov: Do we have such case when admin connection is needed but we don't have a specific admin connection properties file?

MMatten: Not at the moment. If you think we should remove this condition then I'm fine with that.

dbfit-java/build.gradle
```diff
@@ -134,16 +204,39 @@ subprojects {
...
-
     ext.copyConnectionProperties = {
         destination ->
             project.connectionPropertiesFiles.each { file ->
                 copyProperties new File(project.projectDir, file), destination
             }
     }
 
+    ext.execSqlScriptTestDatabase = {
+        scriptFile ->
```
MMatten: Is there a well known standard indentation for methods/params/bodies?

dbfit-java/build.gradle
MMatten: > Is it about avoiding that copy of ext.jdbcUrl = ... in several places?

Yes that's right. But not all subprojects will have executed the tasks `connectionFileProperties` and `adminConnectionFileProperties` hence we'll get a validation (compilation-like) error when these properties are not available.

```
    dbConnectionProperties.doLast {
        ext {
            jdbcUrl =
```
and
```
    dbAdminConnectionProperties.doLast {
```

dbfit-java/build.gradle
javornikolov: I'm not sure is there anything well known. I think for closures it may be fine to do something like:
``` groovy
ext.execSqlScriptTestDatabase = { scriptFile ->
    ...
}
```

If it's longer list of parameters - something like what we've done so far seems OK to me.

dbfit-java/build.gradle
javornikolov: If we're not really using it - seems an extra complication for now.

dbfit-java/build.gradle
```diff
@@ -293,6 +386,40 @@ project('teradata') {
...
+
+    def getJdbcUrl = {
+        service, database ->
+            return deploymentProperties.jdbcDriverUrlPrefix + service +
+                '/TMODE=TERA,DATABASE=' + database
+    }
+
+    dbConnectionProperties.doLast {
+        ext {
+            jdbcUrl = getJdbcUrl(connectionFileProperties.service, connectionFileProperties.database)
```
javornikolov: When setting just a single property it's a bit more concise to have just `ext.jdbcUrl`.

dbfit-java/build.gradle
MMatten: Agreed. I'll chop it out.

dbfit-java/build.gradle
MMatten: Agreed. Good spot.

dbfit-java/build.gradle
javornikolov: What options do we have? So far I can think of:
* What you suggested - to have these properties always defined (with default of null or something)
* `ext.jdbcUrl = { getJdbcUrl(...) }`. That would need a pair of parentheses to get evaluated `jdbcUrl()`. I might be getting the syntax wrong but the general idea is to have that evaluated lazily.

I'm also wondering about defining `getJdbcUrl` as closure vs something with standard parameter list like `def getJdbcUrl(service, database)`. I suspect it would still fail to resolve the dependent properties (maybe even earlier), just I got curious what's the difference since I'm not sure.

dbfit-java/build.gradle
MMatten: We have a mixture of two styles at the moment. E.g.

```
    ext.copyProperties = {
        file, destination ->
            copy {
                from "${file}.default"
                into destination
                rename "(.*)\\.default", '$1'
            }
            copy {
                from "${file}.custom"
                into destination
                rename "(.*)\\.custom", '$1'
            }
    }
```
and
```
   ext.dbdeploy = {
        dbDriver, url, userid, password, delimiterType = 'normal', delimiter = ';',
        dir = "$projectDir/src/integration-test/resources" ->
        ant.taskdef(name: 'dbdeploy',
    ...
```
and
```
    def getJdbcUrl = {
        service, database ->
            return deploymentProperties.jdbcDriverUrlPrefix + service +
```

MMatten: If we're not fussed which we use right now then we can just log another issue to standardise later.

javornikolov: I vote for the one with the indentation. And also for moving the parameters on top next to `{` unless they're too many.

javornikolov: > If we're not fussed which we use right now then we can just log another issue to standardise later.

OK. The changes related to this PR we may directly format as we want. And whatever else needs reformatting - could be done later.

MMatten:
> What options do we have? So far I can think of:

> What you suggested - to have these properties always defined (with default of null or something)
> * ext.jdbcUrl = { getJdbcUrl(...) }. That would need a pair of parentheses to get evaluated jdbcUrl(). I might be getting the syntax wrong but the general idea is to have that evaluated lazily.

> I'm also wondering about defining getJdbcUrl as closure vs something with standard parameter list like def getJdbcUrl(service, database). I suspect it would still fail to resolve the dependent properties (maybe even earlier), just I got curious what's the difference since I'm not sure.

I've tried the options listed here:-
http://mrhaki.blogspot.co.uk/2016/09/gradle-goodness-lazy-project-property.html
and none seem to enable us to avoid the parsing/compilation error. I think the structures need to be present and it's only the values that can be evaluated just in time.

The only other, sledge hammer, approach I've been thinking of is to simply use connection properties files for all of the database types.

javornikolov:
>> What options do we have? So far I can think of:
>> What you suggested - to have these properties always defined (with default of null or something)
>> 
>> ext.jdbcUrl = { getJdbcUrl(...) }. That would need a pair of parentheses to get evaluated jdbcUrl(). I might be getting the syntax wrong but the general idea is to have that evaluated lazily.
>> I'm also wondering about defining getJdbcUrl as closure vs something with standard parameter list like def getJdbcUrl(service, database). I suspect it would still fail to resolve the dependent properties (maybe even earlier), just I got curious what's the difference since I'm not sure.

> I've tried the options listed here:-
> http://mrhaki.blogspot.co.uk/2016/09/gradle-goodness-lazy-project-property.html
> and none seem to enable us to avoid the parsing/compilation error. I think the structures need to be present and it's only the values that can be evaluated just in time.
> 
> The only other, sledge hammer, approach I've been thinking of is to simply use connection properties files for all of the database types.

We already have `ext.connectionPropertiesFiles = []` defined at `allprojects` level. Adding a few more perhaps should work fine.

MMatten:
> We already have ext.connectionPropertiesFiles = [] defined at allprojects level. Adding a few more perhaps should work fine.

A few more added...

MMatten:
Some docker failure occurred. Restarted and completed successfully.

dbfit-java/build.gradle
```diff
@@ -83,7 +83,75 @@ subprojects {
...
+    task connectionFileProperties {
+        doLast {
+            copyConnectionProperties '.'
+            ext {
+                userName = ''
+                passWord = ''
+                service = ''
+                database = ''
+            }
+            if (connectionPropertiesFiles.size() > 0) {
```
javornikolov: It should be enough `if (connectionPropertiesFiles) {` here.

dbfit-java/build.gradle
```diff
@@ -83,7 +83,75 @@ subprojects {
...
+                def props = new Properties()
+                file(connectionPropertiesFiles[0]).withInputStream { props.load(it) }
```
javornikolov: We have that props loading from a file in several places. I'm wondering can we reuse the implementation.

dbfit-java/build.gradle
```diff
@@ -83,7 +83,75 @@ subprojects {
...
+                ext {
+                    userName = props.username
+                    passWord = props.password
+                    service = props.service
+                    database = props.database
```
javornikolov: We have such assignments (of multiple properties) in several places. But I'm not sure is it possible and how to extract it to a single place and still have the properties specific for the task.

dbfit-java/build.gradle
```diff
@@ -83,7 +83,75 @@ subprojects {
...
+    ext.connectionPropertiesFiles = []
+
+    task connectionFileProperties {
+        doLast {
+            copyConnectionProperties '.'
+            ext {
+                userName = ''
+                passWord = ''
+                service = ''
+                database = ''
```
javornikolov: Do we need to define these defaults explicitly? In any case - if needed, we already have assignments below: maybe they can be reused (out of empty properties map)

MMatten:
@javornikolov some more refinements just pushed. I think this addresses the latest review comments.

javornikolov:
@MMatten, sorry for delayed follow up. This looks good enough to me now. Unless you have plans for changes I think we can merge it?

MMatten:
@yavornikilov no problem at all. No more changes planned from me. Thanks for your invaluable input as usual 👍

MMatten:
Ah. One thing. A doc update would be good. I can do that first.

javornikolov:
OK, @MMatten. Waiting for the doc update. 10x :+1:

MMatten:
> Waiting for the doc update

Pushed.

javornikolov:
Thanks! It looks good. Merging :shipit:

PR #610, opened 2017-11-26
Update to SQL Server installer

MMatten:
Update installer to re-align with Microsoft config utility options.


javornikolov:
Do we want to merge that before the release? If that's the case - can we remove the d13c281 commit from here since it's not merged yet to master.

MMatten:
> Do we want to merge that before the release? If that's the case - can we remove the d13c281 commit from here since it's not merged yet to master.

Right. I'd just spotted that. Let me fix this.

MMatten:
@javornikolov Fixed.

javornikolov:
Thank you, @MMatten! I guess it works fine so I'm merging it and I'll rebase the release branch after that.

PR #622, opened 2018-02-06
Allow parameter name patterns to be varied from and restored to default values.

MMatten:
Allow parameter name pattern to be varied from and restored to the default name pattern string for each adapter via a new `Parameter Pattern` option.


dbfit-java/netezza/src/main/java/dbfit/environment/NetezzaEnvironment.java
```diff
@@ -32,10 +33,14 @@ protected String getConnectionString(String dataSource, String database) {
...
-    private static Pattern paramsNames = Pattern.compile(paramNamePattern);
+
+	private String getParamPattern() {
+        return Options.get(Options.OPTION_PARAM_PATTERN).equals("") ?
+            paramNamePattern : Options.get(Options.OPTION_PARAM_PATTERN);
+    }
 
     public Pattern getParameterPattern() {
-        return paramsNames;
+    	return Pattern.compile(getParamPattern());
```
MMatten: Whitespace issue here noted.

javornikolov:
@MMatten I think something that should be OK.

Just we rather have diffent naming since it's hard to guess what's the difference between `getParameterPattern` and `getParamPattern`.

MMatten:
> Just we rather have diffent naming since it's hard to guess what's the difference between getParameterPattern and getParamPattern.

Very true. I was already thinking of that too.

javornikolov:
It's still a bit confusing. The distinguishing part I think is that one of these is the **default** pattern (for the environment), isn't it?

Another question is do we need any of these to be `public`?

javornikolov:
In general seems good to me. I guess we can merge it after polishing the naming.

MMatten:
@javornikolov I still have to test this over some of the adapters but any review comments at this stage would be great.

MMatten:
I'm not sure how my local build didn't pick up that unused import. Fixed and force pushed.

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
```diff
@@ -197,7 +197,17 @@ public Connection getConnection() throws SQLException {
         return hs.toArray(array);
     }
 
-    protected abstract Pattern getParameterPattern();
+    protected String defaultParamPatternString = "";
+
+    protected String getParamPatternString() {
```
javornikolov: Can we keep that method private (if we want a separate method at all)?

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
```diff
@@ -197,7 +197,17 @@ public Connection getConnection() throws SQLException {
...
+        return Options.get(Options.OPTION_PARAMETER_PATTERN).equals("") ?
+            defaultParamPatternString : Options.get(Options.OPTION_PARAMETER_PATTERN);
+    }
+
+    //protected abstract Pattern getParameterPattern();
```
javornikolov: Isn't that comment redundant? Maybe we rather refer to `defaultParamPatternString` attribute as a comment here?

dbfit-java/core/src/main/java/dbfit/api/AbstractDbEnvironment.java
javornikolov: Ah, I see we're using it in subclasses. I'm wondering about the `getParameterPattern().toString()` alternative. My point is that having too many methods which is possible to override might lead to some confusion when implementing a new adapter.

dbfit-java/db2/src/main/java/dbfit/environment/DB2iEnvironment.java
```diff
@@ -23,20 +22,15 @@
 
     public DB2iEnvironment(String driverClassName) {
         super(driverClassName);
+        defaultParamPatternString = "[@:]([A-Za-z0-9_]+)";
+
```
MMatten: I'll delete this extra blank line.

dbfit-java/hsqldb/src/main/java/dbfit/environment/HSQLDBEnvironment.java
```diff
@@ -57,12 +57,9 @@ protected String getConnectionString(String dataSource, String database) {
...
 
-    private static final String paramNamePattern = "@([A-Za-z0-9_]+)";
-    private static final Pattern paramRegex = Pattern.compile(paramNamePattern);
-
-    @Override
-    protected Pattern getParameterPattern() {
-        return paramRegex;
+    protected String parseCommandText(String commandText) {
+        commandText = commandText.replaceAll(getParamPatternString(), "?");
+        return super.parseCommandText(commandText);
```
MMatten: @javornikolov I'm thinking that now I've got this block in every adapter I should merge it into the super-class's method.

MMatten:
@javornikolov I've now tested the current PR state with all adapters.

MMatten:
@javornikolov I think I've tidied things up a little. What do you think about the current methods and visibility of them in `AbstractDbEnvironment`?

MMatten:
@javornikolov do you think this is close to being mergeable yet?

javornikolov:
@MMatten, sorry for the delay! Thank you for the reminder :-)
It looks OK to me though I have some wondering about about protected attributes vs methods: but it's theoretica for now. I'm merging this PR.

nohoyuar:
Hi Team,

I am trying to explore how can i connect to command line terminal and execute unix/shell commands.
Please help suggest me with a sample example on how to do it. I am exploring on this for past several days. I tried even with downloading the latest jar file commandlinefixture.jar but it is throwing error while i am using it in dbfit syntax. Can you elaborate on the steps to perform it. 
Thanks & regards,
Arun K

MMatten:
Use the full class name including the package name part or try |Import| of the package name.

Sent from my iPhone

On 14 Mar 2018, at 10:26, nohoyuar <notifications@github.com<mailto:notifications@github.com>> wrote:


Hi Team,

I am trying to explore how can i connect to command line terminal and execute unix/shell commands.
Please help suggest me with a sample example on how to do it. I am exploring on this for past several days. I tried even with downloading the latest jar file commandlinefixture.jar but it is throwing error while i am using it in dbfit syntax. Can you elaborate on the steps to perform it.
Thanks & regards,
Arun K

PR #627, opened 2018-08-01
Migrate away from deprecated database cookbook for PostgreSQL

MMatten:
The sous-chefs `database` cookbook has been deprecated and also the `postgresql` cookbook has been reworked to avoid the (usually broken) `pg_gem` gem.

Hence reworking the `postgres` cookbook to use the new components.


MMatten:
@javornikolov I'm not sure if you have time to review this. I'll wait a day or so and then merge.

PR #631, opened 2018-09-04
Prerequisite changes to Core for Sybase support

MMatten:
Preparation for Sybase IQ support and for other drivers that might have incomplete JDBC implementations.
* Use JDBC type when driver ResultSetMetaData does not provide the DB type for a column.
* Call appropriate setNull when no user defined type name.
* Allow setObject behaviours to be specified for a specific DB type.


MMatten:
@javornikolov any review would be useful if you're available otherwise I'll merge within a day or two.

PR #608, opened 2017-11-12
Release 4.0.0

javornikolov:
Release 4.0.0

TODO: prepare release notes

PR #649, opened 2020-07-25
Release/4.0.0 bhn

bhuesemann:
@javornikolov 

Here it is,. It is build on top of PR for 4.0.0 
Gradle update to 6.3, code style fixes and fixed java generics, openjdk 14 compatible chef upgrade, general cookbook upgrades and bugfixes, some database upgrades.


bhuesemann:
Don't know how to fix the travis error related to referencing the Netezza nzjdbc3 jar via Maven. I remember having a really hard time to make this up on the internet. Netezza is hardly used anymore although it was quite a nice (and I remember one of the first) shared nothing database in the past.

test_vm/Vagrantfile
```diff
@@ -16,16 +19,32 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
...
+  end
+  
   config.vm.synced_folder "..", "/var/dbfit", CONFIG['synced_folder_options']
 
   config.vm.provision :chef_solo do |chef|
-    chef.add_recipe "dbfit_test"
-
+    chef.add_recipe("dbfit_test")
+    chef.arguments = "--chef-license accept"
+    chef.run_list = ["recipe[firewalld::disable]", "recipe[iptables::disabled]",
```
javornikolov, line 38: Do we need to explicitly list recipes? dbfit_test `default.rb` already includes list of included recipes..., is it failing to do the job?

javornikolov:
@bhuesemann, I wonder whether other changes would work if we keep dbfit 4.0 java7 compatible? And jdk update we may include in a subsequent release.

dbfit-java/.gitignore
```diff
@@ -1,2 +1,3 @@
 download
 *.log
+*/bin
```
javornikolov, line 3: What is using this `/bin` folder? Is it auto-created by gradle or something else?

javornikolov:
@bhuesemann, regarding netezza and others - we have custom repo with 3rd party jars on s3 repo.
Looks like current master branch travis build works OK.

test_vm/Vagrantfile
bhuesemann: I agree with you this looks ugly, but I've had strange chef errors about missing cookbooks. Later I realized that this was due to some cookbooks that seem to be inkompatibel meanwhile an announced to be deprecated. I've more experience with puppet and this is the first time I've used chef so I'll try to revert this part to the original one. It is redundant to the includes...

dbfit-java/.gitignore
bhuesemann: Yes, it is autocreated and I wondered why this was not the case before.

test_vm/Vagrantfile
bhuesemann: I have reverted the enumeration in the runlist and it runs ok now.

dbfit-java/.gitignore
bhuesemann: Java 7 is out of support already. I'm not sure if we can make it compatible to Java 7. At least we should be be compatible to Java 11 which is the current LTS Java version as it seems (better would be Java 8).
[Java Longterm Support Roadmap](https://www.oracle.com/java/technologies/java-se-support-roadmap.html)
[OpenJDK Support](https://adoptopenjdk.net/support.html)

bhuesemann: I've added Java 8 to the travis yml .

bhuesemann:
> @bhuesemann, regarding netezza and others - we have custom repo with 3rd party jars on s3 repo.
> Looks like current master branch travis build works OK.

@javornikolov The current travis build on the branch fails due this missing reference. Is the Netezza JDBC driver on S3? What needs to be done to include it in the build of the branch. Should I send the missing jar to you?

dbfit-java/.gitignore
bhuesemann: Java 8 Build Error: Expected feature release number in range of 9 to 16, but got: 8, so I guess the codebase currently is compatible to JAVA 9.

bhuesemann: Issue resolved. Travis distro was changed (trusty -> xenial) and introduced this problem. I've added a config to the yml and now JDK 8 compiles well. Will try JDK 7 next.

bhuesemann: JDK7 doesn't compile even with gradle so I think JDK8 is best we can get

javornikolov: Newer versions of gradle (since version 6 I think) are not compatible with JDK 7 as far as I remember.

Current master branch used to build fine with jdk7. So I guess we should be able to make it work under it.

javornikolov:
> > @bhuesemann, regarding netezza and others - we have custom repo with 3rd party jars on s3 repo.
> > Looks like current master branch travis build works OK.
> 
> @javornikolov The current travis build on the branch fails due this missing reference. Is the Netezza JDBC driver on S3? What needs to be done to include it in the build of the branch. Should I send the missing jar to you?

We have nzjdbc-7.1.jar in s3. We can upload other versions if that is not good enough.

dbfit-java/.gitignore
bhuesemann: I've upgraded to Gradle 6 which needs jdk8 to run. https://docs.gradle.org/current/userguide/compatibility.html
I can try to compile it with Jdk 7 but this requires to have jdk7 installed in parallel. I think jdk 7 is out of support and hardly anyone will use it in production anymore. We should invest our time on something more productive and useful for users.

bhuesemann:
> > > @bhuesemann, regarding netezza and others - we have custom repo with 3rd party jars on s3 repo.
> > > Looks like current master branch travis build works OK.
> > 
> > 
> > @javornikolov The current travis build on the branch fails due this missing reference. Is the Netezza JDBC driver on S3? What needs to be done to include it in the build of the branch. Should I send the missing jar to you?
> 
> We have nzjdbc-7.1.jar and nzjdbc3-7.2.jar in s3 (and some older one). We can upload other versions if that is not good enough, you can send to me, yes.

Travis error is:
```
     Searched in the following locations:
       - file:/home/travis/build/dbfit/dbfit/custom_libs/nzjdbc3-7.2.jar
       - file:/home/travis/build/dbfit/dbfit/custom_libs/nzjdbc3.jar
       - https://repo.maven.apache.org/maven2/org/netezza/nzjdbc3/7.2/nzjdbc3-7.2.pom
       - https://aws.amazon.com/s3/ivy.xml
     Required by:
         project :dbfit-java:netezza
```
I don't understand yet, why this works on master and not on the branch. The ivy repo should do the trick, right?

javornikolov:
> I don't understand yet, why this works on master and not on the branch. The ivy repo should do the trick, right?

In general, yes. But there might be some caveats with the new gradle.

dbfit-java/.gitignore
javornikolov: @bhuesemann, fair enough. Let's drop the JDK7 support from dbfit 4.0, it's anyway a major release.

I suggest that we create a separate pull request for the java version updates only - I mean, moving to JDK8 as default target.

javornikolov: @bhuesemann, fair enough. Let's make JDK8 as main target and drop jdk7 support. Let's make it in a separate pull request dedicated for Java version update related stuff only.

dbfit-java/build.gradle
```diff
@@ -305,23 +310,24 @@ project('informix') {
...
 project('oracle') {
     description = 'DBFit Oracle engine'
 
     dependencies {
-        compile project(':dbfit-java:core')
-        compile 'dummy:ojdbc6:11.2.0.3.0@jar'
-        compile publicOracleCompileDeps
-        testCompile "org.mockito:mockito-core:${mockitoVersion}"
+        implementation project(':dbfit-java:core')
+		implementation 'dummy:ojdbc8:18.3.0.0.0@jar'
```
javornikolov, line 328: ojdbc is a problem as well, we haven't uploaded it to S3 and it's not in official repos. It's better to have separate pull request if we upgrade ojdbc to higher version. ojdbc6.jar for 11.2 is declared to be compatible with JDK8.

javornikolov:
Here is an example how to make netezza, ojdbc and other dependencies on S3 to work with gradle 6: https://github.com/dbfit/dbfit/pull/650

dbfit-java/build.gradle
bhuesemann: @javornikolov I've reverted to ojdbc6 and resolved the gradle ivy build problem (with your help). Next problem is the oracle and db2 docker containers don't start correctly. Can I help with this (in master the integration of docker containers for the databases is not present)?

bhuesemann: @javornikolov DB2 and Oracle docker image is fixed now. I've got a remaining issue with the dbdeploy ant task.

bhuesemann: @javornikolov dbdeploy and bundle issues are fixed now. Travis CI is ok now.

bhuesemann:
@javornikolov test-vm, travis-ci and all open stuff is fixed now. From my perspective we can take this as baseline for 4.0.0

dbfit-java/build.gradle
```diff
@@ -330,7 +330,8 @@ project('oracle') {
 
     dependencies {
         implementation project(':dbfit-java:core')
-		implementation 'dummy:ojdbc8:18.3.0.0.0@jar'		
+        implementation 'dummy:ojdbc6:11.2.0.3.0@jar'
+		// implementation 'dummy:ojdbc8:18.3.0.0.0@jar'
```
javornikolov, line 335: We don't need to keep that comment. Line formatting is not quite in sync with style used elsewhere (we use spaces for indentation, not tabs)

javornikolov:
@bhuesemann, that sounds great! May we just slice things down into separate pull requests per different topics e.g.:
 * docker refinements
 * supported java version bump up
 * gradle upgrade
 * code style reformatting - it doesn't seem bad, yet I'm not sure we need it in all places where you've done it. In any case - it's good to have such refactoring as separate PR in order to keep the other changes cleaner and easier to review now and in future.
 * gitignore enhancement
and so on

Also some commits may be squashed to avoid convoluted history of commits (like bumping up and down the oracle version).

dbfit-java/core/src/test/java/dbfit/diff/DataRowDiffTest.java
```diff
@@ -13,7 +13,7 @@
 import static org.hamcrest.Matchers.*;
 import static org.junit.Assert.*;
 
-import org.mockito.runners.MockitoJUnitRunner;
+import org.mockito.junit.MockitoJUnitRunner ;
```
javornikolov, line 16: unnecessary whitespace before `;`

javornikolov:
@bhuesemann as soon as you're already touching mockito and renaming things in multiple files, I recall that once we have java version upgraded we may move to junit 5 as well. Again - the suggestion is to have that in separate PRs which we can merge one by one if possible.

javornikolov:
It's also good to bump up FitNesse to most recent version.

bhuesemann:
@javornikolov breaking things into multiple small PRs and doing complete regression test for every PR will take time and I've been working on many different (small) things all over the project. I understand that you want to keep things manageable but it might get hard to separate changes especially when it comes to the gradle build config or travis. I'll see if I can do it in the granularity that you proposed and base everything on top of #608 . But I'll not open multiple PRs and branches in parallel to keep the merge overhead manageable.

javornikolov:
Yep, I imagine opening PRs one by one. Might be just a matter of reordering and tidying up commits here.
Maybe the automated tests would be enough for the intermediate PRs, having the confidence in the final state after merging them all.

Maybe travis and java or some other things need to go together, I don't know. That's OK when things are interdependent.

PR #650, opened 2020-07-28
Configure S3 repo to be gradle 6.x compatible

javornikolov:
Configure S3 repo to be gradle 6.x compatible

Gradle 6.x has different defaults for ivy repos. This change allows using dependencies from Amazon S3.

https://github.com/gradle/gradle/issues/10850


javornikolov:
I was about to close this as we're already on Gradle 6.x and it seems to work. Yet, there might be something like wildcard dependencies so I reopen it as a pointer for further exploration on what it is about.

javornikolov:
Closing as we're not using S3 anymore

PR #651, opened 2020-08-02
Travis CI: docker image switch and minor fixes

bhuesemann:
Includes:
- minor gradlew fix 
- travis ci: 
 -- switch to openjdk
 -- reflect changes in syntax of the travis yml
 -- switch base docker image for db2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,8 @@ services:
 language: java
 
 before_install:
-  - docker run -d -p 50000:50000 -v "`pwd`/test_vm/scripts/db2:/scripts" -e DB2INST1_PASSWORD=db2inst1-pwd -e LICENSE=accept --name dbfitdb2 --privileged=true cwds/db2 &
-  - docker run -d -p 1521:1521 -p 49160:22 -v "`pwd`/test_vm/scripts/oracle/sql:/docker-entrypoint-initdb.d:ro" wnameless/oracle-xe-11g-r2 &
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -1,33 +1,45 @@
...
 services:
   - docker
   - postgresql
 
 language: java
+
 before_install:
-  - docker run --name dbfitdb2 -p 50000:50000 -v "`pwd`/test_vm/scripts/db2:/scripts" -e DB2INST1_PASSWORD=db2inst1-pwd -e LICENSE=accept ibmcom/db2express-c:latest db2start &
-  - docker run -p 1521:1521 -p 49160:22 -v "`pwd`/test_vm/scripts/oracle/sql:/docker-entrypoint-initdb.d:ro" wnameless/oracle-xe-11g &
+  - docker run -d -p 50000:50000 -v "`pwd`/test_vm/scripts/db2:/scripts" -e DB2INST1_PASSWORD=db2inst1-pwd -e LICENSE=accept --name dbfitdb2 --privileged=true cwds/db2 &
```
javornikolov, line 12: Do we need, the `-d`? I think there used to be something with the level of detail logged. We anyway run them in background mode.

javornikolov:
I general looks good. Just this curiousity about -d switch

.travis.yml
bhuesemann: During debugging I had removed running the docker images in the background and thus I've detached the containers. It's not necessary if you run it in background on the OS level. I don't know if this has influence on logging. I'll revert it.

bhuesemann:
Should be ok now.

PR #654, opened 2020-08-04
Bump to jdk8

bhuesemann:
Contains jdk8 bump to prepare gradle upgrade in the next PR...


bhuesemann:
@javornikolov please review

javornikolov:
Thanks, @bluesmann!
There are still some changes of code formatting only which are hard to review through mobile phone and may complicate merges of some other pending PRs.

I will have access to computer from next week, until then I may be able to review simpler PRs.

bhuesemann:
@javornikolov No problem. I'll be patient, after all we do this in our spare time. I really appreciate that you take the time to bring this back to life. I'm convinced the project is worth our efforts. Hopefully @MMatten or others will join in. 

Regarding formatting I would find it useful to include something like https://github.com/google/google-java-format in the gradle build process to standardize formatting apart from checkstyle. The reformatting you see is the result from autoformatting with VSC and this is not optimal in all but many cases. Today I've possibly found a way to include code formatter in VSC here (https://www.sethvargo.com/using-google-java-format-with-vs-code/) and I will test it along with the gradle plugin. Support for the more traditional Java IDEs like Eclipse or IntelliJ are directly included. Nowadays I code more Python or C# than Java and thus I've switched everything from Eclipse to VSC.

build.gradle
```diff
@@ -155,7 +155,7 @@ task wikiDocs(dependsOn: [wikiDocsUpdateList]) {
 }
 
 def fitnesseJar() {
-    fileTree(dir: 'dist/lib', include: 'fitnesse-*.jar').getSingleFile()
+    fileTree(dir: 'dist/lib', include: 'fitnesse-*standalone.jar').getSingleFile()
```
javornikolov, line 158: @bhuesemann, have you faced any issues which required adding this `standalone` part of the jar filename? I think we are not pulling any other `fitnesse*.` jars but I might be wrong...

javornikolov:
> @javornikolov I've rebased this PR on master as you suggested. The release/4.0.0_bhn branch is not relevant anymore, as we have decided to break it into smaller pieces. This is the first smaller PR (jdk8). Next will be gradle update to (6.x). The docker and travis improvements we will then merge in the upcoming PRs. Meanwhile I've already created the Exasol plugin leaving only Snowflake support open on my list. Release 4.0 will be **huge**...

Thank you, @bhuesemann!
- Still a lot of stuff going on here (including conflicts due to not quite essential code reformatting). So I created PR #657 to easier and sooner move forward with the JDK8 bump up.
- Some other small changes I spin of as separate PRs as well and merge them (git settings for vscode and .box)
- I changed the target branch here to dbfit/master. Looks like a lot of extra commits got pulled here. (My practice is to run interactive rebase and get rid of unneeded commits `git rebase -i`).

build.gradle
bhuesemann: @javornikolov I think it is not required for the jdk8 upgrade but will be needed when upgrading gradle itself or the gradle java plugin to java-library plugin (to get advanced dependency management). 

As a side note jdk8 is not enough anymore to run Eclipse / VSC IDE in current or closely upcoming IDE releases (they require JDK 11):
https://www.eclipse.org/lists/eclipse-pmc/msg03821.html
https://github.com/redhat-developer/vscode-java/wiki/JDK-Requirements

This is causing project setup headaches for me, as Eclipse/VSC + Gradle 4.4 doesn't support setting the Gradle jdk via .settings
"java.import.gradle.java.home" (this requires Gradle >= 4.7)

and gradle 4.4 is not able to run with jdk >= JDK10. 

So the gradle update is needed and very closely related. 

It's all intertwined and many changes are needed to get this project into shape. And additionally all individual/intermediate steps we now take require specific handling either in the project setup or gradle config just to eventually change them again later again when we get to the next step (upgrading the next component or library).

javornikolov: @bhuesemann, I bumped up Gradle as well and merged it to master. Resolving deprecation is good idea as well, though I decided to leave it as separate change.

IDEs often can run in separate JRE/JDK than the JDK used to build projects within them.

PR #659, opened 2020-08-25
Bump up Gradle to 6.1.1

javornikolov:


javornikolov:
Intended to be Gradle 6.6, will push separate PR for that

PR #661, opened 2020-08-26
fix bin folder location in gradle eclipse generation of .classpath files

bhuesemann:
Eclipse and VSC IDE integration is supported by gradle by using the command
gradle clean cleanEclipse assemble eclipse

It will generate the .project and .classpath files needed by the IDEs and their internal Java tooling. However, the generated default compile directory might interfere with bin directories of the project and should be put under the gradle build folder in a separate folder. 

This PR will tweak gradle to put the move and rename the bin directories under:
<root>/dbfit-java/<subproject>/build/eclipse
instead of 
<root>/dbfit-java/<subproject>/bin

### DIFF
--- a/dbfit-java/build.gradle
+++ b/dbfit-java/build.gradle
@@ -55,6 +55,25 @@ allprojects {
     dependencies {
         compile "org.fitnesse:fitnesse:${fitNesseVersion}:standalone@jar"
     }
+    
+    // fix gradle .classpath generation for bin folders and put it under /build/eclipse
+	eclipse {
+		classpath {
+            defaultOutputDir = file('build/eclipse') // this should be enough, but it's not
+		    file.whenMerged {
+			   entries.each { entry ->
+					if (entry instanceof org.gradle.plugins.ide.eclipse.model.Output) {
+					   entry.path = entry.path.replace('bin/main', 'build/eclipse/main')
+					}
+					if (entry instanceof org.gradle.plugins.ide.eclipse.model.SourceFolder) {
+					   entry.output = entry.output.replace('bin/main', 'build/eclipse/main')
+                       entry.output = entry.output.replace('bin/test', 'build/eclipse/test')
+                       entry.output = entry.output.replace('bin/integrationTest', 'build/eclipse/integrationTest')
+					}
+			   }
+		   }
+		}
+	}     
 }
 
 subprojects {
--- a/dbfit-java/build.gradle
+++ b/dbfit-java/build.gradle
@@ -57,23 +57,22 @@ allprojects {
     }
     
     // fix gradle .classpath generation for bin folders and put it under /build/eclipse
-	eclipse {
-		classpath {
+    eclipse {
+        classpath {
             defaultOutputDir = file('build/eclipse') // this should be enough, but it's not
-		    file.whenMerged {
-			   entries.each { entry ->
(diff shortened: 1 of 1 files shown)


bhuesemann:
@javornikolov please review. You may remember me adding bin directories to git ignore with one of the previous PR candidates. This is in my opinion a better solution, that is much clearer and explicit. It took my some time to understand what is the root cause of this.

dbfit-java/build.gradle
```diff
@@ -55,6 +55,25 @@ allprojects {
     dependencies {
         compile "org.fitnesse:fitnesse:${fitNesseVersion}:standalone@jar"
     }
+    
+    // fix gradle .classpath generation for bin folders and put it under /build/eclipse
+	eclipse {
```
javornikolov, line 60: Formatting remark: Indentation is a bit inconsistent here. We're using 4-space indentation. No tabs. No trailing spaces on lines. Also better to surround `if` blocks with some empty lines to make it easier to see.

dbfit-java/build.gradle
```diff
@@ -55,6 +55,25 @@ allprojects {
...
+		classpath {
+            defaultOutputDir = file('build/eclipse') // this should be enough, but it's not
+		    file.whenMerged {
+			   entries.each { entry ->
+					if (entry instanceof org.gradle.plugins.ide.eclipse.model.Output) {
+					   entry.path = entry.path.replace('bin/main', 'build/eclipse/main')
```
javornikolov, line 66: Aren't output entries already OK?

javornikolov:
@bhuesemann, this looks good to me in general. Just fix formatting and simplify:
 - maybe it's enough to tweak only src = 'kind' entries. (`entries.fndAll { `it.hasProperty("output") && it.output.startsWith("/bin") }`
 - might be enough to just replace all "bin-like" for more concise expression. Yet no problem to be explicit about those
 - ensure we replace prefix only `output.replaceFirst("^bin", "build/eclipse")

dbfit-java/build.gradle
bhuesemann: No, and I don't know why this was not a problem before. May be gradle changed behavior here and Eclipse has not been used by somebody for some time. I think most professional Java Developers use IDEA and thus this may not have been a problem. What I see is that gradle changed the output dir from "bin" to "bin/[main|test|integrationTest]" with the upgrade to 6.6.

dbfit-java/build.gradle
bhuesemann: good catch! I'll try to get a formatter for gradle files and I've forced space indentation instead of autodetection in this workspace.

bhuesemann:
Good advice to make the replace more robust against future changes and more concise! Will integrate this.

javornikolov:
Merging via squash (2nd commit is more if a fixup)

bhuesemann:
@javornikolov sorry I don't understand. The PR has not been merged right? You want me to merge/squash the two commits and open a new PR?

javornikolov:
> @javornikolov sorry I don't understand. The PR has not been merged right? You want me to merge/squash the two commits and open a new PR?

@bhuesemann, sorry for the confusion! GitHub which suggests to do the squash when merging but seems I didn't got/use it accordingly (I was doing it from mobile phone). Now I reopened and merged into master.

PR #681, opened 2021-10-10
Bump up FitNesse to v20211006

MMatten:
Bump up FitNesse to v20211006.

NB: this has only been tested with `gradlew fastbuild`.


MMatten:
Thanks @javornikolov it's been so long since I've merged anything in GitHub we now have a number of methods: -
image

Which do you generally use?

javornikolov:
A merge commit :-)

Issue #683, opened 2021-10-10
Migrate build dependencies from Aamzon S3 to something else


javornikolov:
Migrate build dependencies from Aamzon S3 to something else.
One option might be separate repo here with binary jars only.

This applies to both:
* gradle build dependencies
* published artifacts (after Travis CI build)

PR #693, opened 2021-12-31
Migrate deps from AWS S3 to github

javornikolov:
Migrate binary dependencies from Amazon S3 to github

This is for the build dependencies only. S3 is still in use for the artefacts (Travis CI build) - to be handled separately

Resolves a part of #683

### DIFF
--- a/dbfit-java/build.gradle
+++ b/dbfit-java/build.gradle
@@ -43,8 +43,7 @@ allprojects {
             url "http://mvnrepository.com"
         }
         ivy {
-            artifactPattern "https://s3.amazonaws.com/dbfit/[artifact]-[revision].[ext]"
-            ivyPattern "https://aws.amazon.com/s3/ivy.xml"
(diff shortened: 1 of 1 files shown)


MMatten:
Hi @javornikolov, I'm unfamiliar with `raw.githubusercontent.com`. Is this actually intended for storing binaries?

If we don't talk again today then....best wishes for a spectacular 2022!

javornikolov:
> Hi @javornikolov, I'm unfamiliar with `raw.githubusercontent.com`. Is this actually intended for storing binaries?
> 
> If we don't talk again today then....best wishes for a spectacular 2022!

Thank you, @MMatten ! Let you have a spectacular and happy 2022! :-)

Hi @MMatten! I just took this `raw.githubusercontent` from Github UI link on files which allows to see their original content. It sems to work OK for now as I have been testing it while playing with Github Actions. So I am going to merge that PR.

PR #694, opened 2022-01-01
Migrate to GitHub actions

javornikolov:
Migrate CI from Travis CI to Github Actions

There is are lated issue and another branch with experiments on this.

This branch is initially experimental and needs a lot of cleanup.

Notes and open questions
* Oracle database name may be changed as this is available in the docker container for newer versions. Or alternatively we might initially stick to 11g. Ideally it's good to parameterize that.

Highlights
* Github Actions configured
* Distributio artifact is availble in builds as well, double zipped (as that seems to be what's available at the moment)
* Postgresql driver has been updated to comply with the default postgres service
* Resolves remaining part of #683

Some findings and caveats
* Fastbuild got stuck a few times so I cancelled it. Maybe is rare issue
* The git checkout is happening after services setup so mounting of the project as volume wasn't possible. I tried to setup docker within a run step to workaround that - however in such case there was some instability (even though it worked sometimes). So I refactored it via copying the necessary files to containers at later stage.
* Two runs appear on branch push when there is PR (I am not sure are they separate jobs in fact or are reports of one and the same)

TODO:
- [x] Tidy up commit history
- [x] Currently we step on branch of #693 (we should merge the other PR first or clean that out as it's a separate change)
- [x] Consider would we need a wide matrix of java versions or just one/few. (Just 8, 11 should be OK for now. Java 17 is not compatible with current gradle anyway and needs more work)
- [x] Oracle database name has changed, maybe we need to update docs or try to change the name. The Vagrant box perhaps would not be aligned with it. (We keep the new name. We could parameterize later if  needed)
- [x] Drop single steps of individual integration tests as they're anyway executed in the final travisbuild. (Refactored as separate jobs which run concurrently)
- [x] Rename `travisbuild` to `cibuild` or something more general. Or add more tasks for Github Actions build

TODO perhaps as future PRs
* Publishing release assets
* Other database integration tests (if easy to add)
* Possibly slice out integration tests of big databases as separate jobs in order to allow them running in parallel (in case that's faster)


javornikolov:
Hi @MMatten,
Happy New Year!

I got the work you started on Github Actions migrations (hacking on your branch and on another one here). It needs some cleanup, yet I believe it covers what essentially we used to have in Travis CI.

javornikolov:
Seems Github Actions builds are working well enough to go, I am merging this PR now so that we have a working CI.

PR #700, opened 2022-01-05
Use gradle action instead of gradle shell command

javornikolov:
* Use gradle-buildaction for the gradle builds https://github.com/gradle/gradle-build-action
* Rearrange fastbuild task so that it's simpler (just one java version) - so that upload artifacts has higher chance not to fail


javornikolov:
* Use gradle-buildaction for the gradle builds
* Rearrange fastbuild task so that it's simpler (just one java version) - so that upload artifacts has higher chance not to fail

